PR #7, opened 2015-02-22
Add histogram support.

brian-brazil:
@beorn7 

I didn't add linearBuckets/exponentialBuckets as Python has range() and list comprehensions are easy.

### DIFF
--- a/prometheus_client/__init__.py
+++ b/prometheus_client/__init__.py
@@ -16,11 +16,13 @@ except ImportError:
 from functools import wraps
 from threading import Lock
 
-__all__ = ['Counter', 'Gauge', 'Summary', 'CollectorRegistry']
+__all__ = ['Counter', 'Gauge', 'Summary', 'Histogram']
 
 _METRIC_NAME_RE = re.compile(r'^[a-zA-Z_:][a-zA-Z0-9_:]*$')
 _METRIC_LABEL_NAME_RE = re.compile(r'^[a-zA-Z_:][a-zA-Z0-9_:]*$')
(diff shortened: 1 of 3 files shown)


prometheus_client/__init__.py
```diff
@@ -300,6 +304,73 @@ def _samples(self):
...
+  elif d == _MINUS_INF:
+    return '-Inf'
+  else:
+    return repr(d)
+
+@_MetricWrapper
+class Histogram(object):
+  _type = 'histogram'
+  _reserved_labelnames = ['histogram']
+  def __init__(self, buckets=(.005, .01, .025, .05, .075, .1, .25, .5, .75, 1.0, 2.5, 5.0, 7.5, 10.0, _INF)):
```
beorn7: This .75 in there really gives me the creeps... I think the default buckets should have a more regular spacing.

prometheus_client/__init__.py
```diff
@@ -300,6 +304,73 @@ def _samples(self):
...
+    self._sum = 0.0
+    self._lock = Lock()
+    buckets = [float (b) for b in buckets]
+    if buckets != sorted(buckets):
+      # This is probably an error on the part of the user,
+      # so raise rather than sorting for them.
+      raise ValueError('Buckets not in sorted order')
+    if not buckets:
+      raise ValueError('Must have at least one bucket')
```
beorn7: But what if that one bucket is _INF?

I'd add the check after adding the _INF and then make sure there are at least two buckets.

beorn7:
:+1: otherwise.

prometheus_client/__init__.py
brian-brazil: Done

PR #11, opened 2015-03-18
PEP8 fixes

mmoya:


brian-brazil:
Shouldn't it be a 4 space indent for pep8?

mmoya:
I prefer 4 spaces but I thought it would be too intrusive. Would resubmit the PR adding that change.

brian-brazil:
We can do that as a separate change then. Can you squash down the commits?

mmoya:
Merged the two pep8 related commits.

juliusv:
@mmoya Woot, thanks!

PR #13, opened 2015-03-18
Add support for exporting metrics to a pushgateway

mmoya:
Add support for exporting metrics to a pushgateway. By default the fqdn is sent is _instance_ wasn't provided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,3 +181,19 @@ write_to_textfile('/configured/textfile/path/raid.prom', registry)
 ```
 
 A separate registry is used, as the default registry may contain other metrics.
+
+## Exporting to a Pushgateway
+
+The [Pushgateway](https://github.com/prometheus/pushgateway)
+allows ephemeral and batch jobs to expose their metrics to Prometheus.
+
+```python
+from prometheus_client import CollectorRegistry,Gauge,build_pushgateway_url,push_to_gateway
+registry = CollectorRegistry()
+g = Gauge('raid_status', '1 if raid array is okay', registry=registry)
+g.set(1)
+url = build_pushgateway_url(job='cooljob', host='pushgateway.mydomain')
+push_to_gateway(url, registry)
+```
+
+A separate registry is used, as the default registry may contain other metrics.
--- a/README.md
+++ b/README.md
@@ -188,12 +188,11 @@ The [Pushgateway](https://github.com/prometheus/pushgateway)
 allows ephemeral and batch jobs to expose their metrics to Prometheus.
 
 ```python
-from prometheus_client import CollectorRegistry,Gauge,build_pushgateway_url,push_to_gateway
+from prometheus_client import CollectorRegistry,Gauge,push_to_gateway
 registry = CollectorRegistry()
 g = Gauge('raid_status', '1 if raid array is okay', registry=registry)
 g.set(1)
-url = build_pushgateway_url(job='cooljob', host='pushgateway.mydomain')
-push_to_gateway(url, registry)
+push_to_gateway(registry, job='somejob', host='pushgateway.mydomain')
 ```
 
 A separate registry is used, as the default registry may contain other metrics.
--- a/README.md
+++ b/README.md
@@ -190,9 +190,9 @@ allows ephemeral and batch jobs to expose their metrics to Prometheus.
 ```python
 from prometheus_client import CollectorRegistry,Gauge,push_to_gateway
 registry = CollectorRegistry()
-g = Gauge('raid_status', '1 if raid array is okay', registry=registry)
-g.set(1)
-push_to_gateway(registry, job='somejob', host='pushgateway.mydomain')
+g = Gauge('job_finished_ok_at', 'last time a batch job successfully finished', registry=registry)
+g.set_to_current_time()
+push_to_gateway(registry, job='batchA', host='pushgateway.mydomain')
 ```
 
 A separate registry is used, as the default registry may contain other metrics.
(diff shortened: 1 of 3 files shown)


prometheus_client/__init__.py
```diff
@@ -434,6 +442,34 @@ def do_GET(self):
...
+
+    url = 'http://{}:{}/metrics/jobs/{}{}'.format(host, port, job, instancestr)
+    return url
+
+
+def push_to_gateway(url, registry, timeout=None):
+    '''Push metrics to the given url'''
+
+    if not HAVE_REQUESTS:
+        raise RuntimeError("requests module is not available")
```
brian-brazil: As painful as it is, we should stick to urllib for the broadest possible compatibility. I want to avoid any dependencies outside of core cpython, so that there's no friction to instrumenting code.

brian-brazil:
I've been holding off on adding this until https://github.com/prometheus/pushgateway/issues/18 is resolved, as the 'instance' handling isn't quite compatible with the usual use cases.

mmoya:
Resubmitted using `urllib2`. Once prometheus/pushgateway#18 is sorted out, I'll update the PR. The _use_fqdn_as_instance_ param was a hack to overcome it.

README.md
```diff
@@ -181,3 +181,19 @@ write_to_textfile('/configured/textfile/path/raid.prom', registry)
...
+
+The [Pushgateway](https://github.com/prometheus/pushgateway)
+allows ephemeral and batch jobs to expose their metrics to Prometheus.
+
+```python
+from prometheus_client import CollectorRegistry,Gauge,build_pushgateway_url,push_to_gateway
+registry = CollectorRegistry()
+g = Gauge('raid_status', '1 if raid array is okay', registry=registry)
+g.set(1)
+url = build_pushgateway_url(job='cooljob', host='pushgateway.mydomain')
```
brian-brazil: This should all be one method

README.md
```diff
@@ -181,3 +181,19 @@ write_to_textfile('/configured/textfile/path/raid.prom', registry)
...
+## Exporting to a Pushgateway
+
+The [Pushgateway](https://github.com/prometheus/pushgateway)
+allows ephemeral and batch jobs to expose their metrics to Prometheus.
+
+```python
+from prometheus_client import CollectorRegistry,Gauge,build_pushgateway_url,push_to_gateway
+registry = CollectorRegistry()
+g = Gauge('raid_status', '1 if raid array is okay', registry=registry)
+g.set(1)
```
brian-brazil: I'd go with `set_to_current_time()` as the example, as pushing the end time of a cronjob will be one of the most common use cases.

prometheus_client/__init__.py
```diff
@@ -434,6 +437,33 @@ def do_GET(self):
         self.wfile.write(generate_latest(REGISTRY))
 
 
+def build_pushgateway_url(job, instance=None, host='localhost', port=9091,
+                          use_fqdn_as_instance=True):
```
brian-brazil: I'd remove all of the instance logic, it's a very rare use case and causes confusion about how the main uses cases are meant to work

prometheus_client/__init__.py
```diff
@@ -434,6 +437,33 @@ def do_GET(self):
...
+    '''
+    Build a valid pushgateway url
+    '''
+
+    if instance is None and use_fqdn_as_instance:
+        instance = socket.getfqdn()
+
+    if instance:
+        instancestr = '/instances/{}'.format(instance)
```
brian-brazil: Any label values coming in need to be url escaped

mmoya:
I messed this up, deleted the branch by accident and that closed the PR.

juliusv:
@brian-brazil @mmoya - just noticed this PR sitting here for almost a month - is it still current?

brian-brazil:
This is still current, we're waiting on the instance label cleanup on the push_gateway side.

oliver006:
Any idea when this might land? Would love to see that functionality.

brian-brazil:
Looks like we're pretty close to having all the required changes elsewhere.

This needs to be updated to set the `instance` label to the empty string if it's not present already before exporting.

prometheus_client/__init__.py
```diff
@@ -434,6 +443,38 @@ def do_GET(self):
         self.wfile.write(generate_latest(REGISTRY))
 
 
+def build_pushgateway_url(job, instance=None, host='localhost', port=9091):
```
brian-brazil: This should be prefixed with _ to mark it as internal

prometheus_client/__init__.py
```diff
@@ -434,6 +443,38 @@ def do_GET(self):
...
+    '''
+    Build a valid pushgateway url
+    '''
+
+    if instance:
+        instancestr = '/instances/{}'.format(instance)
+    else:
+        instancestr = ''
+
+    url = 'http://{}:{}/metrics/jobs/{}{}'.format(host, port,
```
brian-brazil: This should be `job` now

brian-brazil: You're assuming http, best to let the user pass in all that as a prefix

prometheus_client/__init__.py
```diff
@@ -434,6 +443,38 @@ def do_GET(self):
...
+                                                  quote(job),
+                                                  quote(instancestr))
+    return url
+
+
+def push_to_gateway_url(url, registry, timeout=None):
```
brian-brazil: I'd stick with just this function

brian-brazil:
> This needs to be updated to set the instance label to the empty string if it's not present already before exporting.

Wait, that's not right. No change required there as the pushgateway handles that.

You'll need up pull in the latest changes to the repo, and things have been split into separate files.

brian-brazil:
I've updated this and merged it. Thanks!

PR #26, opened 2015-04-13
Add the ability to specify labels as a dict of labelname->labelvalue.

korfuri:
This is useful for code that knows some of the labels in different places.

An example use case is a helper I'm trying to write that would work as such:

c = Counter('http_exceptions_total', 'help', ['method', 'exception_type'])
def handle_request(request):
    with count_exceptions_by_type(c, {'method': request.method}):
        do_something(request)

This helper only knows about the exception_type label and only passes
the other labels through.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,15 @@ c.labels('get', '/').inc()
 c.labels('post', '/submit').inc()
 ```
 
+Labels can also be provided as a dict:
+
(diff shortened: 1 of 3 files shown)


korfuri:
per CONTRIBUTING.md, ping @brian-brazil :)

PR #31, opened 2015-05-26
Prevent users to use test-only code in production

grunskis:
This moves unit test specific code to the tests file to prevent users of the client from using it in production code.


brian-brazil:
This code is intended to be used by users' own unittesting of their collectors, in addition to client_python's own unittests.

brian-brazil:
No response, closing.

PR #36, opened 2015-08-12
State that Summary does not support quantiles

arturhoo:


README.md
```diff
@@ -142,6 +142,8 @@ with s.time():
   pass
 ```
 
+**Summary support is incomplete**. The python client doesn't store or expose quantile information at this time.
```
brian-brazil: That's a bit strong, this is a complete client library per the spec. I'd remove the first sentence.

arturhoo:
I've probably missed something then. From [the docs](http://prometheus.io/docs/concepts/metric_types/) I was expecting Summary to expose:

* streaming φ-quantiles (0 ≤ φ ≤ 1) of observed events, exposed as `<basename>{quantile="<φ>"}`
* the total sum of all observed values, exposed as `<basename>_sum`
* the count of events that have been observed, exposed as `<basename>_count`

But indeed, the [Guidelines] does mention the client libraries are only encouraged to expose quantiles. From my humble understanding either the docs should be more flexible, or the libraries very strict. @brian-brazil what do you think?

brian-brazil:
The docs need to be relaxed, using summaries without quantiles should be much more common than using them with (quantiles are quite expensive).

arturhoo:
Nice, do you recommend opening a PR on the docs then? I've removed the incomplete part and renamed this PR.

README.md
```diff
@@ -142,6 +142,8 @@ with s.time():
   pass
 ```
 
+The python client doesn't store or expose quantile information at this time.
```
brian-brazil: We're capitalising Python elsewhere, can you do the same for consistency?

arturhoo:
Done.

README.md
```diff
@@ -142,6 +142,8 @@ with s.time():
   pass
 ```
 
+The Python Client doesn't store or expose quantile information at this time.
```
brian-brazil: We don't capitalise client :(

arturhoo: k

PR #39, opened 2015-09-03
Added default registry to write_to_file

barkerd427:
@brian-brazil: This is needed to use the default registry with the write_to_file function. I have tested this locally.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -67,7 +67,7 @@ def start_http_server(port, addr=''):
     t.start()
 
 
-def write_to_textfile(path, registry):
+def write_to_textfile(path, registry=core.REGISTRY):
     '''Write metrics to the given path.
 
     This is intended for use with the Node exporter textfile collector.


brian-brazil:
The push gateway is intended primarily for use by batch and other ephemeral jobs where a limited number of metrics are pushed, thus having all metrices being pushed is not a good default as it is not in line with the prometheus architecture and has pitfalls that aren't obvious to the casual user.

It's still possible for you to do this by hand though as `REGISTRY` is publically exposed.

barkerd427:
This is for the node_exporter. It looked like the push gateway was already taking the default registry if none was provided. I'm not familiar enough with python to get it directly. Could you show me and maybe I can add that back for the docs or write a blog post describing my use-case.

brian-brazil:
Hmm, that's a bad default. I should fix that.

The same applies to the textfile collector, but more so as if you've two jobs writing the standard exports to  be read it'll break your metrics collection. These examples of how to use this in the README of this repository, but you definitely don't want to use the default registry with `write_to_file`.

barkerd427:
I see now. Thanks for the help.

PR #42, opened 2015-09-05
Added docstrings for Counter, Summary and Histogram classes

machinelady:


prometheus_client/core.py
```diff
@@ -329,6 +347,19 @@ def _samples(self):
...
+
+    Example use cases for Summaries:
+    - Response latency
+    - Request size
+
+    Example for a Summary:
+
+    from prometheus_client import Summary
+    s = Summary('request_latency_seconds', 'Description of summary')
+    s.observe(4.7)    # Observe 4.7 (seconds in this case)
```
brian-brazil: `time()` would be a better way to demonstrate tracking durations, this example would encourage users to do that by hand. I'd suggest having a time() example, and a separate one for something like bytes

prometheus_client/core.py
```diff
@@ -183,6 +183,24 @@ def collect():
 
 @_MetricWrapper
 class Counter(object):
+    '''
```
brian-brazil: Multi-line docstrings should have a single line at the very start, so move the line below up

https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

prometheus_client/core.py
machinelady: I agree, that could be misleading. Since that example is from the README file, should I go ahead and change it there as well?

brian-brazil: Yes please

machinelady: Done!

prometheus_client/core.py
```diff
@@ -329,6 +346,29 @@ def _samples(self):
...
+
+    Example use cases for Summaries:
+    - Response latency
+    - Request size
+
+    Example for a Summary:
+
+    from prometheus_client import Summary
+    s = Summary('request_size_bytes', 'Request size (bytes)')
+    s.observe(512) # Observe 512 (bytes)
```
brian-brazil: There should be at least two spaces before the #

brian-brazil:
Can you squash your commits please?

machinelady:
Closing this PR because of git issues, I'll open a new one in a moment.

Issue #46, opened 2015-09-13
tests fail for Python 3.4 on Debian 8


taibhse:
I forked the repo, have working tests on Python2, but they fail on Python 3.4.

python3 --version: Python 3.4.2
Debian 8 (jessie)

python3 setup.py test 2> out.txt

```
test_block_decorator (tests.test_client.TestCounter) ... ok
test_function_decorator (tests.test_client.TestCounter) ... ok
test_increment (tests.test_client.TestCounter) ... ok
test_negative_increment_raises (tests.test_client.TestCounter) ... ok
test_block_decorator (tests.test_client.TestGauge) ... ok
test_function_decorator (tests.test_client.TestGauge) ... ok
test_gauge (tests.test_client.TestGauge) ... ok
test_gauge_function (tests.test_client.TestGauge) ... ok
test_counter (tests.test_client.TestGenerateText) ... ok
test_escaping (tests.test_client.TestGenerateText) ... ok
test_gauge (tests.test_client.TestGenerateText) ... ok
test_nonnumber (tests.test_client.TestGenerateText) ... ok
test_summary (tests.test_client.TestGenerateText) ... ok
test_unicode (tests.test_client.TestGenerateText) ... ok
test_block_decorator (tests.test_client.TestHistogram) ... ok
test_function_decorator (tests.test_client.TestHistogram) ... ok
test_histogram (tests.test_client.TestHistogram) ... ok
test_labels (tests.test_client.TestHistogram) ... ok
test_setting_buckets (tests.test_client.TestHistogram) ... ok
test_child (tests.test_client.TestMetricWrapper) ... ok
test_incorrect_label_count_raises (tests.test_client.TestMetricWrapper) ... ok
test_invalid_names_raise (tests.test_client.TestMetricWrapper) ... ok
test_labels_by_dict (tests.test_client.TestMetricWrapper) ... ok
test_labels_coerced_to_string (tests.test_client.TestMetricWrapper) ... ok
test_namespace_subsystem_concatenated (tests.test_client.TestMetricWrapper) ... ok
test_non_string_labels_raises (tests.test_client.TestMetricWrapper) ... ok
test_remove (tests.test_client.TestMetricWrapper) ... ok
test_namespace (tests.test_client.TestProcessCollector) ... ok
test_working (tests.test_client.TestProcessCollector) ... ok
test_working_584 (tests.test_client.TestProcessCollector) ... ok
test_working_fake_pid (tests.test_client.TestProcessCollector) ... ok
[...]

PR #50, opened 2015-10-02
Several small test fixes

andersschuller:
@brian-brazil: This bundles together a few small fixes for the tests. Please let me know if you would prefer them as separate pull requests.

- First of all, `test_client.py` contained two tests with the same name (`test_pushadd_with_groupingkey`), so I renamed the second one to the more appropriate `test_delete_with_groupingkey`.

- I also noticed a race condition in the tests, where the assertions that check the received request could be run just before the requests are stored. I fixed this by calling `send_response` after the request is stored, instead of before.

- I then looked at issue #46. It seems there are some small differences in either the HTTP server or client between Python 2.7 and 3.4, which means a call to `end_headers` is required in the HTTP handler for Python 3.4.

- Finally, I ran into some issues running the tests on my Windows machine. It turns out that `socket.connect` throws an error on Windows when using '0.0.0.0' or the empty string as the target host (although either seems to work fine on Linux). I therefore changed the tests to use 'localhost' as the target host instead. This does also affect the `instance_ip_grouping_key` function exposed in the client module, but from some quick testing on my Linux machine there seems to be no difference between the old and new way.

Let me know if you have any questions or comments.

### DIFF
--- a/tests/graphite_bridge.py
+++ b/tests/graphite_bridge.py
@@ -28,7 +28,9 @@ class TestGraphiteBridge(unittest.TestCase):
         self.t = ServingThread()
         self.t.start()
         
-        self.gb = GraphiteBridge(server.server_address, self.registry, _time=FakeTime())
+        # Explicitly use localhost as the target host, since connecting to 0.0.0.0 fails on Windows
+        address = ('localhost', server.server_address[1])
+        self.gb = GraphiteBridge(address, self.registry, _time=FakeTime())
 
     def test_nolabels(self):
         counter = Counter('c', 'help', registry=self.registry)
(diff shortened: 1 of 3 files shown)


tests/graphite_bridge.py
```diff
@@ -28,7 +28,8 @@ def run(self):
         self.t = ServingThread()
         self.t.start()
         
-        self.gb = GraphiteBridge(server.server_address, self.registry, _time=FakeTime())
+        address = ('localhost', server.server_address[1])
```
brian-brazil: Can you add a comment here so a future Unix-user doesn't undo this?

brian-brazil:
Thanks for all the fixes, can you squash the commits please?

andersschuller:
All right, commits squashed.

brian-brazil:
Thanks!

PR #52, opened 2015-10-02
Fixed graphite.py formatting.

mafagafogigante:
This file had varying indentation levels.
There was also an unused local.


brian-brazil:
Thanks!

PR #54, opened 2015-10-02
WIP: Text format parser for Python

brian-brazil:
@beorn7 Would you mind taking a peek at this to see if I've missed anything major? Main code is in parser.py, the rest is surrounding changes for this to make sense and have the right data structures.

The motivation is that Zalando's Zmon has a partial implementation of the text parser, so I'd like to offer them an official option.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -74,7 +74,14 @@ _METRIC_TYPES = ('counter', 'gauge', 'summary', 'histogram', 'untyped')
 
 
 class Metric(object):
-    '''A single metric and it's samples.'''
+    '''A single metric family and its samples.
+
+    This is intended only for internal use by the instrumentation client.
+
+
+    Custom collectors should use GaugeMetricFamily, CounterMetricFamily
+    and SummaryMetricFamily instead.
+    '''
     def __init__(self, name, documentation, typ):
         self._name = name
         self._documentation = documentation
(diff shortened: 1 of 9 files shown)


beorn7:
I'll have a look ASAP.

prometheus_client/core.py
```diff
@@ -74,18 +75,145 @@ def get_sample_value(self, name, labels=None):
...
 class Metric(object):
-    '''A single metric and it's samples.'''
+    '''A single metric family and its samples.
+
+    This is intended only for internal use by the instrumentation client.
+
+    Custom collectors should use GaugeMetricFamily, CounterMetricFamily
+    and SummaryMetricFamily instead.
+    '''
     def __init__(self, name, documentation, typ):
```
beorn7: Has nothing to do with the change, but: Why `typ` and not `type`?

beorn7: NM, keywords are evil... :-(

prometheus_client/parser.py
```diff
@@ -0,0 +1,197 @@
...
+    labels = {}
+
+    state = 'name'
+
+    for char in text:
+        if state == 'name':
+            if char == '{':
+                state = 'labelname'
+            elif char == ' ' or char == '\t':
+                state = 'endoflabels'
```
beorn7: The text format is whitespace agnostic (excluding new-line), so whitespace between the metric name and the labels is fine.

prometheus_client/parser.py
```diff
@@ -0,0 +1,197 @@
...
+            else:
+                name.append(char)
+        elif state == 'labelname':
+            if char == '=':
+                state = 'labelvaluequote'
+            elif char == '}':
+                state = 'endoflabels'
+            else:
+                labelname.append(char)
```
beorn7: It looks this will include leading whitespace into the labelname. Also, it looks like there is no validation of the labelname.

beorn7:
Before checking further: Let's confirm what the intention WRT whitespace is. The text format is supposed to have any amount of non-new-line whitespace between tokens. `text_parse_test.go` includes things like

```
 # HELP  name2  	doc str"ing 2
  #    TYPE    name2 gauge
name2{labelname="val2"	,basename   =   "basevalue2"		} +Inf 54321
name2{ labelname = "val1" , }-Inf
```

prometheus_client/parser.py
brian-brazil: Updated.

prometheus_client/parser.py
brian-brazil: Fixed. I haven't added any validation for the initial version, as it doesn't matter for the current use cases and the client on the other end should have taken care of it.

beorn7:
OK, looks sane to me. Just sprinkle a few crazy whitespace characters over the test cases to see if it works.

brian-brazil:
The `test_spaces` test should cover that. If we ever make a breaking change to the text format, it'd be good to get rid of all this space flexibility from a parsing standpoint.

beorn7:
Ah, didn't see that... with the pre-squashed commits, it's difficult to see what has changed...

:+1: 

But definitely leave the whitespace flexibility in place. That's what people expect when they see this kind of syntax. Remember that the text format is also meant to be easy for humans to assemble.

Issue #51, opened 2015-10-02
Calling instance_ip_grouping_key leads to a ResourceWarning on Python 3.4


andersschuller:
I noticed this when working on fixing #46. When I run `python3 setup.py test`, I get the following warning in the test output:

> test_instance_ip_grouping_key (tests.test_client.TestPushGateway) ... tests/test_client.py:478: ResourceWarning: unclosed \<socket.socket fd=7, family=AddressFamily.AF_INET, type=SocketKind.SOCK_DGRAM, proto=0, laddr=('127.0.0.1', 58422)>
> self.assertTrue('' != instance_ip_grouping_key()['instance'])

I can also reproduce this warning outside of the tests by running `python3 -Wdefault` and then calling `instance_ip_grouping_key`.

Should this function explicitly close the socket it opens instead of relying on it to be cleaned up automatically?

brian-brazil:
Yes it should, this is simple so I'll add it to the Hacktoberfest list.

kvikas:
Hello Brian, if it is not taken yet, can I take this?

I got to know about prometheus from Hactoberfest. Project looks interesting. I was looking for a TSDB implementation. I'm a go newbie but know a fair bit of python. So I think fixing the python stuff will get me started to know more about prometheus.

Thanks.

brian-brazil:
Go ahead, there's nothing on this issue or in the google doc indicating anyone else is working on it. 

This is a standard resource leak, if you want to get a bit more into Prometheus I'd suggest looking at the bridge options especially if you're already knowledgeable in any other monitoring/TSDB system as that'll give you a feel for how the Prometheus data model compares.

kvikas:
Thanks Brian, done.

brian-brazil:
Fixed by #55, thanks!

PR #55, opened 2015-10-03
Closing open socket in instance_ip_grouping_key()

kvikas:
Issue #51 

Wrapped socket operation inside context manager to makes sure it gets closed.


juliusv:
@kvikas Thanks! :)

PR #61, opened 2015-10-18
Add bridge to New Relic

andersschuller:
This adds a `PrometheusDataSource` class which can be used for exporting metrics to New Relic, as detailed in the additions to the README. Although I started by creating something similar to the Graphite bridge, this ended up being quite different, as it seemed more convenient to have the New Relic agent gather metrics from the registry whenever it is collecting all of them, rather than calling `record_custom_metric` manually or in another background thread. Especially the fact that you can configure the custom data source from code or from a configuration file is quite neat. However it would probably not be much work to implement something more like the Graphite bridge's manual push / background thread if you wanted to support both.

Unfortunately the tests for the custom data source are quite tied to the way that the `@newrelic.agent.data_source_factory` decorator works. I think the only way to avoid this would be to have the data source class without the decorator for testing, and then have a method that adds the decorator and returns an instance of the class.

Let me know what you think.

### DIFF
--- a/prometheus_client/bridge/newrelic.py
+++ b/prometheus_client/bridge/newrelic.py
@@ -1,0 +1,27 @@
+from __future__ import absolute_import
+
+import newrelic.agent
+
+from .. import core
+
+
+@newrelic.agent.data_source_factory(name='Prometheus')
+class PrometheusDataSource(object):
+    def __init__(self, settings, environ):
+        self.registry = settings.get('registry', core.REGISTRY)
+        self.prefix = settings.get('prefix', '')
+
+    def __call__(self):
+        prefixstr = ''
+        if self.prefix:
+            prefixstr = self.prefix + '.'
+
+        for metric in self.registry.collect():
+            for name, labels, value in metric.samples:
+                if labels:
+                    labelstr = '.' + '.'.join(['{0}.{1}'.format(k, v) for k, v in sorted(labels.items())])
+                else:
+                    labelstr = ''
+
+                metric_name = 'Custom/{0}{1}{2}'.format(prefixstr, name, labelstr)
+                yield (metric_name, value)
--- a/README.md
+++ b/README.md
@@ -294,6 +294,42 @@ gb.push()
 gb.start(10.0)
 ```
 
+### New Relic
+
+In order to expose your metrics to New Relic, you need to register a custom data source with the
+[New Relic agent](https://docs.newrelic.com/docs/agents/python-agent):
+
+```python
+import newrelic.agent
+from prometheus_client.bridge.newrelic import PrometheusDataSource
+newrelic.agent.register_data_source(PrometheusDataSource)
+```
+
+Alternatively you can add the following to your `newrelic.ini` configuration file:
+
+```ini
+[data-source:prometheus]
+enabled = true
+function = prometheus_client.bridge.newrelic:PrometheusDataSource
+```
+
+This will export all of your metrics with names like `Custom/metric_name`. By defining settings for the data source as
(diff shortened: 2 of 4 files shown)


prometheus_client/bridge/newrelic.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __call__(self):
+        prefixstr = ''
+        if self.prefix:
+            prefixstr = self.prefix + '.'
+
+        for metric in self.registry.collect():
+            for name, labels, value in metric.samples:
+                if labels:
+                    labelstr = '.' + '.'.join(['{0}.{1}'.format(k, v) for k, v in sorted(labels.items())])
```
brian-brazil: You should sanitize or escape any periods in the label values.

prometheus_client/bridge/newrelic.py
```diff
@@ -0,0 +1,27 @@
...
+        self.registry = settings.get('registry', core.REGISTRY)
+        self.prefix = settings.get('prefix', '')
+
+    def __call__(self):
+        prefixstr = ''
+        if self.prefix:
+            prefixstr = self.prefix + '.'
+
+        for metric in self.registry.collect():
+            for name, labels, value in metric.samples:
```
brian-brazil: In the cases of a Summary and Histogram, would it make sense to return a dict for value with the `_count`/`_sum` as Count/Value for those two samples?

prometheus_client/bridge/newrelic.py
```diff
@@ -0,0 +1,27 @@
...
+
+import newrelic.agent
+
+from .. import core
+
+
+@newrelic.agent.data_source_factory(name='Prometheus')
+class PrometheusDataSource(object):
+    def __init__(self, settings, environ):
+        self.registry = settings.get('registry', core.REGISTRY)
```
brian-brazil: I'm not sure it's possible to reasonably use this setting, I suspect you should drop and leave other registries to the push method.

prometheus_client/bridge/newrelic.py
brian-brazil: https://docs.newrelic.com/docs/plugins/plugin-developer-resources/developer-reference/metric-naming-reference indicates that / * | [ ] also need special handling.

Having the labels as metric attributes using /  would seem a better fit of the data model than a dotted string.

brian-brazil:
This general approach sounds good, we can always add something push based later.

prometheus_client/bridge/newrelic.py
andersschuller: Ah yes, good point. I'll take a look at this when I get a chance.

prometheus_client/bridge/newrelic.py
andersschuller: Yep, I did see [this bit of the docs](https://docs.newrelic.com/docs/agents/python-agent/supported-features/python-custom-metrics#pre-aggregated-metrics) about returning a dictionary, but for some reason didn't think to apply it here. I'll take a look.

prometheus_client/bridge/newrelic.py
andersschuller: It's possible to use it when using `register_data_source`, but the docs do suggest to use string-value settings for everything (as you can only set string values in the .ini file). I think I need to be able to use a custom registry in the tests, but I'll see if I can find another way to do that.

brian-brazil:
Are you still working on this change?

andersschuller:
Sorry, I haven't had any more free time to work on this. If you (or someone else) would like to pick up my changes and make the changes you suggested, that's fine by me.

brian-brazil:
This review is very stale by now.

PR #62, opened 2015-10-19
Add a filter option to the graphite bridge.

iksaif:
This makes it easy to only export a set of the metrics to Graphite
which is useful when you need only part of your metrics in Graphite.

### DIFF
--- a/prometheus_client/bridge/graphite.py
+++ b/prometheus_client/bridge/graphite.py
@@ -50,7 +50,7 @@ class GraphiteBridge(object):
         self._timeout = timeout_seconds
         self._time = _time
 
-    def push(self, prefix=''):
+    def push(self, prefix='', filter_re=None):
         now = int(self._time.time())
         output = []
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
If you want to only export some metrics you should use a custom registry, rather than using a filter.

iksaif:
That means that graphite specific choices will influence the way I use prometheus.

In this case, I'd like to use prometheus as usual, putting all my metrics into /metrics, which is totally what I want but I would also like to export a subset of these metrics to graphite.

brian-brazil:
That should work fine with custom registries, a collector can be in more than one registry.

iksaif:
Right, I guess one way to not affect my program too much is to have two registries. The default one when I store everything and another one that I use only for what I want in graphite. Is that what you meant ?

I guess I won't have too much data so it should be ok to duplicate it.

brian-brazil:
Yes, that's what I meant.

Using a regex is not the ideal solution as you're dealing with the samples as munged strings. rather than time series with labels. There'll be a feature in future for http exposition allowing requesting a subset of metrics, and it'd be best to tie this sort of thing into that rather than a graphite-specific solution.

Issue #84, opened 2016-06-16
An exception in a custom collector results in a 200 with a blank body


FauxFaux:
If you have a custom collector registered, and that collector exceptions, then the client produces a 200 with no metrics; not even the build in ones.

``` python
#!/usr/bin/python3

import time
from prometheus_client import start_http_server
from prometheus_client.core import REGISTRY


class CustomCollector(object):
    def collect(self):
        raise Exception('bang')

REGISTRY.register(CustomCollector())
start_http_server(23456)

while True:
    time.sleep(1e9)
```

```
% curl -v http://localhost:23456/
*   Trying ::1...
* connect to ::1 port 23456 failed: Connection refused
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 23456 (#0)
> GET / HTTP/1.1
> Host: localhost:23456
> User-Agent: curl/7.47.0
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Server: BaseHTTP/0.6 Python/3.5.1+
< Date: Thu, 16 Jun 2016 09:18:40 GMT
< Content-Type: text/plain; version=0.0.4; charset=utf-8
< 
* Closing connection 0
```

```
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 43974)
Traceback (most recent call last):
  File "/usr/lib/python3.5/socketserver.py", line 313, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python3.5/socketserver.py", line 341, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python3.5/socketserver.py", line 354, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.5/socketserver.py", line 681, in __init__
    self.handle()
  File "/usr/lib/python3.5/http/server.py", line 415, in handle
    self.handle_one_request()
  File "/usr/lib/python3.5/http/server.py", line 403, in handle_one_request
    method()
  File "/home/faux/code/client_python/prometheus_client/exposition.py", line 76, in do_GET
    self.wfile.write(generate_latest(core.REGISTRY))
  File "/home/faux/code/client_python/prometheus_client/exposition.py", line 55, in generate_latest
    for metric in registry.collect():
  File "/home/faux/code/client_python/prometheus_client/core.py", line 54, in collect
    for metric in collector.collect():
  File "err.py", line 10, in collect
[...]

PR #85, opened 2016-06-16
Return an explicit 500 if there's an exception generating metrics

FauxFaux:
The current behaviour observed in #84 is pretty bad. This implements the simplest possible fix: if there's any exception of any kind generating metrics, log it as we are now, but change the response code to a 500.

`send_error` is used, which dumps a random bit of html into the response, which doesn't seem like a problem. `send_response(500)` could be used instead, to get a 500 with a blank body.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 
+import logging
 import os
 import socket
 import time
@@ -70,10 +71,18 @@ def generate_latest(registry=core.REGISTRY):
 
 class MetricsHandler(BaseHTTPRequestHandler):
     def do_GET(self):
+        try:
+            content = generate_latest(core.REGISTRY)
+        except:
+            logging.exception('error generating response')
+            self.send_error(500, 'error generating metrics')
+            self.end_headers()
+            return
+
         self.send_response(200)
         self.send_header('Content-Type', CONTENT_TYPE_LATEST)
         self.end_headers()
-        self.wfile.write(generate_latest(core.REGISTRY))
+        self.wfile.write(content)
 
     def log_message(self, format, *args):
         return


prometheus_client/exposition.py
```diff
@@ -70,10 +71,18 @@ def generate_latest(registry=core.REGISTRY):
 
 class MetricsHandler(BaseHTTPRequestHandler):
     def do_GET(self):
+        try:
+            content = generate_latest(core.REGISTRY)
```
brian-brazil: I'd suggest removing the try and letting the exception propagate up.

FauxFaux: That causes an empty body to be returned, not a 500. I felt that a 500 was more useful than just closing the tcp connection (although I suspect Prometheus would handle the case in the same way).

```diff
     def do_GET(self):
-        try:
-            content = generate_latest(core.REGISTRY)
-        except:
-            logging.exception('error generating response')
-            self.send_error(500, 'error generating metrics')
-            self.end_headers()
-            return
-
+        content = generate_latest(core.REGISTRY)
         self.send_response(200)
```

```
% curl -v http://localhost:23456/
*   Trying ::1...
* connect to ::1 port 23456 failed: Connection refused
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 23456 (#0)
> GET / HTTP/1.1
> Host: localhost:23456
> User-Agent: curl/7.47.0
> Accept: */*
> 
* Empty reply from server
* Connection #0 to host localhost left intact
curl: (52) Empty reply from server
```

brian-brazil: Prometheus would handle it the same. If we're going to explicitly return a 500 we should also include the full exception in the response.

FauxFaux: Many security policies discourage returning the stack-trace (or any details about the actual problem) in the response from 500 errors by default. Maybe you could argue that people typically authenticate these metrics endpoints, so it's not a problem? I don't, so would rather the stack-trace wasn't returned. If you decide to put it in, it will just be stripped by our load balancers anyway, so I don't mind either way.

brian-brazil: I would generally expect metrics endpoints not to be exposed to the outside world, and even then for the information you can gleam from the errors to be minimal.

prometheus_client/exposition.py
```diff
@@ -70,10 +71,18 @@ def generate_latest(registry=core.REGISTRY):
...
+        except:
+            logging.exception('error generating response')
```
ulope: Please don't use the "convenience" functions directly on the logging module. They log to the root logger making it impossible to filter (or even redirect) these messages. Instead use a module level logger like so:

```
import logging

log = logging.getLogger(__name__)

# ... code ...

log.exception("whatever")

```

brian-brazil:
Obsoleted by #119

PR #86, opened 2016-06-17
Add CI, make tests pass on non-linux

hynek:
This is mainly to remove friction from contributing.

A couple of notes:

- running tests over all python versions is now a matter of running `tox`
- you’ll probably have to activate the project in Travis
- it doesn’t run tests on 2.6 because…the tests currently fail on it.
- it uses py.test to have a consistent test runner.  It doesn’t use any of its features tho.  Test discovery via `python -m unittest` works differently across python versions.  I could also use nose or something but I chose pytest as it’s more familiar to me.
- The odd-at-first-look coverage construction allows for measuring coverage over [multiple Python versions](https://hynek.me/articles/testing-packaging/).
- I’d recommend to add some more targets but let’s start simple.

Let me know if you have any questions or concerns.


tests/test_exposition.py
```diff
@@ -160,6 +161,10 @@ def test_delete_with_groupingkey(self):
         self.assertEqual(self.requests[0][0].headers.get('content-type'), CONTENT_TYPE_LATEST)
         self.assertEqual(self.requests[0][1], b'')
 
+    @unittest.skipUnless(
+        sys.platform.startswith("linux"),
+        "instance_ip_grouping_key() only works on linux."
```
brian-brazil: I think it's only OS X that's broken.

brian-brazil:
> it doesn’t run tests on 2.6 because…the tests currently fail on it.

I'd like to test 2.6. That looks to be a mix of a dependency issue, and not having the improved float printing semantics for histogram buckets. Could the first be fixed, and the second excluded from testing?

hynek:
While I think that 2.6 is terrible and nobody should use it, I would suggest to put it in a separate PR.

hynek:
r4r btw

brian-brazil:
CentOS 6 is still supported and provides python 2.6, we have to work with what our users have not what we wish they have.

I'd not have separate tests for twisted, there's be various other optional deps over time so that won't scale well.

hynek:
I’m not sure how your comment relates to my suggestion to put it into a separate PR?

hynek:
As for the second part, you want to have code paths that are not covered by CI on purpose?  How do you want to ensure long-term that stuff doesn’t break?

brian-brazil:
The code paths would only be not covered on Python 2.6, the chances of them breaking only there is pretty low as that particular piece of code is stable.

hynek:
Your answer kind of mixes my two questions. :)

Let me try again:

1. do you insist on resolving Python 2.6 within *this* PR or is it OK to put it in a separate one?
2. do you really want to keep the Twisted code from CI?

brian-brazil:
It's ok to have in a separate one.

I'm thinking more that a separate run just for twisted might be a bit much, and it should be combined into the main one.

hynek:
Alright!

So I would suggest the following: let twisted be part of regular run but have *one* `py35-notwisted` that ensures, that prometheus_client keeps working if Twisted is *not* installed.

Simply because it's really easy to screw up by importing something somewhere with unintended consequences.

hynek:
Since your main concern was, that the list of deps is gonna grow, I ended up calling the env `py35-nooptionals`.

Let me know what you think.

brian-brazil:
That looks generally fine, can you add a nooptions on 2.7 too so both major versions are covered?

hynek:
sure; done.

brian-brazil:
Can you squash your commits please?

hynek:
Done.

I’ve activated Travis for my fork and you can see the latest build here btw: https://travis-ci.org/hynek/client_python/builds/143289122

brian-brazil:
Thanks!

PR #87, opened 2016-07-03
Allow passing label values as keyword arguments

alex:


brian-brazil:
This is getting a bit complicated. I think that if we're going to this this then the ability to pass a dict as the first arg should be deprecated. I think there's only one user of that out there.

alex:
I have no objection to that; I find the "raw dict" UX a bit unpleasant myself (hence this PR :-)). Is there a precedent for how deprecations are handled in this library?

brian-brazil:
This would be the first deprecation. Probably best to reach out to users and see if they can take a straight breakage.

alex:
Is there a particular place to do that?

brian-brazil:
You could hit the mailing list, or ping @korfuri

alex:
Sorry it took me a while, mailing list messaged: https://groups.google.com/forum/#!topic/prometheus-developers/KTAvr3DdVi4

alex:
@brian-brazil Been about 5 days and only one person has chimed in. Shall I update this PR to remove the dict-passing API?

brian-brazil:
Sounds like a plan

alex:
Great; done!

README.md
```diff
@@ -185,7 +185,16 @@ c.labels('get', '/').inc()
...
+Labels can also be passed as keyword-arguments:
+
+```python
+from prometheus_client import Counter
+c = Counter('my_requests_total', 'HTTP Failures', ['method', 'endpoint'])
+c.labels(method='get', endpoint='/').inc()
+c.labels(method='post', endpoint='/submit').inc()
+```
+
+Or as a dict:
```
brian-brazil: The example here should be removed

alex: You're looking at a previous version of the diff, it's removed on the latest.

brian-brazil:
Thanks!

alex:
Thank you!

PR #89, opened 2016-07-08
Make tests pass on Python 2.6

hynek:
As discussed in #86, this adds Python 2.6 support for the tests and CI.

Since 2.6’s stdlib doesn’t support test skipping, we have to resort to using the back port unittest2.

There are a few ways to make it less ugly:

- use unittest2 unconditionally everywhere.  has also the benefit of a consistent test environment.
- do the dance once in a helper module (say `tests/utils.py`) and then just import unittest from there.
- use pytest for skipping

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ language: python
 
 matrix:
   include:
+    - python: "2.6"
+      env: TOXENV=py26
     - python: "2.7"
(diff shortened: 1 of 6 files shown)


brian-brazil:
Having anyone on 2.6 have to install an extra module for testing seems fine to me.

Thanks!

Issue #77, opened 2016-04-14
Decorators break function signatures


hynek:
The client does a common anti-pattern of doing a closure with `*args,**kwargs` but that sadly breaks introspection of wrapped methods.

This is a concrete problem e.g. if you wrap Pyramid views that get different arguments depending on how many arguments are present (either `context, request` or just `request`).

You can find more information on this in these blog posts: https://github.com/GrahamDumpleton/wrapt/tree/master/blog

Solutions to it is to either use or copy/paste from the [wrapt](http://wrapt.readthedocs.org/en/latest/) (better) or [decorator](http://pythonhosted.org/decorator/documentation.html) (OK-ish) packages.

brian-brazil:
Would you like to send a pull request to fix this?

wrapt appears to have C dependencies, so that's not going to work out.

hynek:
So you could live with depending on decorator?  It’s python-only without further dependencies.

I still have the problem that I can’t run the tests locally on OS X because of the `s.connect(('localhost', 0))` thing.  If you’d fix that, I’ll happily fix your decorators. :)

brian-brazil:
> So you could live with depending on decorator? It’s python-only without further dependencies.

I'd prefer to copy it in. While it's old enough that it should be everywhere, let's play it safe.

hynek:
As you wish.

I still need you to fix the test suite tho. :)

brian-brazil:
If you've got an alternative way to get the machines IP that also works on Windows and Linux then I'm happy to change it.

hynek:
I‘m fine with if you just skip the test on OS X (and mostly likely all BSDs? at least FreeBSD breaks too).

I just need a way to run tests locally.  Once that’s fixed, I can do the tox/Travis CI/codecov integration for you too if you wish.  By not being able to run the test suite on all BSDs you’re limiting the set of contributors quite a bit.

PR #91, opened 2016-07-20
Bundle decorator.py and fix decorators

hynek:
As discussed in #77, this PR bundles (to avoid dependencies) the decorator.py module and uses it to fix the behavior of prometheus_client's decorators.

The efficacy can be verified by using example view from the README and then doing a:

```pycon
>>> from inspect import getargspec
>>> getargspec(process_request)
ArgSpec(args=['t'], varargs=None, keywords=None, defaults=None)
```

Without this patch, you get:

```pycon
ArgSpec(args=[], varargs='args', keywords='kwargs', defaults=None)
```

instead.

Let me now what you think.

Also pinging @micheles out of courtesy and to ensure he doesn’t feel we violate his copyright.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,8 @@
 [run]
 branch = True
 source = prometheus_client
+omit =
+    prometheus_client/decorator.py
 
 [paths]
 source =
--- a/NOTICE
+++ b/NOTICE
@@ -1,2 +1,5 @@
 Prometheus instrumentation library for Python applications
 Copyright 2015 The Prometheus Authors
+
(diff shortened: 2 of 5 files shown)


brian-brazil:
That looks good, could you add a quick unittest so we don't accidentally regress?

micheles:
Yes, no problems on my part, the decorator module is a single file just to
help people that wants to include it and have less dependencies.

hynek:
I’ve added unit tests for hopefully all decorators and the tests still pass.

brian-brazil:
Thanks!

PR #93, opened 2016-08-03
Clean up the ProcessCollector.

vad:
Code:
- remove unused imports
- remove a useless try/except/raise that wraps a os.path.join()

Style:
- max line length 119
- use 4 spaces for indentation
- fix indentation of continuation lines

### DIFF
--- a/prometheus_client/process_collector.py
+++ b/prometheus_client/process_collector.py
@@ -3,16 +3,14 @@
 from __future__ import unicode_literals
 
 import os
-import time
-import threading
 
 from . import core
 try:
-  import resource
-  _PAGESIZE = resource.getpagesize()
+    import resource
+    _PAGESIZE = resource.getpagesize()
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks. Would you mind adding a small unittest for the missing file?

vad:
So you confirm that "return silently" is the behaviour you expect?

vad:
Ok, I'll write the test ASAP, probably today

vad:
Ok, in the end I looked at this only today and reverted to the old behaviour. The try/except was not around open() but around os.path.join(). IMHO it makes totally sense to raise and exception in this case, because it's probably an error in self._pid()

brian-brazil:
Can you squash your commits and update the description?

vad:
@brian-brazil is it ok?

brian-brazil:
Thanks!

PR #94, opened 2016-08-12
Fix labels documentation

neuroid:
Just a small README fix to reflect reality.

@brian-brazil


neuroid:
Oops. I just noticed this was changed recently :)

Issue #76, opened 2016-04-11
stop_http_server()


tlc:
There is no way to stop the metrics server if you need to reconfigure (change the port).

tlc:
Returning the thread, so the application could manage it would be sufficient.

brian-brazil:
Hmm, I'd have to think on this a bit. We probably shouldn't expose the internals of how we're implementing this to the user.

This is intended mostly as a convenience function, there is the option of having your own variant of it if you've different needs.

rud:
@tlc issue ready to close then?

eddieparker:
In case it helps others - I came here looking for a similar solution.  Here's a quick working example to crib off of:

```
# A wrapper around the python http server and prometheus MetricsHander to allow us to stop the server
class CustomPrometheusHttpServer:
	def __init__(self, port):
		self.port = port
		self.httpd = None
		self.thread = None

	def start(self):
		self.httpd = http.server.HTTPServer(('0.0.0.0', self.port), prometheus_client.exposition.MetricsHandler)
		self.thread = threading.Thread(target=self.httpd.serve_forever)
		self.thread.daemon = True
		self.thread.start()

	def stop(self):
		self.httpd.shutdown()
		self.httpd.server_close()
		self.thread.join()
```

engin-ipek:
> self.httpd.shutdown()
> 		self.httpd.server_close()

hello, since `start_http_server` returns the server and the thread, why can't we do this instead?:

````python
httpd, t = start_http_server()
time.sleep(30)
httpd.shutdown()
httpd.server_close()
t.join()

PR #102, opened 2016-09-27
start_http_server: start HTTPServer in main thread before handing off to daemon thread

rud:
This means if you call `start_http_server` with an already used port/addr combination, you can rescue the `OSError`, and try for a different port. With this, it becomes possible to probe a range of ports to find one that is available.

Idea for this comes from https://github.com/korfuri/django-prometheus/blob/2b6eac500cc9bea402a45f04ca7b63189889785a/django_prometheus/exports.py#L77-L88 - which makes it easy to let each uwsgi worker listen on a port by automatically trying a whole range of ports and picking one that works.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -82,10 +82,15 @@ class MetricsHandler(BaseHTTPRequestHandler):
 def start_http_server(port, addr=''):
     """Starts a HTTP server for prometheus metrics as a daemon thread."""
     class PrometheusMetricsServer(threading.Thread):
+        def __init__(self, httpd):
+            self.httpd = httpd
+            super(PrometheusMetricsServer, self).__init__()
+
         def run(self):
-            httpd = HTTPServer((addr, port), MetricsHandler)
-            httpd.serve_forever()
-    t = PrometheusMetricsServer()
+            self.httpd.serve_forever()
+
+    httpd = HTTPServer((addr, port), MetricsHandler)
+    t = PrometheusMetricsServer(httpd)
     t.daemon = True
     t.start()
 


rud:
A thing worth discussing: it would be entirely possible to rescue the `OSError` and just return a boolean to indicate whether starting the listener succeeded. Given that the previous behaviour was to silently not serve metrics for the current process (instead of halting with an exception like this introduces), a smaller change would be to add a rescue here - and existing setups with duplicate metrics ports would still fail silently.

Thoughts?

brian-brazil:
There's also been requests to be able to stop the httpserver, so it'd be best to consider that with this change.

rud:
Hi @brian-brazil,

As I see it, one option would be to return the handle of the daemon thread from the `start_http_server` method, or save it off somewhere for future reference. Would that be a handy enough API? Return value could be the thread-handle on successful service start, `None` on startup failure. What do you think?

rud:
I see, you probably refer to https://github.com/prometheus/client_python/issues/76 - but as I read that issue, the user ended up finding a different way to start/stop their listener?

rud:
@brian-brazil fwiw, I've created code in my project to just reuse the `prometheus_client.MetricsHandler` to setup my own HTTP listener just like I want it (with automatic probing of a number of ports until a free one is found). 

Thank you for making the `prometheus_client.MetricsHandler` reusable externally, I can only imagine how many variations people need for running their services. 

Feel free to close this pull-request if the design is not something you can use going forward.

brian-brazil:
The PR as-is is  good idea, the question is more around rehandling exceptions. The Pythonic way is to just let them be thrown.

> (with automatic probing of a number of ports until a free one is found).

That smells a bit, your config management should be telling you what port number to use.

> Thank you for making the prometheus_client.MetricsHandler reusable externally, I can only imagine how many variations people need for running their services.

That's the idea.

rud:
I concur with the port probing being potentially problematic, but given that a uwsgi master process spawns a number of child processes at various times, it seems to make sense they each pick a port in a known range where they make metrics available. I'm using the Telegraf daemon to connect to each of these ports and collect the available metrics. There is a simplicity to this I like - and since to the best of my knowledge each uwsgi worker-process should not care too much about its individual identity/number in the flock, it also makes sense that it cannot have a distinct static prometheus listener port.

rud:
So, any specific changes you'd like to see to this code at this time?

brian-brazil:
Sounds like you should be reading #66, that's not a safe way to do multi-process.

rud:
Thank you for your suggestion. 

In #66 I see a way of having workers report their metrics up the process tree to the master process, but it is currently a work in progress. I do not see any discussion of safety properties that preclude individual workers from exposing their own metrics directly, but it may very well be too implicit for me to see it. Is it related to the point in the startup process where workers are spawned, that some internal structure may or may not be correctly setup yet? 

If this is veering off-topic I do apologise.

brian-brazil:
The issue is that there might be state in a dead worker that you want to preserve past its demise, and that correctly collapsing per-process counters requires data from dead workers too.

rud:
Agreed, that would indeed be loss of recent state in this design. And since the uwsgi master process will much too happily `kill -9` workers that are past their prime - dataloss will be guaranteed in that case if the collector did not sweep in soon enough.

To me that comes to the trade off of sending metrics somewhere external immediately while handling an incoming request (to guarantee survivability of metrics, but adding latency), or to buffer up measurements within each worker-process for a configurable short time, and gather the values in bulk every X seconds. All in-memory queueing means the potential for dataloss, but at much greater throughput. Adding additional communication complexity to the master/worker hierarchy as #66 does seems like it might be a good tradeoff, but things do become more difficult to reason about.

I'll close this merge-request now, as I think I have a grasp of the tradeoffs I'm making, and I think you are right the solution provided here is not generally applicable.

PR #106, opened 2016-10-03
Use timeit.default_timer instead of time.time

acdha:
On Python 3, this uses `time.perf_counter` on all platforms.

On Python 2, it will use `time.clock` on Windows and `time.time` everywhere else

### DIFF
--- a/prometheus_client/bridge/graphite.py
+++ b/prometheus_client/bridge/graphite.py
@@ -6,6 +6,7 @@ import re
 import socket
 import time
 import threading
+from timeit import default_timer
 
 from .. import core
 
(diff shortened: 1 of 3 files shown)


brian-brazil:
Thanks!

melnikk:
Just FYI, this leads to short timestamp on my system (Centos 7.3, python 3.4.5):
```
process_resident_memory_bytes 30285824.0 160435
```
Graphite drops that metrics, and, as a workaround, I had to switch back from default_timer:
```
gb = GraphiteBridge((GRAPHITE_HOST, GRAPHITE_PORT), _timer=time.time)
```

brian-brazil:
That would be a bug, would you like to send a PR switching that one back to time.time?

melnikk:
@brian-brazil, sure, PR #175

PR #107, opened 2016-10-13
Google AppEngine Compatibility

mnadel:
The `mmap` package is being imported at the top-level, even if we're not opting for multiprocess support. This causes failures in Google AppEngine, where `mmap` is blacklisted.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -5,7 +5,6 @@ from __future__ import unicode_literals
 import copy
 import json
 import math
-import mmap
 import os
 import re
 import struct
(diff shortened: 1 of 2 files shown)


prometheus_client/core.py
```diff
@@ -5,13 +5,22 @@
...
 import json
 import math
-import mmap
 import os
 import re
 import struct
 import types
 from timeit import default_timer
 
+PROM_MULTIPROC_DIR='prometheus_multiproc_dir'
```
brian-brazil: It'd seem cleaner to lookup the environment here

prometheus_client/core.py
```diff
@@ -5,13 +5,22 @@
...
+
+try:
+    # this may not exist (e.g. in Google AppEngine)
```
brian-brazil: Comments should start with a capital letter, and end with a full stop

mnadel:
Renaming the PR to the more general topic, Google AppEngine Compatibility; I also needed to make a change to the Process Collector to allow it to fail gracefully on GAE.

prometheus_client/core.py
mnadel: Reverted this change and took a different approach, thanks.

mnadel:
How's this look @brian-brazil? It's a less intrusive change. Now mmap is lazily loaded, and only if multiprocess support is enabled. Also, it handles an additional failure mode when running in GAE for the process collector.

prometheus_client/core.py
```diff
@@ -253,12 +252,13 @@ class _MmapedDict(object):
     alignment, and then a 8 byte float which is the value.
     """
     def __init__(self, filename):
+        self.mmap = __import__('mmap')
```
brian-brazil: Can't non-top-level imports cause deadlocks or other problems?

PR #109, opened 2016-10-17
Use realistic timer for Gauge.set_to_current_time

hynek:
set_to_current_time is expected to set the value to a realistic time.
Performance counters are monotonic but don't say anything about the actual
time.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -9,8 +9,8 @@ import mmap
 import os
 import re
 import struct
+import time
 import types
-from timeit import default_timer
 
 try:
     from BaseHTTPServer import BaseHTTPRequestHandler
@@ -19,6 +19,7 @@ except ImportError:
     unicode = str
 
 from threading import Lock
+from timeit import default_timer
 
 from .decorator import decorate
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
My reading of the docs is that this is wall time.

hynek:
What exactly? You docs say that it sets to current unix time which I thought to be epoch time.  However:

```pycon
>>> from timeit import default_timer
>>> from time import time
>>> from datetime import datetime
>>> datetime.fromtimestamp(default_timer())
datetime.datetime(1970, 1, 5, 19, 25, 4, 943824)

>>> datetime.fromtimestamp(time())
datetime.datetime(2016, 10, 17, 19, 50, 9, 261461)
```

This seems rather useless to me and a pretty bad regression IMHO (got a lot of `time()-metric > X` alerts :)).

hynek:
Oh i think I understand now; you’ve been referring to legacy (2.7) docs.  default_timer is `time.perf_counter()` as of Python 3.3: https://docs.python.org/3.5/library/timeit.html#timeit.default_timer

brian-brazil:
Thanks!

hynek:
It would be really cool if you could push a quick new release, since this is a nasty regression on Python 3.3+ and will probably also confuse new users.

PR #110, opened 2016-11-03
Add describe and limiting of http expostion

### LINKED ISSUE #97: can generate malformed data by creating a duplicate Counter
I accidentally generated malformed data by creating two Counters on the same Registry that had the same name, help text.

Example code:

```
import prometheus_client

reg = prometheus_client.CollectorRegistry()
c1 = prometheus_client.Counter("foo", "bar", labelnames=["baz", "stuff"], registry=reg)
c1.labels({"baz": 1, "stuff": 2}).inc(1)
c2 = prometheus_client.Counter("foo", "bar", labelnames=["baz", "stuff"], registry=reg)
c2.labels({"baz": 3, "stuff": 4}).inc(2)
prometheus_client.push_to_gateway('localhost:8000', job='example', registry=reg)
```

Data sent in request to push gateway:

```
# HELP foo bar
# TYPE foo counter
foo{baz="1",stuff="2"} 1.0
# HELP foo bar
# TYPE foo counter
foo{baz="3",stuff="4"} 2.0
```

Response sent from the push gateway:

```
text format parsing error in line 4: second HELP line for metric name "foo"
```

I'd prefer either getting an error from prometheus_client telling me that I can't add two identical metrics.  Alternatively, I'd like for prometheus_client to merge the two metrics and not send two identical HELP lines.

brian-brazil:
This is the first pass on limiting what's returned over http exposition via a url parameter. We'll likely need to support POST also in future to get around URL length limits.

@juliusv @fabxc @beorn7 @SuperQ 

Fixes #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,6 +365,19 @@ REGISTRY.register(CustomCollector())
 
 `SummaryMetricFamily` and `HistogramMetricFamily` work similarly.
 
+A collector may implement a `describe` method which returns metrics in the same
+format as `collect` (though you don't have to include the samples). This is
+used to predetermine the names of time series a `CollectorRegistry` exposes and
+thus to detect collisions and duplicate registrations.
+
+Usually custom collectors do not have to implement `describe`. If `describe` is
+not implemented and the CollectorRegistry was created with `auto_desribe=True`
+(which is the case for the default registry) then `collect` will be called at
+registration time instead of `describe`. If this could cause problems, either
+implement a proper `describe`, or if that's not practical have `describe`
+return an empty list.
+
+
 ## Multiprocess Mode (Gunicorn)
 
 **Experimental: This feature is new and has rough edges.**
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -97,6 +97,35 @@ class CollectorRegistry(object):
             for metric in collector.collect():
                 yield metric
 
+    def restricted_registry(self, names):
+        '''Returns object that only collects some metrics.
+
+        Returns an object which upon collect() will return
+        only samples with the given names.
+
+        Intended usage is:
+            generate_latest(REGISTRY.restricted_registry(['a_timeseries']))
+
+        Experimental.'''
+        names = set(names)
+        collectors = set()
+        with self._lock:
+            for name in names:
+                if name in self._names_to_collectors:
+                    collectors.add(self._names_to_collectors[name])
+        metrics = []
+        for collector in collectors:
+            for metric in collector.collect():
+                samples = [s for s in metric.samples if s[0] in names]
+                if samples:
+                    m = Metric(metric.name, metric.documentation, metric.type)
+                    m.samples = samples
+                    metrics.append(m)
+        class RestrictedRegistry(object):
+            def collect(self):
+                return metrics
+        return RestrictedRegistry()
+
     def get_sample_value(self, name, labels=None):
         '''Returns the sample value, or None if not found.
 
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -38,25 +38,61 @@ class CollectorRegistry(object):
     Metric objects. The returned metrics should be consistent with the Prometheus
     exposition formats.
     '''
-    def __init__(self):
-        self._collectors = set()
+    def __init__(self, auto_describe=False):
+        self._collector_to_names = {}
+        self._names_to_collectors = {}
+        self._auto_describe = auto_describe
         self._lock = Lock()
 
     def register(self, collector):
         '''Add a collector to the registry.'''
         with self._lock:
-            self._collectors.add(collector)
+            names = self._get_names(collector)
+            for name in names:
+                if name in self._names_to_collectors:
+                    raise ValueError('Timeseries already present '
+                            'in CollectorRegistry: ' + name)
+            for name in names:
+                self._names_to_collectors[name] = collector
+            self._collector_to_names[collector] = names
 
     def unregister(self, collector):
         '''Remove a collector from the registry.'''
         with self._lock:
-            self._collectors.remove(collector)
+            for name in self._collector_to_names[collector]:
+                del self._names_to_collectors[name]
+            del self._collector_to_names[collector]
+
+    def _get_names(self, collector):
+        '''Get names of timeseries the collector produces.'''
+        desc_func = None
+        # If there's a describe function, use it.
+        try:
+            desc_func = collector.describe
+        except AttributeError:
+            pass
+        # Otherwise, if auto describe is enabled use the collect function.
+        if not desc_func and self._auto_describe:
+            desc_func = collector.collect
+
+        if not desc_func:
+            return []
+
+        result = []
+        type_suffixes = {
+            'summary': ['', '_sum', '_count'],
+            'histogram': ['_bucket', '_sum', '_count']
+        }
+        for metric in desc_func():
+            for suffix in type_suffixes.get(metric.type, ['']):
+                result.append(metric.name + suffix)
+        return result
 
     def collect(self):
         '''Yields metrics from the collectors in the registry.'''
         collectors = None
         with self._lock:
-            collectors = copy.copy(self._collectors)
+            collectors = copy.copy(self._collector_to_names)
         for collector in collectors:
             for metric in collector.collect():
                 yield metric
@@ -75,7 +111,7 @@ class CollectorRegistry(object):
         return None
 
 
-REGISTRY = CollectorRegistry()
+REGISTRY = CollectorRegistry(auto_describe=True)
 '''The default registry.'''
 
 _METRIC_TYPES = ('counter', 'gauge', 'summary', 'histogram', 'untyped')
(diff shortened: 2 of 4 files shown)


prometheus_client/core.py
```diff
@@ -38,29 +38,94 @@ class CollectorRegistry(object):
...
+            collectors = copy.copy(self._collector_to_names)
         for collector in collectors:
             for metric in collector.collect():
                 yield metric
 
+    def restricted_registry(self, names):
+        '''Returns object that only collects some metrics.
+
+        Returns an object which upon collect() will return
+        only samples with the given names.
```
beorn7: Wasn't the plan to allow regexp matches for names, too? And inverted matches? (i.e. collect all metrics except those matching 'foo.*')?

README.md
```diff
@@ -365,6 +365,19 @@ REGISTRY.register(CustomCollector())
...
+format as `collect` (though you don't have to include the samples). This is
+used to predetermine the names of time series a `CollectorRegistry` exposes and
+thus to detect collisions and duplicate registrations.
+
+Usually custom collectors do not have to implement `describe`. If `describe` is
+not implemented and the CollectorRegistry was created with `auto_desribe=True`
+(which is the case for the default registry) then `collect` will be called at
+registration time instead of `describe`. If this could cause problems, either
+implement a proper `describe`, or if that's not practical have `describe`
+return an empty list.
```
beorn7: Is the intention that in the latter case, collectors returning an empty list will never be included by a restricted_registry?

prometheus_client/core.py
brian-brazil: This is v1, so I didn't add the regexes yet.

Inverted wasn't something I'd planned.

README.md
brian-brazil: Yes. My thinking is that for something like JMX which can be quite expensive, it's best not to call it rather than silently calling it in the background when users thing they're only fetching a few cheap counters.

prometheus_client/core.py
beorn7: I would imagine it's a common use case: You want a scrape as "normal" as possible, just that one expensive metric with high cardinality should go.

brian-brazil: I'm presuming that that'd be handled on the Prometheus side with metric relabelling.

README.md
beorn7: While I understand the resource argument, it's opaque to the user of an `/metrics` endpoint which collectors describe themselves properly, and thus the behavior might appear quite puzzling. Between "accidentally using a lot of resources" and "accidentally and silently not getting expected metrics" I'd prefer the former.

prometheus_client/core.py
beorn7: Which would not avoid collecting the metric (which might be expensive) and sending it over the wire (which might be bandwidth limited).

README.md
brian-brazil: This expected use case of this is high-frequency polling of a very small number of metrics. Using a lot of resources in such a case would be bad.

prometheus_client/core.py
beorn7: I don't understand. If this is the case, why are we doing client-side restrictions (like in this PR) at all?

README.md
beorn7: But I would notice – because it would defeat the purpose of restricting the metrics.

If some metrics are silently missing, I might not notice. I'm imagining the Postmortem review right now...

prometheus_client/core.py
brian-brazil: The cost has never been relevant when dealing with a high-cardinality metric for normal scraping.

It has been relevant when attempting to do high-frequency scraping of a very limited set of metrics.

README.md
brian-brazil: You'd be specifying a small number of names by hand so you'd likely notice. 

The logic here is that calling `collect()` at registration time is problematic, so we should also be tending not to call it when the collector has asked us not to include it in this feature.

prometheus_client/core.py
beorn7: Perhaps we are missing each others point here. I try to rephrase:

- I wish to do high frequency scraping.
- I have identified a single metric that is expensive for any of the following reasons:
  - It comprises 80% of the scrape volume, and I am bandwidth limited.
  - It is expensive to `collect()`.
- Let's assume for the sake of your argument that it's usually the latter and not the former that matters. I could of course hand-pick and whitelist the metrics I'm interested in. But if I can do my high frequency scraping without problem by just blacklisting the one metric known to be expensive to `collect()`?

IIRC the mysqld_exporter has exactly that case. So what we do right now is to switch certain metrics on and off on the command line. At SC, we run separate high-frequency collectors with fewer metrics activated. It would be great if we could just tell the collector via URL params to not collect selected few expensive-to-collect metrics for high-frequency scraping rather than running separate exporter instances with different command line flags.

If I understand you correctly, you are claiming that the high-frequency scrape case will naturally result in scraping a whitelisted set of metrics. But if I understand the mysqld_exporter use case correctly, it's the opposite: One tends to blacklist known expensive metrics and wants to collect all others with high frequency. @SuperQ please correct me if I'm wrong.

README.md
beorn7: OK, after some background processing, I think I got your last point now. You assumption is essentially that collectors that have problems describing themselves will also be those that are expensive to `collect()`.

That's true for the JMX case, I guess, but not in general.

I see the value of allowing collectors to say "I cannot describe myself but I'm also expensive to collect so please never collect me if the user has restricted the scrape in any way". But I also see the potential for bad surprises, and I guess the feeling is so strong for me because I see the "blacklist a few metrics" as a common use case. If suddenly some seemingly unrelated metrics disappear just because I blacklisted a single one, feels weird.

beorn7: I would hope that non-describing collectors will be rare. In direct instrumentation, they should never occur, essentially. It's mostly an exporter thing, so we should encourage clear documentation about metrics not being collected if any restriction is given as a scrape parameter.

prometheus_client/core.py
brian-brazil: > If I understand you correctly, you are claiming that the high-frequency scrape case will naturally result in scraping a whitelisted set of metrics.

Yes. Given the data volumes involved with such systems no other way is practical, and I've never seen any other approach taken. Usually you'd be whitelisting on the order of 10 metrics, and with a blacklisting approach the organic growth of metrics would be sufficient to take out such a system.

To be honest, if you've a system where one metric comprises 80% of data volume you've bigger problems (I came across that exact case once, and greatly decreased the metric cardinality).

> But if I understand the mysqld_exporter use case correctly, it's the opposite: One tends to blacklist known expensive metrics and wants to collect all others with high frequency. 

That's not my understanding. The mysqld exporter isn't covered by this anyway, it's sufficiently weird that it can't implement describe.

README.md
brian-brazil: > You assumption is essentially that collectors that have problems describing themselves will also be those that are expensive to collect().

Yes, or impossible to describe(). JMX is both impossible and expensive. A one-off collect() at startup would be tolerable (ignoring that that breaks with config reloading), but at every collection would be bad.

> If suddenly some seemingly unrelated metrics disappear just because I blacklisted a single one, feels weird.

The whitelist approach doesn't really have that issue. Returning an empty describe signals that you don't support this feature.

> It's mostly an exporter thing, so we should encourage clear documentation about metrics not being collected if any restriction is given as a scrape parameter.

Yeah, this only comes up with custom collectors.

beorn7:
OK, I'll try to summarize the threads above here:

- Collectors that are unable to describe themselves must return an empty description and will be excluded from any restricted collection. (I concur that there is no better solution in the current state of affairs.)
- There remains disagreement if a whitelisting-only approach is sufficient, or if blacklisting AKA negative matching should be part of the API. Input from others welcome.

brian-brazil:
That sounds about right.

Do you have any comments on the spelling of the HTTP API, or thoughts on the restricted_registry? I think the former is fine, but I'm not perfectly happy with the latter.

beorn7:
@SuperQ could you comment on our current use of the hires MySQL exporter? Is that just a few selected metrics we scrape for it, or is it the other way round, i.e. we just excluded the expensive metrics?

I've just looked at our actual setup. Normal MySQL exporter is about 6000 data points per scrape, the hires exporter only about 1000. Still would be good to get @SuperQ 's idea of what approach would be preferable in that real-life scenario.

WRT the HTTP API: I would model it after the federation API, i.e. use `match[]` parameters. But the arguments are not PromQL selectors but just regexps for the metric name. Contract would be:
- Non-supporting targets ignore the pramater.
- Supporting targets will not collect from any collector that cannot describe itself if at least one `match[]` parameter is given. From all other collectors, they will return all metrics whose name matches any of the regexps given.

SuperQ:
The hires case we select for a few specific metrics that we care about.  Some short spiky gauges, some high-churn counters.  With MySQL exporter, we need to know ahead of time, at the scrape request, which collector functions we need to call.  This is to avoid accessing expensive or locking collectors.

So I guess it's a little of both.

brian-brazil:
What I'm thinking will happen with mysqld_exporter is that it'd have URL parameters like `collector[]`, following the pattern set here and on /federate as collector selection is the important thing. For the JMX exporter (if it ever happened) it'd probably be `mbean[]` or similar.

beorn7:
Which would mean we would go for a pretty customized behavior anyway. Without the necessity (or ability) to define a universally used API, I have no problem with deferring negative match until somebody shows up with a use case.

brian-brazil:
Sounds like we're in rough agreement then. Any other comments?

brian-brazil:
Final call, any comments on this? I intend to submit tomorrow.

PR #116, opened 2016-11-11
Fix #115

### LINKED ISSUE #115: Multiproc files not cleaned up if worker gets segfault/oom/...
In multiproc mode, if one of the gunicorn workers gets killed because of a segfault or out-of-memory situation, it doesn't get a chance to clean up its `.db` files. These files are thus left in the multiproc folder, distorting the metrics.

Maybe this can be solved by moving the cleanup hook into the gunicorn "master" process. My suggestion is to solve this in the Python client instead, by deleting all `.db` files with dead PIDs before aggregating the metrics.

jonashaag:
First draft for fixing #115.

I'll add/fix tests and documentation as soon as the general idea has been approved.

### DIFF
--- a/prometheus_client/multiprocess.py
+++ b/prometheus_client/multiprocess.py
@@ -7,6 +7,8 @@ import json
 import os
 import shelve
 
+import psutil
+
 from . import core
 
 class MultiProcessCollector(object):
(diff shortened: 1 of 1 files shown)


brian-brazil:
As discussed, this needs to be handled elsewhere which I believe you've done.

PR #123, opened 2016-12-09
Rename extras_requires to extras_require

mrueg:


brian-brazil:
Can you expand on the reason for this change?

mrueg:
There's no keyword called extras_requires in setuptools. See http://setuptools.readthedocs.io/en/latest/setuptools.html

brian-brazil:
Thanks!

That's wonderfully inconsistent.

mrueg:
Well extras is plural, so it makes sense to drop the s. Still I agree for consistency reasons it would have been a better choice to call it extra_requires

PR #120, opened 2016-11-25
Add handler parameter to pushgateway functions to permit HTTP AUTH etc.

tim-seoss:
First stab at a handler implementation.  No tests yet, please let me know what you think.  Documenting a BASIC AUTH handler would probably be a good idea as per #60

Issue #60, opened 2015-10-15
Please add support for pushgateways protected with basic authentication


kkoppel:
Currently there doesn't seem to be any way to provide basic authentication credentials for talking to pushgateways (urllib2 doesn't allow putting them in the URL). Would be great if this could be added.

kkoppel:
Just noticed that the URL is hardcoded to begin with "http://", so ideally HTTPS should be supported as well.

brian-brazil:
I'll need to think on this a little as there's lots of different options a user may want to specify in this area. How would a function that returned the content and path work, that you could then adjust and send on work for you?

kkoppel:
Yes, I think this solution would be quite OK for me.
Thanks for looking into this!

geertn:
We're hitting this as well. I setup a HTTPS nginx instance to receive metrics and forward them to the pushgateway. Specifying https doesn't work since the http is hardcoded. A redirect doesn't work (and according to the RFC shouldn't work for PUT requests anyway):
Traceback (most recent call last):
  File "testpushd.py", line 12, in <module>
    push_to_gateway('prom-push.icpacs.eu', job='batchA', registry=registry)
  File "/home/geertn/Envs/promtest/lib/python2.7/site-packages/prometheus_client/exposition.py", line 88, in push_to_gateway
    _use_gateway('PUT', gateway, job, registry, grouping_key, timeout)
  File "/home/geertn/Envs/promtest/lib/python2.7/site-packages/prometheus_client/exposition.py", line 122, in _use_gateway
    resp = build_opener(HTTPHandler).open(request, timeout=timeout)
  File "/usr/lib64/python2.7/urllib2.py", line 410, in open
    response = meth(req, response)
  File "/usr/lib64/python2.7/urllib2.py", line 523, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib64/python2.7/urllib2.py", line 442, in error
    result = self._call_chain(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 382, in _call_chain
    result = func(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 608, in http_error_302
    new = self.redirect_request(req, fp, code, msg, headers, newurl)
  File "/usr/lib64/python2.7/urllib2.py", line 569, in redirect_request
    raise HTTPError(req.get_full_url(), code, msg, headers, fp)
urllib2.HTTPError: HTTP Error 301: Moved Permanently

[...]

PR #126, opened 2017-01-15
Add handler parameter to pushgateway functions to permit HTTP AUTH etc. (refs issue #60)

rossigee:
Takes the branch by @tim-seoss (PR #120), fixes a minor regression, moves the default handler code out to be more easily re-used/adapted, and adds a basic unit test for a custom handler.

A simple handler function for adding Basic Auth might then look like this...

```
#!/usr/bin/python

import os, base64

from prometheus_client.handlers.base import handler as default_handler

def handler(url, method, timeout, headers, data):
    '''Handler that implements HTTP Basic Auth using environment variables 'PUSHGW_USERNAME' and 'PUSHGW_PASSWORD'.'''
    username = os.environ['PUSHGW_USERNAME']
    password = os.environ['PUSHGW_PASSWORD']
    auth_value = "{0}:{1}".format(username, password)
    auth_header = "Basic {0}".format(base64.b64encode(bytes(auth_value)))
    headers.append(['Authorization', auth_header])
    default_handler(url, method, timeout, headers, data)
```

@brian-brazil - your thoughts?


rossigee:
I found I needed to pass additional arguments to the handler, such as the username/password, so I have updated the patch to include a 'handler_args' parameter, and included a simple 'basic_auth' handler too.

Usage then becomes something like this...

```
from prometheus_client.handlers.basic_auth import handler as http_basic_auth_handler
auth_args = None
if username is not None:
    auth_args = {
        'username': username,
        'password': password
    }

push_to_gateway(url, job=job, registry=registry, handler=http_basic_auth_handler, handler_args=self.auth_args)
```

brian-brazil:
Can you rebase this?

I'd have presumed that any args are embedded in the handler, so there's no need to pass args beyond that.

rossigee:
Rebased.

Re: 'args embedded in the handler' - I'm not sure what you mean or how to do that. The only way other way I could think to minimise the footprint was to change it from a function to an object, but that didn't feel right for some reason, so I used the 'kwargs' approach I've seen used elsewhere. My python isn't that sharp though, so I'm open to better suggestions.

Anyway, have now tested this branch more widely with several clients and they are now pushing data happily to push gateway services that are using https+basicauth, so I can at least report it works for me.

Also, I've just seen the py26 travis build failure (URLError), but I can't seem to figure out wtf would cause that, other than perhaps a python26/urllib2 bug. I can't see how it would be related to this patch, unless I'm missing something?

brian-brazil:
That's still not rebased.

> Re: 'args embedded in the handler' - I'm not sure what you mean or how to do that.

Use a closure. So instead of:

```
def func(a): return a
func(x)
```

do
 
```
def func(): return x
func()
```

This works with classes too.

rossigee:
@brian-brazil - ok, i think i've got it - how does it look now?

README.md
```diff
@@ -321,6 +321,25 @@ for more information.
...
+
+```python
+from prometheus_client import CollectorRegistry, Gauge, push_to_gateway
+from prometheus_client.handlers.basic_auth import handler as basic_auth_handler
+import os
+
+def my_auth_handler(url, method, timeout, headers, data):
+    username = os.environ['PUSHGW_USERNAME']
+    password = os.environ['PUSHGW_PASSWORD']
+    return basic_auth_handler(url, method, timeout, headers, data, username, password)
```
brian-brazil: I'd hardcode the user/pw here to keep in simpler.

If you use an env example, some users will be confused and think it only works if it comes from the env.

prometheus_client/exposition.py
```diff
@@ -185,13 +223,11 @@ def _use_gateway(method, gateway, job, registry, grouping_key, timeout):
...
 
-    request = Request(url, data=data)
-    request.add_header('Content-Type', CONTENT_TYPE_LATEST)
-    request.get_method = lambda: method
-    resp = build_opener(HTTPHandler).open(request, timeout=timeout)
-    if resp.code >= 400:
-        raise IOError("error talking to pushgateway: {0} {1}".format(
-            resp.code, resp.msg))
+    headers=[('Content-Type', CONTENT_TYPE_LATEST)]
+    if handler is None:
```
brian-brazil: The handler isn't mutable, so you can put it as a default in the def line.

prometheus_client/exposition.py
```diff
@@ -10,18 +10,19 @@
 from wsgiref.simple_server import make_server
 
 from . import core
+
+from .handlers.base import handler as default_handler
```
brian-brazil: I'd put these in this file to keep things all in one place.

prometheus_client/handlers/basic_auth.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/python
+
+import base64
+
+from prometheus_client.handlers.base import handler as default_handler
+
+def handler(url, method, timeout, headers, data, username = None, password = None):
+    def handle():
+        '''Handler that implements HTTP Basic Auth by setting auth headers if
```
brian-brazil: The first sentence in a docstring should be on one line.

README.md
rossigee: Example updated.

prometheus_client/handlers/basic_auth.py
rossigee: OK, done.

prometheus_client/exposition.py
rossigee: Not sure I understand. Do you mean something like...

```
from .handlers.base import default_handler
```

and in base.py...

```
default_handler = handler
```

or something else???

prometheus_client/exposition.py
rossigee: Not sure I can see how that would work, as the function is called with a handler of None by other functions in exposition.py. Do you mean make 'default_handler' the default instead of None for all of them?

brian-brazil: > Do you mean make 'default_handler' the default instead of None for all of them?

Yes

prometheus_client/exposition.py
brian-brazil: I mean put all of this in exposition.py

prometheus_client/exposition.py
rossigee: Done.

prometheus_client/exposition.py
rossigee: Done.

prometheus_client/exposition.py
```diff
@@ -118,7 +118,22 @@ def write_to_textfile(path, registry):
     os.rename(tmppath, path)
 
 
-def push_to_gateway(gateway, job, registry, grouping_key=None, timeout=None):
+def default_handler(url, method, timeout, headers, data):
```
brian-brazil: Can you add a docstring to indicate this is for use with the pgw functions?

prometheus_client/handlers/basic_auth.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/python
+
+import base64
+
+from prometheus_client.exposition import default_handler
+
+def handler(url, method, timeout, headers, data, username = None, password = None):
```
brian-brazil: This should go in exposition.py too.

No spaces around the = here

prometheus_client/exposition.py
rossigee: Done.

brian-brazil:
The unittests aren't passing on all version of Python, could you fix that up and we can get this submitted?

rossigee:
I'm not having much luck fixing the python2.6 test. I have no py26 environment locally to reproduce it with, and from what I can determine from the error/stacktrace, it's more of an issue with python2.6's urllib2 implementation on the server than it is a regression caused by this patch, unless you can see something that I'm missing.

brian-brazil:
Doing some testing, it seems to be the lack of a http:// on the url that's causing the issue.

rossigee:
That's the same conclusion I came to, hence the last patch and revert. I've just tried another workaround, which seems to have cleared it now. Anyway, I couldn't see anything in the patch that would affect that behaviour, so I'm guessing something changed in the py26 test environment since the test was passing. I'd be interested to see if the master branch (unpatched) still passes that test.

brian-brazil:
Thanks!

PR #128, opened 2017-01-18
Return thread object from start_http_server

bdkearns:
When starting a server using port 0 (to allow the OS to choose an open port), one needs to retrieve the selected port after HTTPServer creation. This PR allows doing so through:
```
t = start_http_server(0)
port = t.httpd.server_port
```

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -97,13 +97,17 @@ class MetricsHandler(BaseHTTPRequestHandler):
 
 def start_http_server(port, addr=''):
     """Starts a HTTP server for prometheus metrics as a daemon thread."""
+    e = threading.Event()
     class PrometheusMetricsServer(threading.Thread):
         def run(self):
-            httpd = HTTPServer((addr, port), MetricsHandler)
-            httpd.serve_forever()
+            self.httpd = HTTPServer((addr, port), MetricsHandler)
+            e.set()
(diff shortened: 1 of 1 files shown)


brian-brazil:
You've got a race there.

I've been wary of returning the internals of this function, as it ties us to the current implementation. If this sort of thing is ever to be added it should be via a clean API.

As the function is so simple, it's generally best to implement your own if you need something custom. This function is intended as quick way to get up and running, rather than a full solution as I'd expect most users to already have a port serving http that they'd want to use.

bdkearns:
Care to explain the race?

brian-brazil:
You don't know that it's serving by the time the function returns.

bdkearns:
`HTTPServer.server_bind` (which binds and sets `self.server_port`) is called from `TCPServer.__init__`, so before the event is set and function returns.

brian-brazil:
This is stale.

BlackHoleFox:
@brian-brazil I ran into this issue on a project im working on where the HTTPServer is part of a hot reloadable "cog". Whenever it reloads it crashes due to the server remaining even after its Python parent vanishes. Can this PR be implemented by any chance so its possible to shutdown the thread?

brian-brazil:
This PR has gone stale, and I'd prefer to keep this code nice and simple to let users get up and running quickly.

PR #139, opened 2017-02-13
Changes to allow each HTTP request to be a seperate thread

jamessewell:
@brian-brazil 

This change allows each HTTP request to be served by a separate thread (I think!).

This came about because of a network issue, which was causing one Prometheus to keep a connection open - causing all other Proms to fail the exporter as they couldn't get a connection.

This is also consistent with the Go exporter behavior.

Cheers,
James

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -97,10 +97,9 @@ class MetricsHandler(BaseHTTPRequestHandler):
 
 
 def start_http_server(port, addr=''):
-    """Spawns each HTTPServer in a new thread to prevent blocking."""
+    """Starts an HTTP server for prometheus metrics as a daemon thread"""
     class ThreadingSimpleServer(ThreadingMixIn, HTTPServer):
         pass
-    """Starts a HTTP server for prometheus metrics as a daemon thread."""
     class PrometheusMetricsServer(threading.Thread):
         def run(self):
             httpd = ThreadingSimpleServer((addr, port), MetricsHandler)
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -12,8 +12,8 @@ from wsgiref.simple_server import make_server
 
 from . import core
 try:
-    from BaseHTTPServer import BaseHTTPRequestHandler
-    from BaseHTTPServer import HTTPServer
+    from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+    from  SocketServer import ThreadingMixIn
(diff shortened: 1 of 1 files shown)


prometheus_client/exposition.py
```diff
@@ -97,10 +97,13 @@ def log_message(self, format, *args):
 
 
 def start_http_server(port, addr=''):
-    """Starts a HTTP server for prometheus metrics as a daemon thread."""
+    """Spawns each HTTPServer in a new thread to prevent blocking"""
```
brian-brazil: This should still mention the daemon thread, as that's more important.

prometheus_client/exposition.py
```diff
@@ -12,8 +12,8 @@
 
 from . import core
 try:
-    from BaseHTTPServer import BaseHTTPRequestHandler
-    from BaseHTTPServer import HTTPServer
+    from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+    from  SocketServer import ThreadingMixIn
```
brian-brazil: extra space

prometheus_client/exposition.py
```diff
@@ -12,8 +12,8 @@
 
 from . import core
 try:
```
brian-brazil: You also need to change Python3

prometheus_client/exposition.py
jamessewell: Done

prometheus_client/exposition.py
jamessewell: Done

prometheus_client/exposition.py
jamessewell: Done

prometheus_client/exposition.py
```diff
@@ -97,10 +97,10 @@ def log_message(self, format, *args):
 
 
 def start_http_server(port, addr=''):
-    """Spawns each HTTPServer in a new thread to prevent blocking"""
+    """Spawns each HTTPServer in a new thread to prevent blocking."""
     class ThreadingSimpleServer(ThreadingMixIn, HTTPServer):
         pass
-    """Starts a HTTP server for prometheus metrics in a new thread"""
+    """Starts a HTTP server for prometheus metrics as a daemon thread."""
```
brian-brazil: The daemon thread should be mentioned in start_http_server, as that's the primary docs the user will see.

jamessewell: Ah, fixed

brian-brazil:
Thanks!

Issue #140, opened 2017-02-13
`Histogram.time` reuses the same `_HistogramTimer` instance for multiple requests when used as a decorator


obsoleter:
`Histogram.time` reuses the same `_HistogramTimer` instance for multiple requests when used as a decorator.  This results in incorrectly timed requests.  A simple example app bears this out:

```
import time

from flask import Flask, make_response
from werkzeug.wsgi import DispatcherMiddleware
from prometheus_client import make_wsgi_app, Histogram


app = Flask(__name__)
LATENCY = Histogram('request_latency_seconds', 'Tracking response times')


@LATENCY.time()
@app.route('/')
def index():
    time.sleep(2)
    return make_response('hello')


metrics_app = make_wsgi_app()
application = DispatcherMiddleware(app, {
    '/app': app,
    '/metrics': metrics_app
})


if __name__ == '__main__':
    import waitress
    waitress.serve(application, threads=50)
```

With this app, no requests should ever end up in any of the buckets below 2.5 seconds.  And yet many do.

brian-brazil:
That would be bad, how do you suggest fixing this?

obsoleter:
I mistyped in the details of the pull request.  It will provide the same behavior as using `Histogram.time()` as a context manager (with block).

brian-brazil:
Thanks, could you add a unittest and fix the other two?

brian-brazil:
I can't reeproduce, so closing.

PR #141, opened 2017-02-13
Fixes Histogram decorator usage (#140)

obsoleter:
Without creating a new instance of `_HistogramTimer` for each call, the internal timer is reused across executions of the instrumented function.  This results in erroneous timings.  The solution is to create a new `_HistogramTimer` for each execution of the decorated function.  This provides the same behavior as using the `Histogram.time()` as a decorator within the instrumented function.


brian-brazil:
Closing per #140

PR #143, opened 2017-02-20
start_http_server for any given registry or the default registry

danni-m:
I'm using start_http_server, but unfortunately it only exposes the default registry, all other exposition methods have a registry parameter.

@brian-brazil

### DIFF
--- a/prometheus_client/__init__.py
+++ b/prometheus_client/__init__.py
@@ -18,7 +18,7 @@ Histogram = core.Histogram
 
 CONTENT_TYPE_LATEST = exposition.CONTENT_TYPE_LATEST
 generate_latest = exposition.generate_latest
-MetricsHandler = exposition.MetricsHandler
+MetricsHandler = exposition.generate_metrics_handler(REGISTRY)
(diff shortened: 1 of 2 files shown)


brian-brazil:
You should only ever be exposing the default registry via this method. If you're doing something funky (such as what the snmp or blackbox exporters do), you'll need custom handling anyway.

danni-m:
What I want is to expose 2 different registries on 2 different ports.
start_wsgi_server allows to receive a specific registry while start_http_server doesn't.

How those 2 functions differ except WSGI/SimpleHTTPServer ?

brian-brazil:
> What I want is to expose 2 different registries on 2 different ports.

Why do you want to do this?

A single process should have a single registry.

danni-m:
Is there any reason why `start_wsgi_server` should be different than `start_http_server`?

brian-brazil:
There isn't, that parameter should be removed.

danni-m:
@brian-brazil would you accept a PR that removes the parameter?

brian-brazil:
I would.

PR #146, opened 2017-03-20
Provide robust Gunicorn config in multiprocess docs

StephanErb:
We need to use the recently added Gunicorn `child_exited` callback in order to
correctly clean up after a segfaulted or OOM-killed worker (for details, see
https://github.com/benoitc/gunicorn/pull/1394 and #115).

The try catch is necessary to guard against unexpected race conditions and
errors in the cleanup logic. I have seen this twice, crippling the master
process in a cascading failure scenario.

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,8 +421,9 @@ between Gunicorn runs (before startup is recommended).
 
 Put the following in the config file:
 ```python
-def worker_exit(server, worker):
-    from prometheus_client import multiprocess
+from prometheus_client import multiprocess
+
+def child_exit(server, worker):
     multiprocess.mark_process_dead(worker.pid)
 ```
 


README.md
```diff
@@ -421,9 +421,13 @@ between Gunicorn runs (before startup is recommended).
 
 Put the following in the config file:
 ```python
-def worker_exit(server, worker):
-    from prometheus_client import multiprocess
-    multiprocess.mark_process_dead(worker.pid)
+def child_exit(server, worker):
+    import logging
```
brian-brazil: It's not safe to do imports inside functions

StephanErb: So you are suggest that also `from prometheus_client import multiprocess` has to be moved?

I will need to re-test then, if the try catch is still needed after extracting the import.

brian-brazil: Yes

StephanErb:
I have pushed a squashed commit with the extracted import. I also dropped the try catch. The previous errors I was seeing happened during import, as this is now done up-front I don't see any significant error potential.

brian-brazil:
Thanks!

PR #148, opened 2017-03-21
Feature/params

seanhoughton:
Hello @brian-brazil 

I've added support for query parameters in the collect() function of custom collectors. This is helpful when you want to provide a "target" parameter, similar to the Blackbox and SNMP exporters.

Unfortunately this means a function signature change but I'm not sure there's a better alternative.


brian-brazil:
If you want to do blackbox/snmp things you have to use a custom registry. Allowing arbitrary unstructured data to be passed to collectors is not desirable.

seanhoughton:
Are you concerned with the fact that the argument is a dictionary or that query string key/value pairs are unstructured? Most web python frameworks provide the query strings as dictionaries in a similar way so I'm not sure what the objection is.

brian-brazil:
That data is being passed at all. From the Prometheus client library standpoint, you want to pass an arbitrary blob into a collector. That's not sanely supportable.

seanhoughton:
What if the argument was a list of tuples?
QueryParameter = namedtuple('QueryParameter', ['field', 'value'])

Given that this is a web service written in Python I wouldn't consider a dictionary of query string arguments to be an insane idea. Do you have any suggestions on how to provide arguments to the collector? You mentioned using a custom registry, do you mean writing my own application w/out using client_python?

brian-brazil:
It makes no difference. No data should be passed to the collector, as that's not the api.

As mentioned you need a custom registry, look at how the old snmp_exporter worked.

Issue #150, opened 2017-03-24
Original metrics classes are not directly accessible


beezz:
This is not an issue, just an observation maybe.

Due to the `_MetricWrapper`, metric classes (`Counter`, `Gauge`, `Histogram`, `Summary`) are not directly accessible.

Example:
```python
from unittest.mock import Mock
from prometheus_client import Gauge

set_mock = Mock()
Gauge.set = set_mock

g = Gauge('metric', 'Some metric')

g.set(1)
g.set(2)
g.set(3)
g.set(5)

set_mock.assert_called_once()
```

Output:
```
AssertionError: Expected 'mock' to have been called once. Called 0 times.
```

You can circumvent this issue by accessing the ``__class__`` attribute of metric class instance.

Example:
```python
Gauge('foo', 'bar').__class__.set = set_mock
```

I suggest to make them accessible the same way the `functools.update_wrapper` does, via `__wrapped__` attribute ([update_wrapper code](https://github.com/python/cpython/blob/master/Lib/functools.py#L70)) and of course document it. 

I'll gladly prepare PR for this!

Thanks for your great work and have a nice day!

PR #151, opened 2017-03-24
Expose original metric classes via __wrapped__ attribute

beezz:
Related to issue #150


brian-brazil:
A small unittest would be a good idea I think.

brian-brazil:
Thanks!

beezz:
Unittest is always a good idea 👍

PR #152, opened 2017-03-28
Add PlatformCollector to collect information about the platform

mortenlj:
In the same vein as https://github.com/prometheus/client_java/commit/bdc6f2507a81029c1aec6cd8b5276661bb6c7ee1 and https://github.com/siimon/prom-client/commit/ee561e28e14a62b0ef069f525174ee9a3b66d4d7, this PR adds Python version and some other metadata as labels on a small set of metrics.

This has turned out to be very useful for creating common dashboards across implementation languages, allowing the dashboard to extract metadata from the metrics exposed.

Python, machine and system information is collected once at startup, and
exposed as labels on a metric with value 1.

### DIFF
--- a/prometheus_client/platform_collector.py
+++ b/prometheus_client/platform_collector.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+from __future__ import unicode_literals
+
+import platform as pf
+
+from . import core
+
+
+class PlatformCollector(object):
+    """Collector for platform information"""
+
+    def __init__(self, registry=core.REGISTRY, platform=None):
+        self._platform = pf if platform is None else platform
+        self._metrics = [
+            self._add_metric(*self._python()),
+            self._add_metric(*self._machine()),
+            self._add_metric(*self._system())
+        ]
+        if registry:
+            registry.register(self)
+
+    def collect(self):
+        return self._metrics
+
+    @staticmethod
+    def _add_metric(name, documentation, data):
+        labels = data.keys()
+        values = [data[k] for k in labels]
+        g = core.GaugeMetricFamily(name, documentation, labels=labels)
+        g.add_metric(values, 1)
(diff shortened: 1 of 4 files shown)


brian-brazil:
Machine level information should come from the node exporter, it's not the concern of a python application. Cut this down to just the python stuff.

mortenlj:
Yeah, I can see the point about the machine info, I'll remove that.

I'll remove the windows, linux, darwin stuff too, as that is also something the node exporter can handle.

What do you think about the java info? If you are using jython, which java version your application is using becomes relevant, and it's not something the node exporter can tell you (it might tell you which java versions are installed, but not necessarily which one is used to run your application).

brian-brazil:
Keeping Jython stats makes sense. There's an open question on how to handle a mix of Jython and client_java stats, but noone has asked about it yet.

prometheus_client/platform_collector.py
```diff
@@ -0,0 +1,59 @@
...
+            "implementation": self._platform.python_implementation(),
+            "major": major,
+            "minor": minor,
+            "patchlevel": patchlevel
+        }
+
+    def _java(self):
+        java_version, _, vminfo, osinfo = self._platform.java_ver()
+        vm_name, vm_release, vm_vendor = vminfo
+        system, kernel, _ = osinfo
```
brian-brazil: this is unused

prometheus_client/platform_collector.py
```diff
@@ -0,0 +1,59 @@
...
+# -*- coding: utf-8
+from __future__ import unicode_literals
+
+import platform as pf
+
+from . import core
+
+
+class PlatformCollector(object):
+    """Collector for platform information"""
```
brian-brazil: "Python platform" would be a little clearer.

brian-brazil:
Thanks!

PR #153, opened 2017-03-31
Add an obvious notice for metric initialization of multiprocess collector

justdoit0823:
As a beginner, I build a histogram metric in my Python application with a self-defined collect registry.When deploy under multiple process circumstance, I get a help from README file. According to that, I don't change metric's initialzation and get http 500 when making requests to pushgateway. Finally, I fix the problem and think there should be a more clear and obvious description for beginners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,6 +458,8 @@ def app(environ, start_response):
     return iter([data])
 ```
 
+**Note: don't initialize metric with MultiProcessCollector registry argument.**
+
 **Three**: Instrumentation
 
 Counters, Summarys and Histograms work as normal.


brian-brazil:
The example already indicates how to do it correctly. I don't think it's sane to list every way in why someone managed to not follow the example correctly.

justdoit0823:
while there is no any more explaination about CollectorRegistry class, collector will fail when using multiprocess collector's registry as metric registry argument. It just like a balck box.

brian-brazil:
This is all deep magic. Attempting to explain it would only cause more user confusion.

justdoit0823:
With more detailed document about how CollectorRegitry works with metric, beginners can easily make sense of the structure of metric observation and metric collector.

brian-brazil:
Users shouldn't have to care. The docs provide full examples on how to use this correctly, and anyone looking for more can read the code or ask for help.

Issue #130, opened 2017-01-24
Non-gunicorn multiprocessing (mostly questions)


tlc:
I am trying to use multiprocessing support even though I am not using gunicorn.  

I have a master proc calling multiprocessing.Process() to create workers.  (I'm forking but not execing.)

All my mmap files in `prometheus_multiproc_dir` have the parent process id in the filename.  I *believe* this is not as intended and means that stats are being aggregated as they're saved instead of when they're served.  (A performance hit.)

I will try to import prometheus_client after the fork -- and in another file.  But that file will be imported...  Do the python imports make it impossible for me to have different global _ValueClass's without an exec or monkeying in prometheus_client?

Also, BTW, there is still a reference to shelve in the source.

brian-brazil:
We don't currently support multiprocessing, as it doesn't work the way we expect gunicorn-like things to.

dozer47528:
@brian-brazil  I have the same problem, any plan to support uwsgi?

brian-brazil:
Can uwsgi start a brand new process for requests?

We currently support a partiuclar model of using multiple processes, not particular libraries.

dozer47528:
I use uwsgi to start my application

` uwsgi --http-socket :8000 --master --workers 4 --enable-threads --threads 10 --wsgi-file server/wsgi.py`

There is one master process and four worker process.

But the `client_python` only created one file with the pid of master process.

brian-brazil:
Unless you can find a way to make it fork&exec each worker, it's not going to work currently.

banteg:
uWSGI config example:
```ini
[uwsgi]
http-socket = :8000
enable-threads = yes
lazy-apps = yes
processes = 4
wsgi = app:app
env = prometheus_multiproc_dir=multi
```

esseti:
@banteg can you provide a  full example. this is the uwsgi ini file, but how/where do you start the wsgi_server? what port?

ecksun:
> We don't currently support multiprocessing, as it doesn't work the way we expect gunicorn-like things to.

> We currently support a partiuclar model of using multiple processes, not particular libraries.

There is not much documentation about what is and is not supported, nor what model you are refering to, could you expand on that @brian-brazil ?

I have a kafka multiprocess worker which I want to export metrics from. The code below appears to work (I intend to use systemd ExecStopPost to cleanup afterwards), is there something I'm missing?

```python
def start_prometheus_server():
[...]

PR #156, opened 2017-04-13
Move os.getpid() inside _MmapedValue (#130)

megabuz:
Related to issue #130 
uwsgi imports all python script in master process and after that forks children, so we need move os.getpid() call inside function

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -392,8 +392,7 @@ class _MmapedDict(object):
             self._f = None
 
 
-def _MultiProcessValue(__pid=os.getpid()):
-    pid = __pid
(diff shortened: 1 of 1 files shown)


brian-brazil:
We don't currently support that mode of operation, each child needs to be fork+execd. So this won't work.

megabuz:
It should work in both ways

megabuz:
Do you mean that whole module will not work? Do you have some vars/classes that need to init in child and not in master process?

brian-brazil:
Yes, it depends on the child initting separately.

brian-brazil:
Closing in favour of #169

Issue #158, opened 2017-04-24
namespace/subsystem are not taken into consideration in multiproc mode 


atollena:
The `namespace` and `subsystem` kwargs add prefixes to metric name when running normally. However, when using the multiprocess mode of the library, those prefixes are ignored.

Here is a small program to reproduce this problem:
```
from prometheus_client import Counter, CollectorRegistry, start_http_server, generate_latest, multiprocess
import time

registry = CollectorRegistry()
c = Counter('test', 'Test', namespace='foo', subsystem='bar', registry=registry)
c.inc()

multiprocess.MultiProcessCollector(registry)
print(generate_latest(registry).decode())
```

Running this program prints:
```
# HELP foo_bar_test Test
# TYPE foo_bar_test counter
foo_bar_test 1.0
# HELP test Multiprocess metric
# TYPE test counter
test 1.0
```

atollena:
@brian-brazil `full_name` could simply be passed instead of just `name` at https://github.com/prometheus/client_python/blob/master/prometheus_client/core.py#L537 and https://github.com/prometheus/client_python/blob/master/prometheus_client/core.py#L539, but it's not obvious to me that it is the right fix.

I'll be happy to make this change or another if you think that it is reasonable.

brian-brazil:
That sounds like the right fix, and a PR would be great.

PR #159, opened 2017-04-25
Fix namespace and subsystem in multiprocess mode

atollena:
See https://github.com/prometheus/client_python/issues/158 for details.


atollena:
@brian-brazil here is the suggested fix for https://github.com/prometheus/client_python/issues/158.

brian-brazil:
Thanks!

StephanErb:
@atollena I recently stumbled over this as well. Thanks for the fix :)

PR #167, opened 2017-06-01
Use only a single global lock for multiproc.

### LINKED ISSUE #112: Reduce locks in multiproc mode
Currently there's a lock on the metric, and one on the mmapped file.

Given that multiproc is in use it's unlikely that any threading is going on (and there's still the GIL) so these could both be replaced by a single global lock. This would bring us down to one mutex, which is the same as non-multiproc which means we'd have about the same performance.

brian-brazil:
This cuts ~350ns off the time for an operation, bringing it to
~1700ns.

In multi process mode it's a reasonable assumption that there
isn't any threading going on, so a single lock is faster than
more granular locks.

Fixed #112

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -309,6 +309,7 @@ class _MutexValue(object):
       with self._lock:
           return self._value
 
+
 class _MmapedDict(object):
     """A dict of doubles, backed by an mmapped file.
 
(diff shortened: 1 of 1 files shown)


vad:
@brian-brazil you assume multiproc means no threads, right? Is it still like this? Looks a strong assumption to me (I'm in production with multiproc and multithread, pretty common thing for uWSGI users). Thank you

PR #175, opened 2017-07-05
Should use standart time.time instead of timeit.default_timer

melnikk:
This is a rollback PR of #106 because of buggy timestamp results of timeit.default_timer.


brian-brazil:
default_timer is behaving correctly, however the graphite bridge shouldn't be using it as it wants the actual unix time. Can you rollback just that bit?

melnikk:
done

brian-brazil:
Thanks!

PR #176, opened 2017-07-11
Add (optional) timestamp support

dmick:
This modifies the internal sample tracking to include a timestamp with each sample, and modifies the exposition format to include it if present.  Tests are enhanced to test the new format.

### DIFF
--- a/prometheus_client/bridge/graphite.py
+++ b/prometheus_client/bridge/graphite.py
@@ -68,8 +68,12 @@ class GraphiteBridge(object):
                             for k, v in sorted(labels.items())])
                 else:
                     labelstr = ''
+                if value[1] is not None:
+                    ts = value[1]
+                else:
+                    ts = now
                 output.append('{0}{1}{2} {3} {4}\n'.format(
-                    prefixstr, _sanitize(name), labelstr, float(value), now))
+                    prefixstr, _sanitize(name), labelstr, float(value[0]), ts))
 
         conn = socket.create_connection(self._address, self._timeout)
         conn.sendall(''.join(output).encode('ascii'))
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -368,7 +368,7 @@ class _MmapedDict(object):
             pos += 8
 
     def read_all_values(self):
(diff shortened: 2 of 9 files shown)


dmick:
A possible issue: read/write to multiprocess mmaped files is now 16 bytes/sample, which may be less atomic than it was.  As I understand it, multiprocess mode has one file per process, so I think this is not an issue, but I'm not 100% sure.

dmick:
@brian-brazil

dmick:
(repushed with python2.6 bug fixed: no format() with empty {})

brian-brazil:
Thanks for the PR, implementing this was on my backlog.

Timestamps are a very niche feature that are required for some advanced use cases. I wasn't intending to expose them to the *MetricFamilies to avoid misuse, and there's no use cases for it with direct instrumentation such as Counters. So if you remove all that extra stuff this will be simpler and multi-proc doesn't come into it.

I was also thinking to switch the base type from a tuple to a namedtuple so it's easier to understand the code and work with.

dmick:
I guess I'm not sure what a collector written without Counter et al or MetricFamilies looks like; supporting those seems to be the whole point of the  client lib?

dmick:
(and namedtuples also tempted me.)

brian-brazil:
> I guess I'm not sure what a collector written without Counter et al or MetricFamilies looks like; supporting those seems to be the whole point of the client lib?

MetricFamilies are convenience methods. My thoughts are that for the first version of this anyone who wants to use timestamps can do things the old fashioned way from before they were added.

We've seen a lot of abuse of timestamps by users trying to convert Prometheus to push and other various misunderstandings, vastly outweighing the handful of legitimate use cases for timestamps. Thus I want to play this on the safe side, and not make them too tempting for misuse.

brian-brazil:
We discussed this offline and there was a misunderstanding of the Prometheus architecture, so this PR is no longer the approach Dan is taking. I'm leaving the PR lying around so I can borrow from it whenever I get around to implementing this feature.

brian-brazil:
Someone else added this.

PR #178, opened 2017-07-12
Remove hack to encode __all__ as ascii

bz2:
Previously the code included:

    from __future__ import unicode_literals

But with that gone in the current version, it makes no sense to mess with the contents of `__all__` at all.


brian-brazil:
Thanks!

PR #181, opened 2017-07-31
Test and support Python 3.6

bz2:
Add py36 to the Travis CI config, and `setup.py` classifiers.

Also tweak local tox config, install twisted for 3.6 but make `noptionals` variant use 2.7 and 3.5 only.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = coverage-clean,py26,py27,py34,py35,py36,pypy,{py27,py35,py36}-nooptionals,coverage-report
+envlist = coverage-clean,py26,py27,py34,py35,py36,pypy,{py27,py36}-nooptionals,coverage-report
 
 
 [base]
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = coverage-clean,py26,py27,py34,py35,py36,pypy,{py27,py35,py36}-nooptionals,coverage-report
+envlist = coverage-clean,py26,py27,py34,py35,py36,pypy,{py27,py35}-nooptionals,coverage-report
```
brian-brazil: Remove 3.5 from the nooptionals, we only need a 2 and a 3 there.

brian-brazil: Let's test the noooptionals with 3.6 rather than 3.5

bz2: Sure thing.

bz2:
Okay, 9c07a12 has the nooptionals variant run for 2.7 and 3.6 only for both travis and tox.

brian-brazil:
Thanks!

PR #184, opened 2017-08-05
Convert MetricFamily labels -> list

manics:
On Python 3 the labels of the `python_info` metric is a reference to the `dict_keys` of the data object used to create it. This PR ensures all labelnames in MetricFamilies it are a static tuple instead.

This PR is motivated by some broken tests in https://github.com/korfuri/django-prometheus with recent versions of prometheus-client: https://github.com/korfuri/django-prometheus/pull/46#issuecomment-320453360

Although the docs do not say Metrics should be serializable this change seems more in keeping with the intention that labels should be owned by the Metric and not a reference to an external object.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -187,7 +187,7 @@ class CounterMetricFamily(Metric):
             raise ValueError('Can only specify at most one of value and labels.')
         if labels is None:
           labels = []
-        self._labelnames = labels
+        self._labelnames = tuple(labels)
         if value is not None:
           self.add_metric([], value)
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
This doesn't seem like the right fix, this should be down in GaugeMetricFamily I think. Also a unittest would be good.

manics:
Do you mean GaugeMetricFamily should ensure labels is a list of strings in __init__?: https://github.com/prometheus/client_python/blob/master/prometheus_client/core.py#L215

brian-brazil:
Yes

manics:
@brian-brazil I've force-pushed away the original commit, and converted labelnames to a tuple in all MetricFamilies

brian-brazil:
Thanks!

zoidbergwill:
Yay, thanks for this PR.  🎉

PR #190, opened 2017-08-29
Add asyncio support

### LINKED ISSUE #187: asyncio support
Hi, have you thought about support of [asyncio](https://docs.python.org/3/library/asyncio.html)?

```python
import asyncio
from prometheus_client import start_http_server_async

loop.create_task(start_http_server_async(8000))
loop.run_forever()
```

koorgoo:
Fixes #187 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ dist
 *.egg-info
 *.pyc
 *.swp
+.cache
 .coverage.*
 .coverage
 .tox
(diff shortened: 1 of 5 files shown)


koorgoo:
@brian-brazil review, please.

brian-brazil:
As I said I think this might be handled best in a separate package rather than in the core client as there'll be things beyond this required, see https://pypi.python.org/pypi/prometheus-async/17.4.0 for example.

koorgoo:
Hmm... prometheus_async provides with useful decorators for async functions. Without them prometheus_client won't be as useful as the former. Suppose, you are right. A separate package is a better solution.

PR #202, opened 2017-10-05
ConstLabels equivalent

flaminscotsman:
This aims to provide similar functionality to the ConstLabels in the Go client by allowing one or more labels to be bound at metric initialisation time.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -11,6 +11,7 @@ import re
 import struct
 import time
 import types
+from itertools import chain
 
 try:
     from BaseHTTPServer import BaseHTTPRequestHandler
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks for the PR. This doesn't make sense for the Python client, as unlike the Go client we require that only one of a metric exist and don't try to patch that up under the covers in the Registry.

flaminscotsman:
No problem. It was more for the "simple" behaviour of ConstLabels to avoid constructs like this:

https://gist.github.com/flaminscotsman/8b6ab867462b830a344b99f782330bef

brian-brazil:
That construct is actually strongly discouraged, you want to use a target label on the Prometheus end of things. An individual target doesn't know about things like "production" and Prometheus already has the instance label.

PR #203, opened 2017-10-05
Make WSGI metrics server not log requests

bz2:
Logging using `start_http_server()` and `start_wsgi_server()` is now consistent.

Also simplifies and documents the server creation logic without otherwise changing behaviour.


brian-brazil:
Thanks!

PR #205, opened 2017-10-20
Fix #95: catch OSError when collecting file descriptors.

### LINKED ISSUE #95: process_collector.py assumes /proc/self/fd is readable for process_open_fds
Apache with prefork MPM, not threaded, runs process_collector.py as the chosen non-root user, but /proc/self/fd/ is still owned by root:root so its 0500 permissions cause an OSError exception, e.errno == EACCES, on the os.listdir() attempt.  A patch to give up on that metric when that occurs: [process_open_fds.txt](https://github.com/prometheus/client_python/files/424729/process_open_fds.txt)

blopker:
I also ran into the issue described in #95. It's a show stopper for me right now so I made a PR :)

Cheers!


brian-brazil:
Thanks!

PR #206, opened 2017-10-25
helpful error message when multiprocessing dir env is unset

dbalan:
On head it fails with:
```python
ERROR in app: Exception on /metrics [GET]
AttributeError: 'NoneType' object has no attribute 'endswith'
```
With this change:
```python
  File "build/bdist.linux-x86_64/egg/prometheus_client/multiprocess.py", line 16, in __init__
    raise ValueError('env prometheus_multiproc_dir is not set or not accessible')
ValueError: env prometheus_multiproc_dir is not set or not accessible
```

### DIFF
--- a/prometheus_client/multiprocess.py
+++ b/prometheus_client/multiprocess.py
@@ -12,8 +12,8 @@ from . import core
 class MultiProcessCollector(object):
     """Collector for files for multi-process mode."""
     def __init__(self, registry, path=os.environ.get('prometheus_multiproc_dir')):
-        if not path or not os.path.exists(path):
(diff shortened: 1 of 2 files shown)


prometheus_client/multiprocess.py
```diff
@@ -12,6 +12,8 @@
 class MultiProcessCollector(object):
     """Collector for files for multi-process mode."""
     def __init__(self, registry, path=os.environ.get('prometheus_multiproc_dir')):
+        if not path or not os.path.exists(path):
```
brian-brazil: You should check that it is a directory

dbalan:
@brian-brazil  I've changed the code to check for directories, I can squash the commits if you prefer.

Issue #208, opened 2017-10-31
Cannot parse the correct scheme of URL in Python26


caosiyang:
I work on CentOS 6.6 and Python is 2.6.6. Pushgateway runs on localhost:9091.

When I push metrics with a URL starts with neither "http://" or "https://", just like `push_to_gateway('localhost:9091', job='demo', registry=registry)`, error occurs as follow:
```
# python push_demo.py 
Traceback (most recent call last):
  File "push_demo.py", line 19, in <module>
    push_to_gateway('localhost:9091', job='dev', registry=registry)
  File "/usr/lib/python2.6/site-packages/prometheus_client-0.0.21-py2.6.egg/prometheus_client/exposition.py", line 198, in push_to_gateway
    _use_gateway('PUT', gateway, job, registry, grouping_key, timeout, handler)
  File "/usr/lib/python2.6/site-packages/prometheus_client-0.0.21-py2.6.egg/prometheus_client/exposition.py", line 265, in _use_gateway
    headers=headers, data=data)()
  File "/usr/lib/python2.6/site-packages/prometheus_client-0.0.21-py2.6.egg/prometheus_client/exposition.py", line 133, in handle
    resp = build_opener(HTTPHandler).open(request, timeout=timeout)
  File "/usr/lib64/python2.6/urllib2.py", line 391, in open
    response = self._open(req, data)
  File "/usr/lib64/python2.6/urllib2.py", line 414, in _open
    'unknown_open', req)
  File "/usr/lib64/python2.6/urllib2.py", line 369, in _call_chain
    result = func(*args)
  File "/usr/lib64/python2.6/urllib2.py", line 1223, in unknown_open
    raise URLError('unknown url type: %s' % type)
urllib2.URLError: <urlopen error unknown url type: localhost>
```
I think it's a bug of `urlparse` in Python26.
```
Python 2.6.6 (r266:84292, Jan 22 2014, 09:42:36) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import urlparse
>>> urlparse.urlparse('localhost:9091')
ParseResult(scheme='localhost', netloc='', path='9091', params='', query='', fragment='')
>>>
```

PR #209, opened 2017-10-31
Fix issue #208: parse the correct scheme of URL in Python26

caosiyang:
If python version is 2.6 or older, check if scheme is one of “http" or "https".


brian-brazil:
Thanks!

PR #212, opened 2017-11-13
multiprocess: pass None as default argument

dzen:
The goal here is to evaluate the environ only at parse time : I tried to mock the variable to make some unittests, but I hit those lines : I cannot patch the environ to fill a temp destination for that test

@brian-brazil any comment on this ?


brian-brazil:
Thanks!

Issue #214, opened 2017-12-04
Provide a MetricHandler for multiprocess


ioggstream:
### I expect

that a MetricHandler is provided for multiprocess mode

### Instead

I need to define a new one using external libs (Eg. flask)

ioggstream:
@brian-brazil is that what you intended? It seems cleaner ;)

PR #215, opened 2017-12-04
Fix: #214. Pass a registry to MetricHandler via start_http_server.

ioggstream:
## What does this PR do?
  
Add `registry` parameter to start_http_server, so you can serve data
from a custom registry.

## How it's implemented

Added:
 - a `MultiprocessRegistry` wrapping `CollectorRegistry`
 - the optional named parameter `registry` to `start_http_server`
 - create a  MetricHandler tied to the passed `registry` via a MetricHandler_factory 
 - if `prometheus_multiproc_dir` is set, gather metrics the multiprocess way

## How do you use it
Create a new or a multiprocess registry
Pass `registry=myregistry` to start_http_server.

```
from prometheus_client import multiprocess, start_http_server
...
# Serve multiprocess metrics
registry = multiprocess.MultiProcessRegistry()
start_http_server(8000, registry=registry)

# Server core.REGISTRY metrics
start_http_server(8001) 
...
```
## Further notes.

Changes are backward compatible.

## Complete example

```
import os
from time import sleep
from multiprocessing import Process
from random import randint
from urllib import urlopen
from shutil import rmtree

dpath = os.environ['prometheus_multiproc_dir'] = "/tmp/test_prometheus_multi"
if os.path.isdir(dpath) and dpath:
    rmtree(dpath)
os.makedirs(dpath)


from prometheus_client import multiprocess, start_http_server, core
from prometheus_client import Counter, CollectorRegistry


NUM_COUNTER = Counter("num_requests", "Example counter", ["pid"])


def runner(*args):
    """A process thread.
    """
    while True:
        pid = os.getpid()
        NUM_COUNTER.labels(pid).inc(randint(0, 3))
        sleep(1)


if __name__ == '__main__':
    # Create a multiprocess registry.
    mpr = multiprocess.MultiProcessRegistry()
    start_http_server(8001, registry=mpr)

    # Serve default registry.
    start_http_server(8000)

    # Run 5 processes
    p = [Process(target=runner, args=('o',)) for x in range(5)]
    for j in p:
        j.start()

    for i in range(5):
        sleep(1)
        print(urlopen("http://localhost:8000").read())
        print(urlopen("http://localhost:8001").read())

```

### DIFF
--- a/prometheus_client/__init__.py
+++ b/prometheus_client/__init__.py
@@ -17,6 +17,7 @@ Histogram = core.Histogram
 
 CONTENT_TYPE_LATEST = exposition.CONTENT_TYPE_LATEST
 generate_latest = exposition.generate_latest
+MetricsHandler = exposition.MetricsHandler
 make_wsgi_app = exposition.make_wsgi_app
 start_http_server = exposition.start_http_server
 start_wsgi_server = exposition.start_wsgi_server
--- a/prometheus_client/__init__.py
+++ b/prometheus_client/__init__.py
@@ -17,7 +17,6 @@ Histogram = core.Histogram
 
 CONTENT_TYPE_LATEST = exposition.CONTENT_TYPE_LATEST
 generate_latest = exposition.generate_latest
-MetricsHandler = exposition.MetricsHandler
 make_wsgi_app = exposition.make_wsgi_app
 start_http_server = exposition.start_http_server
 start_wsgi_server = exposition.start_wsgi_server
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -11,6 +11,7 @@ from contextlib import closing
 from wsgiref.simple_server import make_server, WSGIRequestHandler
 
 from . import core
+from .multiprocess import MultiProcessCollector
 try:
     from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
     from SocketServer import ThreadingMixIn
@@ -82,10 +83,12 @@ def generate_latest(registry=core.REGISTRY):
 
 
 class MetricsHandler(BaseHTTPRequestHandler):
-    """HTTP handler that gives metrics from ``core.REGISTRY``."""
+    """HTTP handler that gives metrics from ``MetricsHandler.registry``."""
 
     def do_GET(self):
-        registry = core.REGISTRY
+        registry = MetricsHandler.registry
+        if 'prometheus_multiproc_dir' in os.environ:
+            MultiProcessCollector(registry)
         params = parse_qs(urlparse(self.path).query)
         if 'name[]' in params:
             registry = registry.restricted_registry(params['name[]'])
@@ -107,8 +110,9 @@ class _ThreadingSimpleServer(ThreadingMixIn, HTTPServer):
     """Thread per request HTTP server."""
 
 
-def start_http_server(port, addr=''):
+def start_http_server(port, addr='', registry=core.REGISTRY):
     """Starts an HTTP server for prometheus metrics as a daemon thread"""
+    MetricsHandler.registry = registry
     httpd = _ThreadingSimpleServer((addr, port), MetricsHandler)
     t = threading.Thread(target=httpd.serve_forever)
     t.daemon = True
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -83,21 +83,13 @@ def generate_latest(registry=core.REGISTRY):
 
 
 def factory_MetricsHandler(registry=core.REGISTRY):
-    '''Returns a MetricsHandler tied to a given registry.
+    '''Returns a MetricsHandler class tied to a given registry.
     '''
 
     class MetricsHandler(BaseHTTPRequestHandler):
-        """HTTP handler that gives metrics from ``MetricsHandler.registry``."""
+        """HTTP handler that gives metrics from the context ``registry``."""
         def do_GET(self):
-            # In the _multiprocess case, the CollectorRegistry
-            # should be instantiated at every call or we
-            # get duplicated metrics.
-            if core._ValueClass._multiprocess:
-                r = core.CollectorRegistry()
-                MultiProcessCollector(r)
-            else:
-                r = registry
-
+            r = registry
             params = parse_qs(urlparse(self.path).query)
             if 'name[]' in params:
                 r = r.restricted_registry(params['name[]'])
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -82,28 +82,38 @@ def generate_latest(registry=core.REGISTRY):
     return ''.join(output).encode('utf-8')
 
 
-class MetricsHandler(BaseHTTPRequestHandler):
-    """HTTP handler that gives metrics from ``MetricsHandler.registry``."""
+def factory_MetricsHandler(registry=core.REGISTRY):
+    '''Returns a MetricsHandler tied to a given registry.
+    '''
 
-    def do_GET(self):
-        registry = MetricsHandler.registry
-        if 'prometheus_multiproc_dir' in os.environ:
-            MultiProcessCollector(registry)
-        params = parse_qs(urlparse(self.path).query)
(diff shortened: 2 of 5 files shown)


ioggstream:
Close and Reopen for travis

prometheus_client/multiprocess.py
```diff
@@ -90,6 +97,29 @@ def collect(self):
         return metrics.values()
 
 
+class MultiProcessMetricsHandler(BaseHTTPRequestHandler):
```
brian-brazil: It doesn't make sense to add this, it's too use-case specific. Allowing a reigstry to be passed into the metricshandler would make more sense.

ioggstream: Once you pass registry to MetricsHandler, you still need to call `MultiProcessCollector(registry)`.

Is there a smart way to do it?

@brian-brazil Thx++ for your feedback!

brian-brazil: There's no smart way, it's just a line of code.

ioggstream:
Reopen for travis

prometheus_client/exposition.py
```diff
@@ -82,10 +83,12 @@ def generate_latest(registry=core.REGISTRY):
 
 
 class MetricsHandler(BaseHTTPRequestHandler):
-    """HTTP handler that gives metrics from ``core.REGISTRY``."""
+    """HTTP handler that gives metrics from ``MetricsHandler.registry``."""
 
     def do_GET(self):
-        registry = core.REGISTRY
+        registry = MetricsHandler.registry
+        if 'prometheus_multiproc_dir' in os.environ:
```
brian-brazil: This is incorrect, you don't know that the user wants this. They have to add it to the registry themselves.

ioggstream: How can I decide whether to run the collector?

Can I add an hook like?

```
if hasattr(registry, 'collect_metrics'):
    registry.collect_metrics(registry)
```

Any hint?

ioggstream: How can I decide whether to run this?
```
MultiProcessCollector(registry)
```

Adding a hook?
```
if hasattr(registry, 'collect_metrics'):
    registry.collect_metrics(registry)
```

Hints welcome!

brian-brazil: In https://github.com/prometheus/client_python#multiprocess-mode-gunicorn the user has to manage the registry themselves, this should work the same way.

prometheus_client/exposition.py
```diff
@@ -11,6 +11,7 @@
 from wsgiref.simple_server import make_server, WSGIRequestHandler
 
 from . import core
+from .multiprocess import MultiProcessCollector
```
ioggstream: Unused now ;) removing...

prometheus_client/__init__.py
```diff
@@ -17,7 +17,6 @@
 
 CONTENT_TYPE_LATEST = exposition.CONTENT_TYPE_LATEST
 generate_latest = exposition.generate_latest
-MetricsHandler = exposition.MetricsHandler
```
brian-brazil: Removing this is a breaking change, please keep it around.

prometheus_client/exposition.py
```diff
@@ -81,35 +82,40 @@ def generate_latest(registry=core.REGISTRY):
...
-            self.send_error(500, 'error generating metric output')
-            raise
-        self.send_response(200)
-        self.send_header('Content-Type', CONTENT_TYPE_LATEST)
-        self.end_headers()
-        self.wfile.write(output)
-
-    def log_message(self, format, *args):
-        """Log nothing."""
+def factory_MetricsHandler(registry=core.REGISTRY):
```
brian-brazil: This name doesn't feel right. Do you know if there's conventions around this?

prometheus_client/__init__.py
ioggstream: Ok. I think we need a test in that case.

brian-brazil: I don't feel a strong need for a test for this.

prometheus_client/exposition.py
ioggstream: I had a fast glance and didn't found anything.

I didn't try that, but it could work. Do you like it?

```
class MetricsHandler:
  ...old class stuff..
  @staticmethod
  def factory(registry):
     MH = type('MetricsHandler', (MetricsHandler,), {'registry': registry})
     return MetricsHandler
```

prometheus_client/__init__.py
ioggstream: Explicit is better than implicit ;) my 2¢

prometheus_client/exposition.py
brian-brazil: That would work I think.

ioggstream: Ok, here in Italy it's holiday. I'll give it a try asap. Thanks for all your support!

prometheus_client/exposition.py
```diff
@@ -102,14 +103,30 @@ def do_GET(self):
     def log_message(self, format, *args):
         """Log nothing."""
 
+    @staticmethod
+    def factory(registry):
```
ioggstream: @brian-brazil Dunno if python does dead code elimination, but variables are required to better document the code.

prometheus_client/multiprocess.py
```diff
@@ -9,6 +9,37 @@
 
 from . import core
 
+
+class MultiProcessRegistry(object):
```
brian-brazil: You don't need this test.

tests/test_multiprocess.py
```diff
@@ -129,6 +129,15 @@ def get_pid():
         self.assertEqual(3, self.registry.get_sample_value('c'))
         self.assertEqual(1, c1._value.get())
 
+    def test_multiprocess_registry(self):
```
brian-brazil: This isn't needed

prometheus_client/exposition.py
```diff
@@ -102,14 +103,30 @@ def do_GET(self):
...
+        """Returns a dynamic MetricsHandler class tied
+           to the passed registry.
+        """
+        # This implementation relies on MetricsHandler.registry
+        #  (defined above and defaulted to core.REGISTRY).
+
+        # As we have unicode_literals, we need to create a str()
+        #  object for type().
+        cls_name = str('MetricsHandler')
+        MyMetricsHandler = type(cls_name, (MetricsHandler, object),
```
brian-brazil: The base class should be BaseHTTPRequestHandler

ioggstream: When `type()` is invoked with three arguments, it returns a new-style class.

`BaseHTTPRequestHandler`:
  -  is still in the mro 
  -  is an old-style class and it requires the `object` to be added to the hierarchy.

```
# Create two old-style classes.
class OldStyle: pass
class Child(OldStyle): pass
# a new-style class can't have only classic bases
MyChild = type('MyChild', (Child,), {"a":1}) #  TypeError: 

# Add object to bases.
MyChild = type('MyChild', (Child,object), {"a":1})
# No need to add all the chain.
issubclass(MyChild, OldStyle)
MyChild.mro()

```

tests/test_multiprocess.py
ioggstream: I TDD'd with it ;) is that really worth removing?

Thanks again for all your help! :DDD

brian-brazil: You don't need a MultiProcessRegistry, so you don't need the test

ioggstream: I use MultiProcessRegistry to wrap `MultiProcessCollector`.

How can I gather stats from the directory without it?

brian-brazil: The user can create a registry themselves, no need for a special one in the client.

ioggstream:
Close for now... I'll reopen after a discussion ;)

ioggstream:
@brian-brazil Reopen limited to the MultiProcessRegistry part.

brian-brazil:
It's two lines of code to create one yourself, no need for it to be in the client library.

PR #218, opened 2017-12-05
py26 has fixed dependencies

ioggstream:
## what does this PR do

for py26: fix dependencies: `pytest`, `py`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,7 @@ deps =
     pytest
 
 [testenv:py26]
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -7,11 +7,16 @@ deps =
     coverage
     pytest
 
+[testenv:py26]
+deps =
```
brian-brazil: Can you add a comment as to why this is special cased?

brian-brazil:
Thanks!

PR #220, opened 2017-12-07
Add UntypedMetricFamily to prometheus_client.core

idsvandermolen:
I thought this one was missing from prometheus_client.core, so that's why I propose to add it. But their may be a reason why it wasn't added in the first place.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -176,6 +176,29 @@ class Metric(object):
                 and self.samples == other.samples)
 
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
It's more that noone has asked. Can you add a unittest please?

idsvandermolen:
sure, I'll have a look.

idsvandermolen:
I hope that covers it.

brian-brazil:
Thanks!

PR #227, opened 2017-12-19
check Python version with sys.version_info

minrk:
sys.version_info is a parsed tuple of numbers, so not vulnerable to incorrect string version comparison (e.g. '10' < '3') or parse errors on sys.version.

Discovered because prometheus-client cannot be imported on development versions of Python which have version strings like `3.7.0a0`.


brian-brazil:
Thanks!

PR #237, opened 2018-02-04
Set default timeout of pushgateway actions to 30s

bz2:
Explicitly passing `None` will still trigger the no timeout behaviour of the underlying Python http/socket modules.

This is motived by hitting an odd case in a k8s cluster job where some job pods were hung for 2 hours in the finally-post-metrics step after some cluster unhappiness that had since been resolved. However, I couldn't later reproduce the behaviour locally, as my machine at least does timeout the socket after a reasonable window regardless.

Even so, was worth adapting our application code to set a timeout, and having a sensible default in the client library here also seems reasonable.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -30,6 +30,11 @@ CONTENT_TYPE_LATEST = str('text/plain; version=0.0.4; charset=utf-8')
 
 PYTHON26_OR_OLDER = sys.version_info < (2, 7)
 
+# Default socket timeout for http requests, such as to a pushgateway.
+# Note: Setting this after import time will not change the default used.
+_REQUEST_TIMEOUT = 30
+
+
 def make_wsgi_app(registry=core.REGISTRY):
     '''Create a WSGI app which serves the metrics from a registry.'''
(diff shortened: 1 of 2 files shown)


prometheus_client/exposition.py
```diff
@@ -30,6 +30,11 @@
 
 PYTHON26_OR_OLDER = sys.version_info < (2, 7)
 
+# Default socket timeout for http requests, such as to a pushgateway.
+# Note: Setting this after import time will not change the default used.
+_REQUEST_TIMEOUT = 30
```
brian-brazil: It's be simpler to hardcode it, this is more than enough time.

bz2: Using a (pseudo) const is mostly a hint to use a consistent value for anything doing http requests. Can just use literal everywhere if you'd prefer?

brian-brazil: I'd prefer a literal.

bz2: @brian-brazil Done.

brian-brazil:
Thanks!

PR #238, opened 2018-02-11
Add tox env for running flake8 lint

bz2:
Add ignore rules for all current lint failures, they can be tackled
in follow up branches along with enabling each rule.


brian-brazil:
Thanks!

PR #239, opened 2018-02-11
Fix F811 redefinition lint

bz2:
Depends on lint setup pr #238 and fixes and enables one currently failing rule.

The code change unhides two tests that should be being run.


brian-brazil:
Thanks!

PR #240, opened 2018-02-11
Remove travis pypy version hack and use latest

bz2:
From the linked issue, it's now possible with travis to explictly
select a newer pypy version to use, so pick the latest 5.10.0 one.


bz2:
Actually, seems travis aren't full up to date, but do have `5.8.0` at least and that's unversioned default is that rather than something ancient.

brian-brazil:
Thanks!

PR #242, opened 2018-02-14
Fix lint for line continuation indents

bz2:
Resolves E126, E127, and E128.

Drive-by slight improvement to CollectorRegistry error raising by
including all duplicated names in the message.


brian-brazil:
Thanks!

PR #243, opened 2018-02-15
process_collector: export swap size of a process

pborzenkov:
Swap size is a useful metric to have when trying to understand why some
service has suddenly became slower or its response latencies have grown.

@brian-brazil, please review.


brian-brazil:
This is not one of the standard exports that have the process_ prefix. Frankly if you start swapping you have already lost, and I would not enable swap on a modern server.

PR #245, opened 2018-02-16
Fix lint around unused and star imports

bz2:
Resolves F401, F403, and F405.

Removes unused imports. Makes test imports explicit and consistent.


tests/test_multiprocess.py
```diff
@@ -1,30 +1,43 @@
...
 import os
 import shutil
 import tempfile
-import time
 import unittest
 
-import prometheus_client
-from prometheus_client.core import *
-from prometheus_client.multiprocess import *
+from prometheus_client.core import (
```
brian-brazil: I'm not a big fan of these parens

bz2: I agree the import blocks aren't very nice, I'd generally just bring in the module and prefix everything, but for tests specifically it's at least clear what it being used by stating each name.

Specifically for parens, support was added in 2.4 and this multiline style is generally preferred:
https://www.python.org/dev/peps/pep-0328/
https://stackoverflow.com/questions/14376900/is-there-a-recommended-format-for-multi-line-imports

bz2:
Got too overzealous removing things, fixed.

brian-brazil:
Thanks!

PR #246, opened 2018-02-17
Fix lint related to vertical whitespace

bz2:
Resolves E123, E302, E303, E305, and E306.


tests/test_exposition.py
```diff
@@ -79,25 +80,31 @@ def test_escaping(self):
...
-        class MyCollector():
-            def collect(self):
-                metric = Metric("nonnumber", "Non number", 'untyped')
-                metric.add_sample("nonnumber", {}, MyNumber())
-                yield metric
         self.registry.register(MyCollector())
         self.assertEqual(b'# HELP nonnumber Non number\n# TYPE nonnumber untyped\nnonnumber 123.0\n', generate_latest(self.registry))
 
 
+class MyNumber(object):
```
brian-brazil: I'd prefer to keep these where they are so it's clear what's going on.

bz2: I think inlining helper classes to test methods is generally a bad idea as it pushes test authors to making bigger more complex methods to avoid duplicating components.

That said, have reversed the change and will push now.

brian-brazil:
Thanks!

PR #249, opened 2018-02-21
Make registry optional for MultiProcessCollector

marksteward:
@brian-brazil this is a trivial first PR to make it possible to make it easier to add more collectors during a multiproc scrape. This is so the pattern eventually can be:

```
    registry.register(MultiProcessCollector())
    registry.register(PlatformCollector())
    # add your own CustomCollectors here
```

instead of having to explicitly pass None to each Collector.

I would also have included the change for PlatformCollector, but I'm concerned that might break existing code. Should it be accompanied by deprecation warnings, or is it not worth doing?

### DIFF
--- a/prometheus_client/multiprocess.py
+++ b/prometheus_client/multiprocess.py
@@ -11,13 +11,13 @@ from . import core
 
 class MultiProcessCollector(object):
     """Collector for files for multi-process mode."""
-    def __init__(self, registry, path=None):
+    def __init__(self, registry=None, path=None):
         if path is None:
             path = os.environ.get('prometheus_multiproc_dir')
         if not path or not os.path.isdir(path):
             raise ValueError('env prometheus_multiproc_dir is not set or not a directory')
         self._path = path
-        if registry:
+        if registry is not None:
             registry.register(self)
 
     def collect(self):


brian-brazil:
> I would also have included the change for PlatformCollector, but I'm concerned that might break existing code. 

What is there to change?

brian-brazil:
It's usually expected that you will pass the registry to a collector, rather than having to register it yourself. I'm not sure what you are proposing is actually saving code.

marksteward:
> What is there to change?

To make PlatformCollector default to not registering itself with the default registry, except for the default instance. So:

```
    def __init__(self, registry=None, platform=None):
...
PLATFORM_COLLECTOR = PlatformCollector(core.REGISTRY)
```

This is more explicit and means someone creating their own PlatformCollector doesn't end up with one already registered by accident.

marksteward:
> It's usually expected that you will pass the registry to a collector, rather than having to register it yourself. I'm not sure what you are proposing is actually saving code.

This wasn't clear. The README has:

```
REGISTRY.register(CustomCollector())
```

but then:

```
```

When I first looked at this, I assumed MultiProcessCollector was doing some sort of quirky transformation on the registry, rather than just registering itself.

My hope is to make multiproc more self-explanatory and fit better with patterns used in Django, Flask, etc.

marksteward:
I'll raise a separate PR to improve the README.

PR #252, opened 2018-03-06
PEP333 does not require QUERY_STRING to be present.

daroot:
Ensure that if a WSGI server does not provide the QUERY_STRING in the
environ, that we don't throw an exception and cause a 500 to be returned
to the scraping client.

@brian-brazil


brian-brazil:
Thanks!

PR #257, opened 2018-03-21
Use `collections.defaultdict` to speed up metrics aggregation

umax:
Hi!

This is the next performance pull request. This change is about using `collections.defaultdict` for counting samples. It works much faster than `setdefault` method:
```
In [1]: from collections import defaultdict

In [2]: d={}
In [3]: %timeit d.setdefault('key', 0.0);d['key']+=1
197 ns ± 7.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
In [4]: d
Out[4]: {'key': 8111111.0}

In [5]: d=defaultdict(float)
In [6]: %timeit d['key']+=1
80.9 ns ± 0.841 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
In [7]: d
Out[7]: defaultdict(float, {'key': 81111111.0})
```

### DIFF
--- a/prometheus_client/multiprocess.py
+++ b/prometheus_client/multiprocess.py
@@ -2,6 +2,8 @@
 
 from __future__ import unicode_literals
 
+from collections import defaultdict
+
 import glob
 import json
 import os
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

PR #258, opened 2018-03-21
Use `ujson` as option for JSON processing

umax:
Hi!

The next one is about using high performance [json](https://github.com/esnme/ultrajson) library _ujson_ (as option).

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -3,7 +3,6 @@
 from __future__ import unicode_literals
 
 import copy
-import json
 import math
 import mmap
 import os
@@ -12,6 +11,11 @@ import struct
 import sys
 import time
 import types
+
+try:
+    import ujson as json
+except ImportError:
+    import json
 
 from threading import Lock
 from timeit import default_timer
--- a/prometheus_client/multiprocess.py
+++ b/prometheus_client/multiprocess.py
@@ -5,8 +5,12 @@ from __future__ import unicode_literals
 from collections import defaultdict
 
 import glob
-import json
 import os
+
+try:
+    import ujson as json
+except ImportError:
+    import json
 
 from . import core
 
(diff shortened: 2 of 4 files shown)


brian-brazil:
I avoid dependencies in this module, so I'd prefer not to add one if it isn't strictly needed.

umax:
This dependency is optional, when you install package like `pip install prometheus_client`, standard _json_ package will be used. I think, _ujson_ has a chance to be optional dependency for those, who really need high performance. _ujson_ is much faster than standard _json_:
```
In [4]: %timeit json.dumps(('name', (('l1', 'v1'), ('l2', 'v2')), 123))
3.85 µs ± 115 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [5]: %timeit ujson.dumps(('name', (('l1', 'v1'), ('l2', 'v2')), 123))
839 ns ± 13.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

brian-brazil:
From a support and maintenance standpoint I'd prefer to not have two different ways that users could be running things. I'd not be against doing something more custom if it's faster, I only use json for convenience.

umax:
I understand you, the more dependencies brings, the more unpredictable code is going to be. But for Python world it's common practice to try to use fast C implementations for different libraries instead of standard ones if it brings significant speed up.

I'd not propose these changes, but samples aggregation works slow for a huge dataset, that is why I'm trying to speed up it where possible. _ujson_ is stable for a long time and has a lot of tests. Again, _ujson_ set as optional.

So, it's up to you to merge these changes or not. In any way, thanks for quick reply and responsibility.

brian-brazil:
The json in CPython has been a fast C implementation for quite a long time now.

umax:
http://artem.krylysov.com/blog/2015/09/29/benchmark-python-json-libraries/

brian-brazil:
I'm not willing to take this particular change, but I am open to not using JSON.

PR #260, opened 2018-03-22
Use constants for metric types instead of hardcoded strings

umax:
I propose to use constants for metric types. Ideally, it would be nice to have separate modules for metrics, values, registries.


brian-brazil:
Python interns strings, so I wouldn't expect this to make a difference.

umax:
It's not about performance, it's about code quality and possible mistakes. Also, for Python 3 it brings type annotation possibility.

brian-brazil:
I'm happy with this code as-is.

PR #262, opened 2018-03-23
Prevent double reading all values when collect multiprocess metrics

umax:
In multiprocess mode, when reading metrics from files, there is no need to read all values in constructor, because `read_all_values` method called explicitly in multiprocess collector.


brian-brazil:
Thanks!

PR #263, opened 2018-03-26
Optimizations for tuples using. Removed unused types transformations.

umax:
Simple optimizations of tuple using.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -537,11 +537,11 @@ class _LabelWrapper(object):
         if labelkwargs:
             if sorted(labelkwargs) != sorted(self._labelnames):
                 raise ValueError('Incorrect label names')
-            labelvalues = tuple([unicode(labelkwargs[l]) for l in self._labelnames])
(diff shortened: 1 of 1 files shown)


prometheus_client/core.py
```diff
@@ -185,9 +185,8 @@ def __init__(self, name, documentation, value=None, labels=None):
         Metric.__init__(self, name, documentation, 'untyped')
         if labels is not None and value is not None:
             raise ValueError('Can only specify at most one of value and labels.')
-        if labels is None:
-            labels = []
-        self._labelnames = tuple(labels)
+
+        self._labelnames = tuple(labels or [])
```
brian-brazil: What performance difference does this make?

umax: No difference, just more shorted and more Pythonyc.
If you prefer, I can remove this changes from PR.

brian-brazil: It's fine as-is, and I'd expect it to be faster this way

umax: Updated.

brian-brazil:
Thanks!

PR #266, opened 2018-04-04
Use precompiled struct methods for pack, unpack data

umax:
Notes for struct usage optimizations: https://docs.python.org/2/library/struct.html#classes
Regarding string formatting:

```
In [1]: %timeit '%s' % 1
14.7 ns ± 0.325 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)

In [2]: %timeit '{0}'.format(1)
134 ns ± 2.93 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -29,6 +29,11 @@ _INF = float("inf")
 _MINUS_INF = float("-inf")
 _INITIAL_MMAP_SIZE = 1 << 20
 
+_pack_integer = struct.Struct(b'i').pack_into
+_pack_double = struct.Struct(b'd').pack_into
+_unpack_integer = struct.Struct(b'i').unpack_from
+_unpack_double = struct.Struct(b'd').unpack_from
+
 
 class CollectorRegistry(object):
     '''Metric collector registry.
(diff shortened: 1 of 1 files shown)


prometheus_client/core.py
```diff
@@ -29,6 +29,11 @@
 _MINUS_INF = float("-inf")
 _INITIAL_MMAP_SIZE = 1 << 20
 
+pack_integer = struct.Struct(b'i').pack_into
```
brian-brazil: put an _ in front of these so they aren't exported

prometheus_client/core.py
```diff
@@ -376,19 +381,26 @@ def _init_value(self, key):
...
         self._positions[key] = self._used - 8
 
     def _read_all_values(self):
         """Yield (key, value, pos). No locking is performed."""
+
         pos = 8
-        while pos < self._used:
-            encoded_len = struct.unpack_from(b'i', self._m, pos)[0]
+
+        # performance tricks
```
brian-brazil: Explain what you're doing a bit more. Not everyone will have seen the relevant talks by Raymond Hettinger

umax: https://wiki.python.org/moin/PythonSpeed/PerformanceTips

prometheus_client/core.py
umax: ok

prometheus_client/core.py
brian-brazil: I know why this works, but someone reading this code might not. A quick comment will help future readers of the code.

umax: fixed

brian-brazil:
Thanks!

Issue #80, opened 2016-05-17
ValueError when getting open file descriptors limit


anti1869:
Code in `process_collectors.py` relies on parsing `/proc/PID/limits
`  when gettings system limits. However, on some configurations, there could be "unlimited" string instead of a numeric value in that file. This lead to ValueError when this string is passed into `float` method.

Full example exception is as follows:

```
Traceback (most recent call last):
  File "/venv/lib/python3.5/site-packages/tornado/web.py", line 1443, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/venv/lib/python3.5/site-packages/video/stats/events/endpoint/endpoint.py", line 194, in get
    resp = generate_latest(core.REGISTRY)
  File "/venv/lib/python3.5/site-packages/prometheus_client/exposition.py", line 33, in generate_latest
    for metric in registry.collect():
  File "/venv/lib/python3.5/site-packages/prometheus_client/core.py", line 54, in collect
    for metric in collector.collect():
  File "/venv/lib/python3.5/site-packages/prometheus_client/process_collector.py", line 83, in collect
    'Maximum number of open file descriptors.', value=float(line.split()[3]))
ValueError: could not convert string to float: 'unlimited'
```

This happens in a Docker container with ubuntu, for instance, which has:

```
root@87220b5caf91:/app# cat /proc/76888/limits
Limit                     Soft Limit           Hard Limit           Units
Max cpu time              unlimited            unlimited            seconds
Max file size             unlimited            9223372036854775807  bytes
Max data size             unlimited            unlimited            bytes
Max stack size            10485760             35184372088832       bytes
Max core file size        unlimited            9223372036854775807  bytes
Max resident set          unlimited            1073741824           bytes
Max processes             unlimited            4000                 processes
Max open files            unlimited            65536                files
Max locked memory         unlimited            1073741824           bytes
Max address space         unlimited            unlimited            bytes
Max file locks            unlimited            unlimited            locks
Max pending signals       128                  512                  signals
Max msgqueue size         unlimited            8192                 bytes
```

brian-brazil:
Hmm, in that case we need to use the hard limit. I wonder if other languages have the same issue.

brian-brazil:
Also how is it possible for the soft limit to be higher than the hard limit? That sounds like a kernel bug.

anti1869:
[...]

PR #271, opened 2018-04-26
address prometheus/client_python#80

jgysland:
sdc-docker (Joyent's implementation of Docker that sits on top of their Triton orchestration platform) sometimes produces non-float values (the string `unlimited`) for open file descriptor limits. When they're cast at floats, a ValueError is raised. This commit catches that exception and passes.

This problem has been previously discussed in issue #80


brian-brazil:
This would cause the current used fds to not be reported.

brian-brazil:
This has gone stale, let us know if you're still interested in this.

PR #273, opened 2018-05-18
Fix lint related to operator positioning

bz2:
Resolves W503.

Also drive-by move definition of gauge multiprocess modes.


bz2:
Done, referring to the wiki page.

https://github.com/prometheus/prometheus/wiki/DCO-signing

Probably worth updating CONTRIBUTING.md with this, want me to do a pr for that?

brian-brazil:
Thanks!

That's needed in most repos, it's only in node exporter atm.

PR #274, opened 2018-05-19
Fix lint related to assigned but unused variables

bz2:
Resolves F841.

Also improves several tests to assert on relevant behaviours.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -53,7 +53,7 @@ class TestCounter(unittest.TestCase):
         try:
             f(True)
         except ValueError:
-            raised = True
+            pass
         self.assertEqual(1, self.registry.get_sample_value('c'))
 
     def test_block_decorator(self):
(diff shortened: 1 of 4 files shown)


tests/test_core.py
```diff
@@ -316,6 +316,7 @@ def test_invalid_names_raise(self):
 
     def test_empty_labels_list(self):
         h = Histogram('h', 'help', [], registry=self.registry)
+        h.observe(0)
```
brian-brazil: This changes the test

bz2: It does, but the test still fails if the change from commit e5e73e52 is reversed. It's really testing the constructor not recorded values, which is covered by `TestHistogram.test_histogram`.

Is there a better way of asserting "this isn't a `_LabelWrapper` that needs `labels()` called before using? I don't really want to use `isinstance()`.

brian-brazil: You could not assign to a variable

bz2: Okay, I wanted a better assertion but if you're fine with just checking the registry.

PR #282, opened 2018-06-07
Text parser optimization (~4.5x perf)

mfpierre:
Hi,

As we extensively use this library for parsing prometheus text format, we have observed that for big payloads (like https://github.com/kubernetes/kube-state-metrics for a big cluster) that can contains 400k+ lines the text parsing can be quite slow. (up to 27secs)

We tried to optimize the parser by dropping the state machine and leveraging native python functions such as `index` or `find` and it gives us an average of x5 performances.

Here are some benchmark using `timeit`:
```
call (x100000): _parse_sample('simple_metric 1.513767429e+09')
previous implementation: 1.10845804214
new implementation: 0.277444839478
improvement: x3.99523755508

call (x100000): _parse_sample('kube_service_labels{label_app="kube-state-metrics",label_chart="kube-state-metrics-0.5.0",label_heritage="Tiller",label_release="ungaged-panther",namespace="default",service="ungaged-panther-kube-state-metrics"} 1')
previous implementation: 7.58089208603
new implementation: 1.48280119896
improvement: x5.11254785291
```

For the KSM payload (400k lines) the parsing goes from ~27sec to ~4.7sec

**Note**: We could go up to almost 10x performance if we dropped some edge-cases treatment (like escaping, tab/space, etc...) could we consider a "strict" parsing mode that we could optionally use for "good citizens"?

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -182,12 +182,8 @@ class Metric(object):
                 self.samples == other.samples)
 
     def __repr__(self):
-        return str(self)
-
-    def __str__(self):
-        return "<Metric name: %s, documentation: %s, type: %s, samples: %s>" % (
-            self.name, self.documentation, self.type, self.samples)
-
+        return "Metric(%s, %s, %s, %s)" % (self.name, self.documentation,
+            self.type, self.samples)
 
 class UntypedMetricFamily(Metric):
     '''A single untyped metric and its samples.
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -181,6 +181,13 @@ class Metric(object):
                 self.type == other.type and
                 self.samples == other.samples)
 
+    def __repr__(self):
+        return str(self)
+
+    def __str__(self):
+        return "<Metric name: %s, documentation: %s, type: %s, samples: %s>" % (
+            self.name, self.documentation, self.type, self.samples)
+
 
 class UntypedMetricFamily(Metric):
     '''A single untyped metric and its samples.
--- a/prometheus_client/parser.py
+++ b/prometheus_client/parser.py
@@ -19,129 +19,99 @@ def text_string_to_metric_families(text):
     for metric_family in text_fd_to_metric_families(StringIO.StringIO(text)):
         yield metric_family
 
+def _replace_help_escaping(s):
+    return s.replace("\\n", "\n").replace('\\\\', '\\')
 
-def _unescape_help(text):
-    result = []
-    slash = False
+def _replace_escaping(s):
+    return s.replace("\\n", "\n").replace('\\\\', '\\').replace('\\"', '"')
 
-    for char in text:
-        if slash:
-            if char == '\\':
-                result.append('\\')
-            elif char == 'n':
-                result.append('\n')
-            else:
-                result.append('\\' + char)
-            slash = False
-        else:
-            if char == '\\':
-                slash = True
-            else:
-                result.append(char)
 
-    if slash:
-        result.append('\\')
+def _parse_labels(labels_string):
+    labels = {}
+    # Return if we don't have valid labels
+    if "=" not in labels_string:
+        return labels
 
-    return ''.join(result)
+    escaping = False
+    if "\\" in labels_string:
+        escaping = True
+
+    # Copy original labels
+    sub_labels = labels_string
+    try:
+        # Process one label at a time
+        while sub_labels:
+            # The label name is before the equal
+            value_start = sub_labels.index("=")
+            label_name = sub_labels[:value_start]
+            sub_labels = sub_labels[value_start + 1:].lstrip()
+            # Find the first quote after the equal
+            quote_start = sub_labels.index('"') + 1
+            value_substr = sub_labels[quote_start:]
+
+            # Find the last unescaped quote
+            i = 0
+            while i < len(value_substr):
+                i = value_substr.index('"', i)
+                if value_substr[i - 1] != "\\":
+                    break
+                i += 1
+
+            # The label value is inbetween the first and last quote
+            quote_end = i + 1
+            label_value = sub_labels[quote_start:quote_end]
+            # Replace escaping if needed
+            if escaping:
+                label_value = _replace_escaping(label_value)
+            labels[label_name.strip()] = label_value.strip()
+
+            # Remove the processed label from the sub-slice for next iteration
+            sub_labels = sub_labels[quote_end + 1:]
+            next_comma = sub_labels.find(",") + 1
+            sub_labels = sub_labels[next_comma:].lstrip()
+
+        return labels
+
+    except ValueError:
+        raise ValueError("Invalid labels: %s" % labels_string)
+
+
+# If we have multiple values only consider the first
+def _parse_value(s):
+    s = s.lstrip()
+    separator = " "
+    if separator not in s:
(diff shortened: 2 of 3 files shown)


brian-brazil:
> could we consider a "strict" parsing mode that we could optionally use for "good citizens"?

Those aspects of the text format are not optional, they must be implemented to have a correct parser.

prometheus_client/core.py
```diff
@@ -181,6 +181,10 @@ def __eq__(self, other):
                 self.type == other.type and
                 self.samples == other.samples)
 
+    def __repr__(self):
```
brian-brazil: A __str__ would make more sense I think

prometheus_client/parser.py
```diff
@@ -20,128 +20,96 @@ def text_string_to_metric_families(text):
...
-                result.append('\n')
-            else:
-                result.append('\\' + char)
-            slash = False
-        else:
-            if char == '\\':
-                slash = True
-            else:
-                result.append(char)
+def _replace_escaping(s):
```
brian-brazil: Help and label values have different escaping rules (double quote is the difference), you need two functions for this.

prometheus_client/parser.py
```diff
@@ -20,128 +20,96 @@ def text_string_to_metric_families(text):
...
+    return s.replace("\\n", "\n").replace('\\\\', '\\').replace('\\"', '"')
 
-    if slash:
-        result.append('\\')
 
-    return ''.join(result)
+def _parse_labels(labels_string):
+    labels = {}
+    # return if we don't have valid labels
```
brian-brazil: Please keep comments as full sentances

prometheus_client/parser.py
```diff
@@ -20,128 +20,96 @@ def text_string_to_metric_families(text):
...
+        name = text[:label_start].strip()
+        # we ignore the starting curly brace
+        label = text[label_start + 1:label_end]
+        # the value is after the label end (ignoring curly brace and space)
+        value = float(_parse_value(text[label_end + 2:]))
+        return name, _parse_labels(label), value
+
+    # we don't have labels
+    except ValueError:
+        # detect what separator is used
```
brian-brazil: Any mix of any number of spaces and tabs is permitted

prometheus_client/parser.py
```diff
@@ -20,128 +20,96 @@ def text_string_to_metric_families(text):
...
-    value = []
-    labels = {}
-
-    state = 'name'
+    # detect the labels in the text
+    try:
+        label_start, label_end = text.index("{"), text.rindex("}")
+        # the name is before the labels
+        name = text[:label_start].strip()
+        # we ignore the starting curly brace
```
brian-brazil: there can be whitespace after the brace, and basically everywhere else between things

prometheus_client/parser.py
```diff
@@ -20,128 +20,96 @@ def text_string_to_metric_families(text):
...
-    name = []
-    labelname = []
-    labelvalue = []
-    value = []
-    labels = {}
-
-    state = 'name'
+    # detect the labels in the text
+    try:
+        label_start, label_end = text.index("{"), text.rindex("}")
```
brian-brazil: A label value could contain a }

prometheus_client/parser.py
```diff
@@ -20,128 +20,96 @@ def text_string_to_metric_families(text):
...
+        # the name is before the labels
+        name = text[:label_start].strip()
+        # we ignore the starting curly brace
+        label = text[label_start + 1:label_end]
+        # the value is after the label end (ignoring curly brace and space)
```
brian-brazil: There can be a trailing comma after the last "

prometheus_client/core.py
mfpierre: renamed it to `__str__` and having `repr` call `str` because it's useful for comparing unit test output:

```
E       First differing element 0:
E       <Metric name: a, documentation: help, type: counter, samples: [(u'a', {u'foo': u'bar'}, 1), (u'a', {u'foo': u'baez'}, 2), (u'a', {u'foo': u'buz'}, 3)]>
E       <Metric name: a, documentation: help, type: counter, samples: [(u'a', {u'foo': u'bar'}, 1.0), (u'a', {u'foo': u'baz'}, 2.0), (u'a', {u'foo': u'buz'}, 3.0)]>
E
E       - [<Metric name: a, documentation: help, type: counter, samples: [(u'a', {u'foo': u'bar'}, 1), (u'a', {u'foo': u'baez'}, 2), (u'a', {u'foo': u'buz'}, 3)]>]
E       ?                                                                                                                  -
E
E       + [<Metric name: a, documentation: help, type: counter, samples: [(u'a', {u'foo': u'bar'}, 1.0), (u'a', {u'foo': u'baz'}, 2.0), (u'a', {u'foo': u'buz'}, 3.0)]>]
E       ?
```
instead of 
```
E       First differing element 0:
E       <prometheus_client.core.CounterMetricFamily object at 0x108018c50>
E       <prometheus_client.core.Metric object at 0x108018cd0>
E
E       - [<prometheus_client.core.CounterMetricFamily object at 0x108018c50>]
E       ?                          -------      ------                    ^
E
E       + [<prometheus_client.core.Metric object at 0x108018cd0>]
```

prometheus_client/parser.py
mfpierre: This already taken into account with the use of `rindex`, added a test case to validate this point.

prometheus_client/parser.py
mfpierre: this should be covered, I added one additional test case in `test_spaces`

prometheus_client/parser.py
mfpierre: should be already covered, `test_commas` is validating this

prometheus_client/parser.py
mfpierre: Added some additional unit tests around this 👍

prometheus_client/parser.py
mfpierre: added another function specific for `Help`

prometheus_client/core.py
brian-brazil: Usually str calls repr rather than the other way around. repr should usually be an instantiatable version of object, while str is more human readable.

prometheus_client/parser.py
```diff
@@ -19,129 +19,99 @@ def text_string_to_metric_families(text):
...
+            sub_labels = sub_labels[value_start + 1:].lstrip()
+            # Find the first quote after the equal
+            quote_start = sub_labels.index('"') + 1
+            value_substr = sub_labels[quote_start:]
+
+            # Find the last unescaped quote
+            i = 0
+            while i < len(value_substr):
+                i = value_substr.index('"', i)
+                if value_substr[i - 1] != "\\":
```
brian-brazil: What if if you have `x=""` as the label? `i - 1` will be -1, which might have unexpected results

mfpierre: added a unit-test around empty labels, but this works fine 👍

prometheus_client/core.py
mfpierre: Modified `repr` to fit what it's supposed to be, still way more readable for tests outputs 👍

brian-brazil:
Thanks!

mfpierre:
Hey @brian-brazil, any plans to do a release soon? 🙇

PR #285, opened 2018-06-21
added support for CustomAsyncCollector

blenessy:
Please see the updated README.md for this improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,40 @@ registration time instead of `describe`. If this could cause problems, either
 implement a proper `describe`, or if that's not practical have `describe`
 return an empty list.
 
+## Async Collectors
+
+Custom collectors might need to sample from an URL, which is typically a blocking
+operation in python. With only one custom collector, this blocking operation
+can be done implicitly in the `collect` method. However, with multiple collectors,
+sampling in `collect` will block each collector sequentially.
+We can add an `async_refresh` method to a custom collector, to work with the
+standard `asyncio` package, which well be called **before** `collect` is called
+in any of the custom collectors. This facilitates means to kick-off **all**
+async tasks in parallell, and thus introducing minimal delay.
+
+Here is the anatomy of a custom async collector:
+
+```python
+import asyncio
+
+from prometheus_client.core import GaugeMetricFamily, CounterMetricFamily, REGISTRY
+
+class CustomAsyncCollector(object):
+    def __init__(self):
+        self._async_task = None
+
+    def async_refresh(self):
+        self._async_task = asyncio.get_event_loop().create_task( ... )
+
+    def collect(self):
+        asyncio.get_event_loop().run_until_complete(self._async_task)
+        if self._async_task.exception() is not None:
+          result = self._async_task.result()
+          yield ...
+
+REGISTRY.register(CustomAsyncCollector())
+```
+
 
 ## Multiprocess Mode (Gunicorn)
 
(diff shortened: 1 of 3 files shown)


brian-brazil:
I can see what you're trying to do, but am unsure if this belongs in the core from an API cleanliness standpoint.

This particular solution has races in any case, as concurrent scrapes can occur.

blenessy:
Thanks @brian-brazil for the quick feedback!

For more context I'm adding one of our CustomAsyncCollectors:

```python
import aiohttp
import async_timeout
import asyncio
import json
import logging
import os

from prometheus_client.core import CounterMetricFamily

__ALL__ = ("CoreSystemCollector",)

BASE_URL = os.environ.get("RADIXDLT_CORE_BASE_URL", "https://localhost")
VALIDATE_CERT = os.environ.get("RADIXDLT_CORE_VALIDATE_CERT", "no") == "yes"

async def async_http_get(session, url, timeout=10):
    with async_timeout.timeout(timeout):
        async with session.get(url) as response:
            assert response.status == 200
            return await response.text()

class CoreSystemCollector(object):
    URL = "{0}/api/system".format(BASE_URL)

    def __init__(self):
        self._loop = asyncio.get_event_loop()
        self._session = aiohttp.ClientSession(loop=self._loop, connector=aiohttp.TCPConnector(verify_ssl=VALIDATE_CERT))
        self._async_task = None

    def async_refresh(self):
        try:
            self._async_task.cancel()  # cancel previous requests if any
        except AttributeError:
            pass  # 'NoneType' object has no attribute 'cancel'
        self._async_task = self._loop.create_task(async_http_get(self._session, self.URL))

    def collect(self):  # pylint: disable=no-self-use
        if self._async_task is None:
            self.async_refresh()  # backward compatibility

        self._loop.run_until_complete(self._async_task)
        if self._async_task.exception() is None:
            metrics = json.loads(self._async_task.result())
            yield CounterMetricFamily('radixdlt_system_memory_max',
                                      'Maximum memory available for the radixdlt core on this node',
                                      value=metrics["memory"]["max"])
        else:
            logging.error("failed to scrape metics from the core", exc_info=True)

```

In regards to race conditions, not sure what you mean ... but wouldn't `self._async_task.cancel()` in the example prevent that ?

In regards to keeping this outside of the core - I tried the following, but then my collector will not be called:

```pyhton
class MyCollectorRegistry(prometheus_client.core.CollectorRegistry):
    def collect(self):
        '''Yields metrics from the collectors in the registry.'''
        collectors = None
        with self._lock:
            collectors = copy.copy(self._collector_to_names)
        # call async_refresh to kick off async refresh tasks
        for collector in collectors:
            async_refresh = getattr(collector, "async_refresh", None)
            if callable(async_refresh):
                collector.async_refresh()
        # collect fresh metrics
        for collector in collectors:
                yield metric

prometheus_client.core.REGISTRY = MyCollectorRegistry()
prometheus_client.core.REGISTRY.register(CoreSystemCollector())
```

What am I missing ?

brian-brazil:
> What am I missing ?

The reason you probably need this would be for writing an exporter, so you could do everything you need inside a single collector (which is how most exporters are written in any case).

> but wouldn't self._async_task.cancel() in the example prevent that ?

You've still got a race, as one of the requests will now fail.

Also, radixdlt_system_memory_max is a gauge.

blenessy:
ahh, very good tips. Thanks again for putting me straight.

PR #288, opened 2018-07-11
Multithreading fixes

### LINKED ISSUE #287: time() decorator is not thread-safe?
Good day,

I have a suspicion that time() decorator is not thread-safe in some way. Consider the following code:

```python
import sys
import time
from concurrent.futures import ThreadPoolExecutor

S_TIME = Summary("timing_with_time", "Time spent processing request")
S_MANUAL = Summary("timing_manually", "Time spent processing request")

@S_TIME.time()
def process_request():
    start = default_timer()
    time.sleep(1)
    S_MANUAL.observe(max(default_timer() - start, 0))
    
    
if __name__ == "__main__":
    start_http_server(8000)

    threads = int(sys.argv[1])
        
    pool = ThreadPoolExecutor(max_workers=threads)
    for i in range(1000):
        pool.submit(process_request)
```

Running it as `python bug.py 1` produce the following metrics:
```
timing_with_time_count 12.0
timing_with_time_sum 12.013219268759713
timing_manually_count 12.0
timing_manually_sum 12.012461822945625
```
I.e. both _sum values are similar as expected. However if I increase the concurrency, then the metric that is
tracked with `time()` does NOT have its _sum increasing:

With `python bug.py 2`:
```
timing_with_time_count 10.0
timing_with_time_sum 5.004416585084982
timing_manually_count 10.0
timing_manually_sum 10.009211090975441
```
I.e. the count matches, but the sum of the first one is wrong! I did 10 requests each 1 second long, not 0.5 second long. The problem scales linearly. E.g. with `python bug.py 10`:
```
timing_with_time_count 60.0
timing_with_time_sum 6.0030247694812715
timing_manually_count 60.0
timing_manually_sum 60.06247438094579
```
I.e. _sum metric measured with `time()` decorator always thinks it's single threaded.

Python 3.6.5 on ubuntu 16.04 64-bit, P8s client version 0.2.0 (also happens with 0.1.1)

haizaar:
Fixes #287

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -972,7 +972,7 @@ class _Timer(object):
     def __init__(self, callback):
         self._callback = callback
         self._storage = local()
-        self.key = "k_{}".format(id(self))
+        self.key = "k_{1}".format(id(self))
 
     def __enter__(self):
         setattr(self._storage, self.key, default_timer())
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -13,7 +13,7 @@ import sys
 import time
 import types
 
-from threading import Lock
+from threading import local, Lock
 from timeit import default_timer
 
 from .decorator import decorate
@@ -765,7 +765,7 @@ class Gauge(object):
 
         Can be used as a function decorator or context manager.
         '''
-        return _GaugeTimer(self)
+        return _Timer(self.set)
 
     def set_function(self, f):
         '''Call the provided function to return the Gauge value.
@@ -829,7 +829,7 @@ class Summary(object):
 
         Can be used as a function decorator or context manager.
         '''
-        return _SummaryTimer(self)
+        return _Timer(self.observe)
 
     def _samples(self):
         return (
@@ -919,7 +919,7 @@ class Histogram(object):
 
         Can be used as a function decorator or context manager.
         '''
-        return _HistogramTimer(self)
+        return _Timer(self.observe)
 
     def _samples(self):
         samples = []
@@ -930,24 +930,6 @@ class Histogram(object):
         samples.append(('_count', {}, acc))
         samples.append(('_sum', {}, self._sum.get()))
         return tuple(samples)
-
-
-class _HistogramTimer(object):
-    def __init__(self, histogram):
-        self._histogram = histogram
-
-    def __enter__(self):
-        self._start = default_timer()
-
-    def __exit__(self, typ, value, traceback):
-        # Time can go backwards.
-        self._histogram.observe(max(default_timer() - self._start, 0))
-
-    def __call__(self, f):
-        def wrapped(func, *args, **kwargs):
-            with self:
-                return func(*args, **kwargs)
-        return decorate(f, wrapped)
 
 
 class _ExceptionCounter(object):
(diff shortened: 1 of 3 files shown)


haizaar:
@brian-brazil 
What am I to do about DCO for my past commits?

prometheus_client/core.py
```diff
@@ -986,34 +968,18 @@ def wrapped(func, *args, **kwargs):
...
-        return decorate(f, wrapped)
-
-
-class _GaugeTimer(object):
-    def __init__(self, gauge):
-        self._gauge = gauge
+class _Timer(object):
+    def __init__(self, callback):
+        self._callback = callback
+        self._storage = local()
```
brian-brazil: Is this safe for reentrant use within the same thread?

brian-brazil:
All commits needs the DCO, it's easiest to squash.

prometheus_client/core.py
haizaar: You have a point here - `local()` returns reference to the same object within a thread and therefore it will work incorrectly, particularly for nested timers or regular timers in async code. I'm thinking about using `id(self)` as a local storage key.
What do you think?

brian-brazil: If `id(self)` would work, then we would be able to store in the `self` directly as it's just the memory address of `self`.

haizaar: It will bring us back to the original problem when a single instance of timer used in different threads.

haizaar:
Updated the code. Lets converge to LGTM and then I'll fix the DCO.

prometheus_client/core.py
brian-brazil: My point is that `id(self)` doesn't help, as you wouldn't be seeing a race in the first place if that worked. So this doesn't fix reentrancy. We may need a stack here.

haizaar: OK, lets define the problem :)
Having
```
s1 = Summary("s1",...)
s2 = Summary("s2",...)
```

The original issue is that:
```
@s1.time()
def f(): pass
```
Produces wrong timings when `f()` is called simultaneously from different threads. My last code solves inter thread issue using TLS, and uses `id(self)` as TSL key to support multiple timer instances within a thread. Such that both `f()` and `g()` (below) work
```
@s1.time()
def g():
    s2.time()(lambda : time.sleep(1))()
```
And both `f()` and `g()` cases are covered with tests.

Now, what do you refer as reentrancy? This? -
```
@s1.time()
def h():
    @s1.time()
    def i(): pass
    i()
```
This will work as well, since each call to `time()` will produce different instances of the `_Time()` with distinct `id(self)`.

P.S. I actually think that the better and simpler idea is to rework both decorator and context manager to use function closure to record the start time. It will be also a bit faster IMHO. I need to think whether its possible to implement it without changing the `time()` semantics.

brian-brazil: Reentrant means that the function calls itself, e.g.:

```
def f():
  @s1.time():
     f()
```

haizaar: What's the use case?
Recursion like this?
```
@s1.time()
def f(i=0):
    f(i+1) if i > 5 else True
```
Then no, `_Timer` is not reentrant. But neither is the current implementation (as in master).

haizaar:
Closing in favor of #290

PR #292, opened 2018-07-25
Add ssl support to start_http_server

yrivardmulrooney:


brian-brazil:
Thanks for your PR. As a matter of policy we currently don't support any server-side auth/encryption features (see https://prometheus.io/docs/operating/security/#authentication/authorisation/encryption). Accordingly if you want to do something more custom, I'd suggest duplicating this function in your own code. start_http_server is meant more to get people up and running quickly, rather than as a full-on solution.

PR #294, opened 2018-07-31
Added set() to Counter so it is now possible to set an specific value

ffknob:
I (think I ) needed this because I will not track the counter through my app processing. I'll get the value from a DB table where another processes save results. I can not change these other processes to track the counter, so what I wanted to do is to watch the results direct from the log table and send the deterministic value to Prometheus.


ffknob:
DCO signed

brian-brazil:
Please see https://www.robustperception.io/setting-a-prometheus-counter

PR #297, opened 2018-08-17
Example on how to expose metrics in a flask app

banjoh:
I'm working on a flask application that needs to expose metrics to prometheus. It was not very clear how that could be done without reading the code. This PR updates the README with a simple example of doing so. Hopefully it saves someone some time of trying to figure that out.

@brian-brazil

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,36 @@ from prometheus_client import start_wsgi_server
 start_wsgi_server(8000)
 ```
 
+#### Flask
+
+To use Prometheus with [Flask](http://flask.pocoo.org/) we need to serve metrics through a Prometheus WSGI application. This can be achieved using [Flask's application dispatching](http://flask.pocoo.org/docs/latest/patterns/appdispatch/). Below is a working example.
+
+Save the snippet below in a `myapp.py` file
+
+```python
+from flask import Flask
+from werkzeug.wsgi import DispatcherMiddleware
+from prometheus_client import make_wsgi_app
+
+# Create my app
+app = Flask(__name__)
+
+# Add prometheus wsgi middleware to route /metrics requests
+app_dispatch = DispatcherMiddleware(app, {
+    '/metrics': make_wsgi_app()
+})
+```
+
+Run the example web application like this
+
+```bash
+# Install uwsgi if you do not have it
+pip install uwsgi
+uwsgi --http 127.0.0.1:8000 --wsgi-file myapp.py --callable app_dispatch
+```
+
+Visit http://localhost:8000/metrics to see the metrics
+
 ### Node exporter textfile collector
 
 The [textfile collector](https://github.com/prometheus/node_exporter#textfile-collector)


README.md
```diff
@@ -254,6 +254,30 @@ reactor.listenTCP(8000, factory)
...
+
+```python
+from flask import Flask, Response, request
+from prometheus_client import generate_latest, CONTENT_TYPE_LATEST, core
+
+app = Flask(__name__)
+
+@app.route("/metrics")
+def metrics():
+    r = core.REGISTRY
```
brian-brazil: I don't want this logic duplicated here, I want you to use one of the existing functions for it. This code will change over time, and copy&pasted code in user applications won't get updated.

banjoh: Do you mean using like so?

```python
from prometheus_client import REGISTRY
def metrics():
    r = REGISTRY
```

If not, I'm probably missing the point. I my initial implementation I was hoping/expecting to find a clean function that returns a response having parsed out the parameters and done its magic. A new public api could be added that does something like this

```python
def generate_latest(registry, query_string)
    """Generate the latest metrics taking the query string into account"""
    pass
```

brian-brazil: No, I mean reusing something like MetricsHandler that does all the http stuff.

banjoh: Since flask's response is a uwsgi application, I'll try to see how I can reuse ```make_wsgi_app``` as is.

README.md
```diff
@@ -254,7 +254,57 @@ reactor.listenTCP(8000, factory)
 reactor.run()
 ```
 
-#### WSGI
+#### Flask
+
+To use prometheus with [flask](http://flask.pocoo.org/) we need to serve metrics through a promethues wsgi application. This can be achieved using [flask's application dispatching|http://flask.pocoo.org/docs/latest/patterns/appdispatch/]. Below is a working example.
```
brian-brazil: Flask

README.md
```diff
@@ -254,7 +254,57 @@ reactor.listenTCP(8000, factory)
 reactor.run()
 ```
 
-#### WSGI
```
brian-brazil: Wsgi should stay first

README.md
```diff
@@ -254,7 +254,57 @@ reactor.listenTCP(8000, factory)
...
+app_dispatch = DispatcherMiddleware(app, {
+    '/metrics': make_wsgi_app()
+})
+
+@app.route('/')
+def hello():
+    return "Hello world !!"
+
+# Expose some basic metrics of the application
+REQUEST_LATENCY = Histogram('request_latency_seconds', 'API Request Latency', ['method', 'endpoint'])
```
brian-brazil: Keep this minimal, just enough to expose metrics at all like in the other examples.

banjoh: Done

README.md
```diff
@@ -280,6 +280,34 @@ from prometheus_client import start_wsgi_server
 start_wsgi_server(8000)
 ```
 
+#### Flask
+
+To use prometheus with [flask](http://flask.pocoo.org/) we need to serve metrics through a promethues wsgi application. This can be achieved using [flask's application dispatching](http://flask.pocoo.org/docs/latest/patterns/appdispatch/). Below is a working example.
```
brian-brazil: Flask Prometheus WSGI

README.md
```diff
@@ -280,6 +280,34 @@ from prometheus_client import start_wsgi_server
...
+# Create my app
+app = Flask(__name__)
+
+# Add prometheus wsgi middleware to route /metrics requests
+app_dispatch = DispatcherMiddleware(app, {
+    '/metrics': make_wsgi_app()
+})
+```
+
+To test this example, save the above snippet in a `myapp.py` file and run the command below
```
brian-brazil: We mention filenames before the relevant example.

banjoh: Done

README.md
brian-brazil: Proper nouns should be capitalised.

banjoh: Done

README.md
```diff
@@ -280,6 +280,36 @@ from prometheus_client import start_wsgi_server
 start_wsgi_server(8000)
 ```
 
+#### Flask
+
+To use Prometheus with [Flask](http://flask.pocoo.org/) we need to serve metrics through a Promethues WSGI application. This can be achieved using [Flask's application dispatching](http://flask.pocoo.org/docs/latest/patterns/appdispatch/). Below is a working example.
```
brian-brazil: s/Promethues//

banjoh: Oops :) Typo fixed

brian-brazil:
Thanks!

PR #301, opened 2018-08-31
Add detailed Garbage Collection statistics.

tcolgate:
We've found this useful in monitoring pythons Garbage Collection. If you are
interested in including this:

- Should the name default to python_gc perhaps?
- Are you happy for these all to be histograms?
- If so, how would you prefer the buckets to be set up (presumably defaulted and
  pass in to the constructor?)
- I'm not actually sure if having uncollected as a histogram is of much use. Uncollected
  should be monitored, but histogram is probably overkill.
- I'll fill out the tests?

### DIFF
--- a/prometheus_client/gc_collector.py
+++ b/prometheus_client/gc_collector.py
@@ -1,0 +1,59 @@
+#!/usr/bin/python
+
+from __future__ import unicode_literals
+
+import gc
+import time
+
+from . import core
+
+class GCCollector(object):
+    """Collector for Garbage collection statistics."""
+    def __init__(self, registry=core.REGISTRY, gc=gc):
+        if not hasattr(gc, 'callbacks'):
+            return
+
+        collected = core.Histogram(
+           'python_gc_collected_objects',
+           'Objects collected during gc',
+           ['generation'],
+           buckets=[500, 1000, 5000, 10000, 50000],
+           registry=registry
+        )
+
+        uncollectable = core.Histogram(
+           'python_gc_uncollectable_objects',
+           'Uncollectable object found during GC',
+           ['generation'],
+           buckets=[500, 1000, 5000, 10000, 50000],
+           registry=registry
+        )
+
+        latency = core.Histogram(
+           'python_gc_duration_seconds',
+           'Time spent in garbage collection',
+           ['generation'],
+           registry=registry
+        )
+
+        times = {}
+
+        def _cb(phase, info):
+            gen = info['generation']
+
+            if phase == 'start':
+                times[gen] = time.time()
+
+            if phase == 'stop':
+                delta = time.time() - times[gen]
+                latency.labels(gen).observe(delta)
+                if 'collected' in info:
+                    collected.labels(gen).observe(info['collected'])
+                if 'uncollectable' in info:
+                    uncollectable.labels(gen).observe(info['uncollectable'])
+
+        gc.callbacks.append(_cb)
+
+
+GC_COLLECTOR = GCCollector()
+"""Default GCCollector in default Registry REGISTRY."""
--- a/prometheus_client/__init__.py
+++ b/prometheus_client/__init__.py
@@ -4,6 +4,7 @@ from . import core
 from . import exposition
 from . import process_collector
 from . import platform_collector
+from . import gc_collector
 
 __all__ = ['Counter', 'Gauge', 'Summary', 'Histogram', 'Info', 'Enum']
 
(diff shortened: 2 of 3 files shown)


prometheus_client/gc_collector.py
```diff
@@ -0,0 +1,69 @@
...
+
+import gc as gc_module
+import os
+import time
+
+from . import core
+
+class GCCollector(object):
+    """Collector for Garbage collection statistics."""
+    def __init__(self, namespace='', registry=core.REGISTRY, gc=gc_module):
```
brian-brazil: Why is the namespace configurable?

prometheus_client/gc_collector.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/python
+
+from __future__ import unicode_literals
+
+import gc as gc_module
```
brian-brazil: Does this work on all Python implementations?

prometheus_client/gc_collector.py
```diff
@@ -0,0 +1,69 @@
...
+        uncollectable = core.Histogram(
+           self._prefix + 'uncollectable_objects',
+           'uncollectable object found during GC',
+           ['generation'],
+           buckets=[500, 1000, 5000, 10000, 50000],
+           registry=registry
+        )
+
+        latency = core.Histogram(
+           self._prefix + 'latency_seconds',
```
brian-brazil: duration would be more typical

prometheus_client/gc_collector.py
```diff
@@ -0,0 +1,69 @@
...
+           'Time spent in garbage collection',
+           ['generation'],
+           buckets=[0.001, 0.01, 0.05, 0.1, 0.5, 1.0, 2.0],
```
brian-brazil: This is close enough to the defaults to stick with them

tcolgate:
@brian-brazil removed the configurable namespace (this was cargo-culted from another module in this package). I've also removed the bucket specification for the timings.

re: other packages. The gc.callbacks was added in python 3.3, the gc package is certainly in 2.7 and the docs allude to it having been in as far as 2.2.

I've tested the basic import of gc on jython (2.5.3 available on debian), and this appears to work without problems.

What is your minimal supported version?

tcolgate:
@brian-brazil udpated to include a test.
pypy has it's own gc, those stats wont get picked up here. Perhaps an interested party will come along.

prometheus_client/gc_collector.py
```diff
@@ -0,0 +1,62 @@
...
+
+from . import core
+
+class GCCollector(object):
+    """Collector for Garbage collection statistics."""
+    def __init__(self, registry=core.REGISTRY, gc=gc_module):
+        if not hasattr(gc, 'callbacks'):
+            return
+
+        _prefix = 'python_gc_'
```
brian-brazil: I prefer to have all the name together, makes it easier to grep

prometheus_client/gc_collector.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/python
+
+from __future__ import unicode_literals
+
+import gc as gc_module
```
brian-brazil: There's no "gc" import you're shadowing here, so no need to rename.

prometheus_client/gc_collector.py
```diff
@@ -0,0 +1,62 @@
...
+import os
```
brian-brazil: This is unused

prometheus_client/gc_collector.py
tcolgate: do you mean that I should drop the prefix and just have the complete names in the registration?

prometheus_client/gc_collector.py
tcolgate: okay, I thought it made it a bit clearer when default the gc passed to the __init__ (which I needed for the tests)

tcolgate:
It would break if no gc module exists, but jython, python and pypy all have one, I can probably do something to work around the eventuality if you prefer.

prometheus_client/gc_collector.py
brian-brazil: Yes, have complete names rather than building them up.

brian-brazil:
If the main implementations have one since 2.6 then we're good for now, we can wait until someone complains to see if it's an actual problem.

tcolgate:
The docs for the gc module in 2.7 have some functions labeled as "new in 2.2", and I'd seen some other references to 2.2 as well.

tcolgate:
@brian-brazil I've rebased this on master. Reviewed all the comments and , so far as I can tell, they are all resolved.

prometheus_client/gc_collector.py
```diff
@@ -0,0 +1,59 @@
...
+
+class GCCollector(object):
+    """Collector for Garbage collection statistics."""
+    def __init__(self, registry=core.REGISTRY, gc=gc):
+        if not hasattr(gc, 'callbacks'):
+            return
+
+        collected = core.Histogram(
+           'python_gc_collected_objects',
+           'objects collected during gc',
```
brian-brazil: Start with a capital letter

tcolgate:
@brian-brazil updated.

brian-brazil:
Thanks!

Issue #275, opened 2018-05-21
Gunicorn/multiprocess with restarting workers, .db files are not cleaned up


Dominick-Peluso-Bose:
We are using Gunicorn with multiple workers with the Gunicorn `max_requests` option. We found that memory usage in our containers increases with every worker restart. By reducing the `max_requests` value so workers restart very frequently, the problem became apparent. 

Using the method described in the readme:

```py
from prometheus_client import multiprocess

def child_exit(server, worker):
    multiprocess.mark_process_dead(worker.pid)
```

While running a performance test, I see the following in the container:
```bash
[root@mock-adapter-8568fdcf6d-lz8tx app]# cd /app/prometheus_tmp
[root@mock-adapter-8568fdcf6d-lz8tx prometheus_tmp]# ls
counter_10.db   counter_214.db  counter_319.db  counter_41.db   counter_91.db     histogram_196.db  histogram_296.db  histogram_401.db  histogram_69.db
counter_111.db  counter_222.db  counter_31.db   counter_425.db  counter_97.db     histogram_1.db    histogram_305.db  histogram_409.db  histogram_77.db
counter_117.db  counter_232.db  counter_328.db  counter_433.db  counter_9.db      histogram_202.db  histogram_311.db  histogram_419.db  histogram_84.db
counter_130.db  counter_23.db   counter_338.db  counter_444.db  histogram_10.db   histogram_214.db  histogram_319.db  histogram_41.db   histogram_91.db
counter_138.db  counter_244.db  counter_345.db  counter_450.db  histogram_111.db  histogram_222.db  histogram_31.db   histogram_425.db  histogram_97.db
counter_148.db  counter_251.db  counter_353.db  counter_458.db  histogram_117.db  histogram_232.db  histogram_328.db  histogram_433.db  histogram_9.db
counter_154.db  counter_258.db  counter_361.db  counter_470.db  histogram_130.db  histogram_23.db   histogram_338.db  histogram_444.db
counter_175.db  counter_269.db  counter_36.db   counter_48.db   histogram_138.db  histogram_244.db  histogram_345.db  histogram_450.db
counter_187.db  counter_277.db  counter_374.db  counter_56.db   histogram_148.db  histogram_251.db  histogram_353.db  histogram_458.db
counter_18.db   counter_286.db  counter_384.db  counter_61.db   histogram_154.db  histogram_258.db  histogram_361.db  histogram_470.db
[...]

PR #302, opened 2018-09-05
Refactor MultiProcessCollector.collect() to allow for arbitrary merging.

bloodearnest:
Factors out a merge() method from the previous collect() method, which
is parameterized, and thus can be used for arbitrary merging of samples.
For motivation, see discussion in issue #275 around merging dead worker's
data into a single mmaped file.

This basically allows us to parameterize the files to be merged, and
also whether to accumulate histograms or not. Accumulation is on by
default, as that is what the prometheus format expects. But it can now
be disabled, which allows merged values to be correctly written back to
an mmaped file. For the same reason, the order of labels is preserved
via OrderedDict.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -1016,6 +1016,11 @@ def _floatToGoString(d):
         return '-Inf'
     elif math.isnan(d):
         return 'NaN'
+    elif sys.version_info[:2] == (2, 6):
+        format = '%g' % float(d)
+        if isinstance(d, float) and '.' not in format:
+            format += '.0'
+        return format
     else:
         return repr(float(d))
 
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -586,6 +586,13 @@ class _MmapedDict(object):
             self._f = None
 
 
+def _mmap_key(metric_name, name, labelnames, labelvalues):
+    """Format a key for use in the mmap file."""
+    # ensure labels are in consistent order for identity
+    labels = dict(zip(labelnames, labelvalues))
+    return json.dumps([metric_name, name, labels], sort_keys=True)
+
+
 def _MultiProcessValue(_pidFunc=os.getpid):
     files = {}
     values = []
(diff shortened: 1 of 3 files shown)


prometheus_client/multiprocess.py
```diff
@@ -4,6 +4,11 @@
 
 from collections import defaultdict
 
+try:
+    from collections import OrderedDict
+except ImportError:
+    OrderedDict = dict
```
brian-brazil: The code should still work on 2.6

bloodearnest: The code works, but the order is not preserved on 2.6, so it's different behaviour

brian-brazil: Rather than changing this, how about changing the format of the multiprocess key to be a dict for the labels and using sort_keys=True?

prometheus_client/multiprocess.py
```diff
@@ -39,10 +49,10 @@ def collect(self):
                 if typ == 'gauge':
                     pid = parts[2][:-3]
                     metric._multiprocess_mode = parts[1]
-                    metric.add_sample(name, tuple(zip(labelnames, labelvalues)) + (('pid', pid), ), value)
+                    metric.add_sample(name, tuple(labels.items()) + (('pid', pid), ), value)
                 else:
                     # The duplicates and labels are fixed in the next for.
-                    metric.add_sample(name, tuple(zip(labelnames, labelvalues)), value)
+                    metric.add_sample(name, tuple(labels.items()), value)
```
brian-brazil: Is the ordering here deterministic?

prometheus_client/core.py
```diff
@@ -844,6 +851,11 @@ def _floatToGoString(d):
         return '-Inf'
     elif math.isnan(d):
         return 'NaN'
+    elif sys.version_info[:2] == (2, 6):
```
brian-brazil: Why do you need this?

prometheus_client/multiprocess.py
bloodearnest: No, good catch.

prometheus_client/core.py
bloodearnest: It was todo with trying to fix that fact that repr of float in 2.6 is different. Ended up just skipping those tests on 2.6.

brian-brazil:
Can you resolve the conflict? I just merged a large reworking of the library to support openmetrics.

bloodearnest:
Rebased on latest master

brian-brazil:
Thanks!

bloodearnest:
Awesome, thanks :)

xavfernandez:
Thanks @bloodearnest for this feature.

@brian-brazil : any idea when you might release the next version including this ?

brian-brazil:
I'm waiting to get all the openmetrics stuff in, and then release.

PR #303, opened 2018-09-06
Refactor internal sample to use tuple for labels

bloodearnest:
This preserves the order of labels at all times. The only place a dict
of labels was used in the public api is in get_sample_value(), where we
convert to dict for the comparison, which seems reasonable for
a testing helper.

Originally tried just using OrderedDict, for the best of both worlds,
but sadly python 2.6 doesn't have OrderedDict. If dropping 2.6 support is on the table, then just using OrderedDict would probably be a better change, as you'd get the ordering and the convenient dict API

Signed-off-by: Simon Davy <simon.davy@canonical.com>

### DIFF
--- a/prometheus_client/bridge/graphite.py
+++ b/prometheus_client/bridge/graphite.py
@@ -65,7 +65,7 @@ class GraphiteBridge(object):
                     labelstr = '.' + '.'.join(
                         ['{0}.{1}'.format(
                             _sanitize(k), _sanitize(v))
-                            for k, v in sorted(labels.items())])
+                            for k, v in sorted(labels)])
                 else:
                     labelstr = ''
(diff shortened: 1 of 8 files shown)


bloodearnest:
There is a pypi ordereddict module, but I felt that adding a dependency for py2.6 support was a worse option.

brian-brazil:
> The only place a dict of labels was used in the public api is in get_sample_value()

They're used in more places than that. The parser output is all public, as are the samples of the Metric types. So this would be a breaking change for many users.

I'd suggest a more targeted change to do what you need for ordering for multiproc.

bloodearnest:
Would you be OK with a multiproc change that didn't have any effect on py2.6? Because I'm not sure of any other way to do it.

PR #307, opened 2018-09-21
Only open the mmap for reading if read_mode=True

coderanger:
Also fixes a syntax-y error from https://github.com/prometheus/client_python/commit/682e5ecf5c1522bdd7223e49e7e383925a6bbe36

@brian-brazil Pinging as requested :)


brian-brazil:
Thanks!

PR #308, opened 2018-09-23
Add support for tornado.

### LINKED ISSUE #114: Add support for Tornado based exporter
We're considering using this in the [Jupyter](https://jupyter.org/) project, and would love for it to have a Tornado exporter similar to the twisted one.

jmcarp:
AFAICT this `doesn't require anything silly`, so should be acceptable based on conversation in #114.

[Resolves #114]

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,29 @@ reactor.listenTCP(8000, factory)
 reactor.run()
 ```
 
+#### Tornado
+
+To use prometheus with [tornado](https://www.tornadoweb.org/), there is `MetricsHandler` which exposes metrics as a tornado handler.
+
+```python
+from prometheus_client.tornado import MetricsHandler
+
+import tornado.ioloop
+import tornado.web
+
+def make_app():
+    return tornado.web.Application([
+        (r"/", MainHandler),
+    ])
+
+if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


jmcarp:
Yes, but according to https://www.tornadoweb.org/en/stable/wsgi.html#running-wsgi-apps-on-tornado-servers, 

> WSGI is a synchronous interface, while Tornado’s concurrency model is based on single-threaded asynchronous execution. This means that running a WSGI app with Tornado’s WSGIContainer is less scalable than running the same app in a multi-threaded WSGI server like gunicorn or uwsgi. Use WSGIContainer only when there are benefits to combining Tornado and WSGI in the same process that outweigh the reduced scalability.

Which reminds me that I might also need async versions of `generate_latest` and `collect` so that slow collectors don't block the event loop. Which it sounds like don't belong in this library. Will think about this more.

vribeiro1:
Are there any plans on merging this pull request? It's very useful for a project of mine.

brian-brazil:
There's not, but you're free to host it in another repository.

luiscoms:
+1

luiscoms:
It could be added in the documentation as an example

FullDataAlchemist:
+1

PR #312, opened 2018-09-25
OM: Validate utf8 in labelvalues

leecalcote:
Working with @girishranganathan to knock out a TODO for OpenMetrics compliant client.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,3 @@ dist
 .coverage
 .tox
 .*cache
-htmlcov
-.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,6 @@ dist
 .coverage.*
 .coverage
 .tox
+.*cache
+htmlcov
 .vscode
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist
 .coverage.*
 .coverage
 .tox
+.vscode
--- a/prometheus_client/openmetrics/parser.py
+++ b/prometheus_client/openmetrics/parser.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 import math
+import sys
 
 try:
     import StringIO
@@ -113,7 +114,10 @@ def _parse_labels(it, text):
                     raise ValueError("Invalid line: " + text)
                 utf8_str = ''.join(labelvalue)
                 try:
-                    utf8_str.encode('utf-8')
+                    if sys.version_info >= (3,):
+                        utf8_str.encode('utf-8')
+                    else:
+                        utf8_str.decode('utf-8')
                 except UnicodeEncodeError:
                     raise ValueError("Invalid line: " + text)
                 labels[''.join(labelname)] = utf8_str
--- a/prometheus_client/openmetrics/parser.py
+++ b/prometheus_client/openmetrics/parser.py
@@ -1,0 +1,339 @@
+#!/usr/bin/python
+
+from __future__ import unicode_literals
+
+try:
+    import StringIO
+except ImportError:
+    # Python 3
+    import io as StringIO
+
+from .. import core
+
+
+def text_string_to_metric_families(text):
+    """Parse Openmetrics text format from a unicode string.
+
+    See text_fd_to_metric_families.
+    """
+    for metric_family in text_fd_to_metric_families(StringIO.StringIO(text)):
+        yield metric_family
+
+
+def _unescape_help(text):
+    result = []
+    slash = False
+
+    for char in text:
+        if slash:
+            if char == '\\':
+                result.append('\\')
+            elif char == '"':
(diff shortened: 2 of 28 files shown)


prometheus_client/openmetrics/parser.py
```diff
@@ -136,6 +140,8 @@ def _parse_labels(it, text):
                 raise ValueError("Invalid line: " + text)
     return labels
 
+def _validate_utf8(char):
+    return ord(char) < 65536
```
brian-brazil: Why is this a test for a utf-8 character?

I think you were going the right direction with the encode.

leecalcote: @brian-brazil ok. Starting on line 109, how's this instead?

`utf8_str = ''.join(labelvalue)
                try:
                    utf8_str.encode('utf')
                except:
                    raise ValueError("Invalid line: " + text`

brian-brazil: That sounds about right.

prometheus_client/openmetrics/parser.py
```diff
@@ -106,11 +106,16 @@ def _parse_labels(it, text):
             elif char == '"':
                 if not core._METRIC_LABEL_NAME_RE.match(''.join(labelname)):
                     raise ValueError("Invalid line: " + text)
-                labels[''.join(labelname)] = ''.join(labelvalue)
+                utf8_str = ''.join(labelvalue)
+                try:
+                    utf8_str.encode('utf')
```
brian-brazil: I believe it's `utf-8`

brian-brazil:
That looks right, the test you added is failing though.

prometheus_client/openmetrics/parser.py
```diff
@@ -106,11 +106,16 @@ def _parse_labels(it, text):
             elif char == '"':
                 if not core._METRIC_LABEL_NAME_RE.match(''.join(labelname)):
                     raise ValueError("Invalid line: " + text)
-                labels[''.join(labelname)] = ''.join(labelvalue)
+                utf8_str = ''.join(labelvalue)
+                try:
+                    utf8_str.encode('utf-8')
+                except:
```
brian-brazil: Catch the specific exception type that you're looking for

leecalcote: @brian-brazil we're scratching our heads on finding a set of characters as a test case...

leecalcote: Added UnicodeEncodeError exception type.

prometheus_client/openmetrics/parser.py
```diff
@@ -111,11 +112,19 @@ def _parse_labels(it, text):
             elif char == '"':
                 if not core._METRIC_LABEL_NAME_RE.match(''.join(labelname)):
                     raise ValueError("Invalid line: " + text)
-                labels[''.join(labelname)] = ''.join(labelvalue)
+                utf8_str = ''.join(labelvalue)
+                try:
+                    if sys.version_info >= (3,):
+                        utf8_str.encode('utf-8')
+                    else:
+                        utf8_str.decode('utf-8')
```
brian-brazil: This doesn't make sense to me, it should be an encode in both cases.

brian-brazil: Maybe you need a different test for Python 2

.gitignore
```diff
@@ -7,4 +7,5 @@ dist
 .coverage
 .tox
 .*cache
-htmlcov
\ No newline at end of file
+htmlcov
+.vscode
```
brian-brazil: This should go in your global gitignore

tests/openmetrics/test_parser.py
```diff
@@ -247,6 +247,22 @@ def test_labels_with_curly_braces(self):
         metric_family.add_metric(["bar", "b{a}z"], 1)
         self.assertEqual([metric_family], list(families))
 
+    def test_labels_with_invalid_utf8_values(self):
+        try:
+            if sys.version_info < (2, 7):
```
brian-brazil: I think you want to go back to >= 3

Not sure why this is failing on 2.6

tests/openmetrics/test_parser.py
```diff
@@ -252,7 +252,7 @@ def test_labels_with_invalid_utf8_values(self):
             if sys.version_info >= (3,):
                 inj = u'\uD802'
             else:
-                inj = '\xf8\xa1\xa1\xa1\xa1'
+                inj = '\xfc'
```
brian-brazil: \xff should do it

girishranganathan: so using this or any in this way dint help but using them as a binary string seemed to have done the trick

brian-brazil:
Thanks!

leecalcote:
Excellent. :)

PR #313, opened 2018-09-29
Refactoring and metric wrapper unification

### LINKED ISSUE #210: @_MetricWrapper breaks pylint argument checks
`@_MetricWrapper` changes the call signature of the classes it wraps, breaking pylint argument checks. This causes us to have to annotate every single prometheus object with `# pylint: disable-msg=no-value-for-parameter` to avoid erroneous lint warnings. This gets rather silly.

If the mandatory arguments in the call signature of the wrapper and the metrics classes (like Histogram, Counter, etc) matched, pylint would no longer raise this warning.

Code pointers:
Wrapper: https://github.com/prometheus/client_python/blob/d858f0d39536c2fe10b78d6e26109272b6b2ea46/prometheus_client/core.py#L544
Histogram `__init__()`: https://github.com/prometheus/client_python/blob/d858f0d39536c2fe10b78d6e26109272b6b2ea46/prometheus_client/core.py#L847

The linter complains that `labelvalues` isn't being set because it thinks the two positional arguments I pass are `name` and `labelnames`, when in fact I'm passing `name` and `documentation`.

Note that this is not a bug with PyLint, which can only be expected to do static analysis. Because decorators are arbitrary Python code, PyLint would have to execute the decorator in order to find the wrapper arguments.

akx:
This PR does a couple things. (If you want, I can split these into separate PRs.)

🔧  Firstly, it refactors the 1200+-line `core.py` into separate modules for each functionality. This does not affect imports for client code; all the same `prometheus_client`-provided non-underscored names are re-exported from that file. Some previously underscored names were brought into the light.

✨ Secondly (and this is what prompted me to really do this work), it fixes #210 by removing the factory function magic decorator `MetricWrapper`; as it was before, say, an invocation `Gauge(...)` could either return a `Gauge` object, as you'd expect, or a `_LabelWrapper` if you'd passed in a list of label names. This tripped up code completion (and tbqh, me :) ) and static analysis. 
Instead, the six metric wrappers (Counter, Enum, Gauge, Histogram, Info, Summary) are now plain old classes (though with a slightly arcane diamond-shaped inheritance hierarchy that lets them be used both with and without `.labels()`).

👓 Thirdly, there was a mostly mechanical `autopep8` pass to placate `tox flake8` errors. (That check isn't run by CI, it seems?)

No test code was changed for this (aside from the removal of the `__wrapped__` attribute as there is nothing to wrap anymore), so I would say this is backwards API compatible (with the exception of client code that might have done checks like `isinstance(x, _LabelWrapper)`).

I'm glad to answer any questions and/or modify things as needed! :)


brian-brazil:
I'm in the middle of adding various openmetrics-related stuff, so I'll look at this properly once that's all released. One question from a quick peek is why you need two classes to replace label wrapper.

akx:
With this class structure, one can do 

```python
g = Gauge(..., ['label1', 'label2'])
g.labels('v1', 'v2').set(123)
```

but

```python
g = Gauge(..., ['label1', 'label2'])
g.labels('v1', 'v2').labels('v3', 'v4').set(123)
```

is impossible, since `.labels()` returns a `_Gauge`, which is a "leaf" metric object that doesn't have `.labels()`.

If we want to allow `.labels().labels()` (with whatever semantics, either a runtime error or augmentation/replacement of the labels), that'll simplify the class hierarchy.

akx:
Hmm, I'm seeing deadlocks in multiproc mode in the GC collector (which is actually understandable, since GC callbacks could get triggered at any point) with this PR. @brian-brazil, do you think changing `Lock()` to `RLock()` would be safe? I'm not sure what the lock does in general in multiproc.

Haven't yet tested whether `master` has the same problem.

brian-brazil:
I suspect it does have the same problem. @tcolgate

akx:
Hmm, no, 0.4.0 does not have the same problem. I suspect it has to do with how `.labels()` metrics are initialized. Either way, this fork needs to be rebased anyway, so I'll get back to it when I can and just use 0.4.0 fttb.

akx:
Closing as superseded by #344.

PR #318, opened 2018-10-05
Fix import of openmetrics, and make accept handling more general.

brian-brazil:


RichiH:
LGTM

Issue #322, opened 2018-10-09
GCCollector / multiprocess mode deadlock


akx:
Ref #313, #321. 

There's a chance for a deadlock with the GCCollector and multiprocess mode (when `prometheus_multiproc_dir` is set).

For instance, when running our test suite with `py.test`, the process seems to hang, and ctrl+c yields this (on 0.4.0):

```
Traceback (most recent call last):
  File "prometheus_client/gc_collector.py", line 49, in _cb
    latency.labels(gen).observe(delta)
  File "prometheus_client/core.py", line 747, in labels
    self._metrics[labelvalues] = self._wrappedClass(self._name, self._labelnames, labelvalues, **self._kwargs)
  File "prometheus_client/core.py", line 1088, in __init__
    self._sum = _ValueClass(self._type, name, name + '_sum', labelnames, labelvalues)
  File "prometheus_client/core.py", line 636, in __init__
    with lock:
```

This looks like a problem with GC collector callbacks firing while another metric's value object (a `_MultiProcessValue`) is being modified (which holds the `Lock()` shared by all values created [here](https://github.com/prometheus/client_python/blob/9e65e887694137bfa53fef3667fbdde0ba7b0772/prometheus_client/core.py#L623)).

One possible option might be to use an `RLock()` instead of a `Lock()`, but I'm not sure what that might cause.

akx:
@tcolgate See here: https://github.com/prometheus/client_python#multiprocess-mode-gunicorn

The crux of the issue is that in multiprocess mode there's an additional lock that prevents concurrent/reentrant writes of values (since they're backed by a file, not bare Python memory).

Since the GC callback may occur at any time, and Python runs the callback code in the main thread, there's a nontrivial chance that the GC callback is called when something else is updating a metric value, causing a deadlock.

tcolgate:
Why is that a deadlock rather than just a lock contention?
Can multiproc be detected? I'd suggest skipping the GC collector if so.

akx:
> Why is that a deadlock rather than just a lock contention?

Because the GC callbacks [are run as "interrupts"](https://github.com/python/cpython/blob/master/Modules/gcmodule.c#L1153-L1190) in the same thread that happens to make an allocation that causes garbage collection, so you may get this happening in a single thread:

1. metric value is updated
2. acquire lock for metric writing
3. allocation occurs for something within metric writing, causes GC
4. GC callback is called, updates a metric
5. deadlock attempting to reacquire same lock we already hold

[...]

PR #324, opened 2018-10-11
Disable GCCollector in multiprocess mode

akx:
To avoid deadlocks (#322), disable the GC collector in multiprocess mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ dist
 .coverage.*
 .coverage
 .tox
+.*cache
+htmlcov
--- a/prometheus_client/gc_collector.py
+++ b/prometheus_client/gc_collector.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 import gc
+import os
 import time
 
 from . import core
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -6,3 +6,6 @@ dist
 .coverage.*
 .coverage
 .tox
+.*cache
```
brian-brazil: The projects policy is that such things belong in your global gitignore, rather than us ending up with everything in ours.

prometheus_client/gc_collector.py
```diff
@@ -3,13 +3,17 @@
...
 import gc
+import os
 import time
 
 from . import core
 
 class GCCollector(object):
     """Collector for Garbage collection statistics."""
     def __init__(self, registry=core.REGISTRY, gc=gc):
+        if 'prometheus_multiproc_dir' in os.environ:
```
brian-brazil: Can you add a comment with a link to the issue? We can revert this once it's fixed.

brian-brazil:
Can you add the DCO please?

akx:
DCO done.

brian-brazil:
Thanks!

xavfernandez:
A patch release with this PR would be nice :+1: 
Any idea when this could be done ?

PR #325, opened 2018-10-11
Code style fixups

akx:
This PR fixes some wonky indentation (there were occasions of non-4-divisible indentation 😱 ) both mechanically (`autopep8`) and by hand where manually rewrapping code lead to more readable statements. (That, too, could be automated with something like `black`.)

In addition, there were some duplicate imports, which were cleaned up with `isort`. A `setup.cfg` file was added to configure the multi-line import style to match the one most used in the project.

### DIFF
--- a/prometheus_client/bridge/graphite.py
+++ b/prometheus_client/bridge/graphite.py
@@ -4,8 +4,8 @@ from __future__ import unicode_literals
 import logging
 import re
 import socket
-import time
 import threading
+import time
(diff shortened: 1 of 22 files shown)


brian-brazil:
You have a build failure.

akx:
Ah yeah. Old Pythons don't like commas in argument lists. Fixed.

brian-brazil:
Thanks!

PR #326, opened 2018-10-14
Update CONTRIBUTING.md for DCO, test, and lint

bz2:
Mention the DCO signing requirement and link wiki page with details.

Add section on testing requirements and tools.

Add section on code style and lint details.

PR #327, opened 2018-10-14
Enable flake8 import lint checking

bz2:
With #325 by @akx has landed, the flake8 rules on import style lint can be enabled.

Combined the isort `setup.cfg` into `tox.ini` and added a couple of extra config options so it passes the default flake8 rules.

As it wasn't very obvious that there was existing lint tooling #326 updates `CONTRIBUTING.md` to point it out.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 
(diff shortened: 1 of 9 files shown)


brian-brazil:
Thanks!

PR #328, opened 2018-10-17
In multiprocess mode, ensure that metrics initialise to the correct file

bloodearnest:
Without this, when forking, children can initialise metrics to the
cached master pid file. This is racy, and can corrupt the master's
mmaped file, breaking all subsequent collections.


brian-brazil:
Thanks!

PR #330, opened 2018-10-23
Add namespace support to platform

dannysauer:
There was no namespace support in the platform collector.  That seemed weird, so here it is - complete with extra tests. :)

### DIFF
--- a/prometheus_client/platform_collector.py
+++ b/prometheus_client/platform_collector.py
@@ -10,14 +10,16 @@ from . import core
 class PlatformCollector(object):
     """Collector for python platform information"""
 
-    def __init__(self, registry=core.REGISTRY, platform=None):
+    def __init__(self, namespace='', registry=core.REGISTRY, platform=None):
+        self._namespace = namespace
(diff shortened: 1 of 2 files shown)


brian-brazil:
Why do you want this?

dannysauer:
I prefer for all metrics exported by my application to fall under a consistent namespace, per the Prometheus metric naming guidelines: https://prometheus.io/docs/practices/naming/

brian-brazil:
That's not what the naming guidelines say, the namespace here is `python`.

dannysauer:
I would've called python the subsystem, while my application would be the namespace. I'm willing to accept the possibility that I've been doing it wrong all along, though. :D

brian-brazil:
In order to differentiate different applications you should rely on target labels, not metric names. It'd be quite rare that all metrics of a given application would all share a prefix.

dannysauer:
Good enough; if there's insufficient perceived value in adding symmetry to this collector, feel free to close without implementing. From my end, since I hadn't used `tox` previously, I got enough benefit out of the process of simply writing the code. :)

PR #331, opened 2018-10-23
flake8 compliance

dannysauer:
The contribution guide says that new code should be flake8 compliant, so this fixes ordering in a couple of import statements which were previously causing that test to fail. :)

I didn't add `- env: TOXENV=flake8` to the `travis.yml` file, though maybe that wouldn't be a terrible idea.

@brian-brazil


brian-brazil:
Thanks!

Issue #334, opened 2018-10-29
No changelog


adamchainz:
I'm upgrading from 0.4.1 to 0.4.2, it would nice to be able to see what changed. I've hunted around and can't find any changelog/release notes. If you want some guidance on keeping a changelog, there's https://keepachangelog.com/en/1.0.0/

brian-brazil:
It's at https://github.com/prometheus/client_python/releases

PR #335, opened 2018-10-29
Link to GitHub release notes from README

adamchainz:
As per #334 I didn't find the release notes, as using GitHub releases is pretty uncommon in Python land. A link would be appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 The official Python 2 and 3 client for [Prometheus](http://prometheus.io).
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Our releases notes are in the standard location for Github, I feel adding a 2nd place that says "Releases" will add more clutter than anything.

adamchainz:
It's pretty uncommon for python packages to use the release notes so that's why I didn't find them. Please help make them discoverable, it's better to have too many links than too few.

PR #338, opened 2018-11-05
Check for negative counter-like and guage histogram values.

brian-brazil:
@SuperQ per the last two meetings.


prometheus_client/openmetrics/parser.py
```diff
@@ -112,11 +113,16 @@ def _parse_labels(it, text):
...
+                try:
+                    utf8_str.encode('utf-8')
+                except UnicodeError:
+                    raise ValueError("Invalid line: " + text)
+                labels[''.join(labelname)] = utf8_str
                 labelname = []
                 labelvalue = []
                 state = 'endoflabelvalue'
-            else:
+            else:
```
SuperQ: Trailing spaces?

brian-brazil: Fixed

PR #339, opened 2018-11-07
Permit subclassing of MetricsHandler

jeanralphaviles:
@brian-brazil 

Changes MetricsHandler.factory from a staticmethod to a classmethod to be able to determine the correct subclass to return. This way if a user wants to extend the behavior of their MetricsHandler, they don't have to reimplement the factory() method.


brian-brazil:
Thanks!

PR #343, opened 2018-11-13
Avoid re-entrant calls to GC collector's callback

akx:
It looks like the Python GC engine can call the GC callback
while it's still being called (likely if the callback does significant
enough work), which may, if you're unlucky enough, lead into a stack
overflow Python is unable to recover from.

This patch should fix it, though no empirical tests have been made.

Refs this issue comment: https://github.com/prometheus/client_python/issues/322#issuecomment-438021132

### DIFF
--- a/prometheus_client/gc_collector.py
+++ b/prometheus_client/gc_collector.py
@@ -47,19 +47,32 @@ class GCCollector(object):
 
         times = {}
 
+        # Avoid _cb() being called re-entrantly
+        # by setting this flag and clearing it once
+        # the callback operation is complete.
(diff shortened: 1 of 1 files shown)


brian-brazil:
We should probably also try to avoid allocating memory in this function, such as by caching the children. That'll reduce the amount of data we lose due to this.

akx:
Is that worth the added complexity? Either way, maybe another issue should be made for that, and this merged so people don't have their Pythons fatally crashing...

brian-brazil:
> Is that worth the added complexity? 

It'd avoid losing data on all bar the first GC, and be a little faster too.

PR #344, opened 2018-11-13
Refactoring and metric wrapper unification (mark 2)

### LINKED ISSUE #210: @_MetricWrapper breaks pylint argument checks

akx:
This is a rebase of #313 (not my favorite rebase btw; that's why the N refactoring commits in #313 become one big one here) – the comments below are mostly the same.

---

🔧  This PR refactors the 1200+-line `core.py` into separate modules for each functionality. This does not affect imports for client code; all the same `prometheus_client`-provided non-underscored names are re-exported from that file. Some previously underscored names were brought into the light.

✨ Secondly (and this is what prompted me to really do this work), it fixes #210 by removing the factory function magic decorator `MetricWrapper`; as it was before, say, an invocation `Gauge(...)` could either return a `Gauge` object, as you'd expect, or a `_LabelWrapper` if you'd passed in a list of label names. This tripped up code completion (and tbqh, me :) ) and static analysis. 
Instead, the six metric wrappers (Counter, Enum, Gauge, Histogram, Info, Summary) are now plain old classes (though with a slightly arcane diamond-shaped inheritance hierarchy that lets them be used both with and without `.labels()`).

👓  No significant test code was changed for this (aside from the removal of the `__wrapped__` attribute as there is nothing to wrap anymore), so I would say this is backwards API compatible (with the exception of client code that might have done checks like `isinstance(x, _LabelWrapper)`).

---

I'm glad to answer any questions and/or modify things as needed! :)


prometheus_client/metric_wrappers.py
```diff
@@ -16,21 +16,105 @@
...
+
+        This is factored as a separate function to allow for deferred initialization.
+        """
+        raise NotImplementedError('_metric_init() must be implemented by %r' % self)
+
+    def _samples(self):  # pragma: no cover
+        raise NotImplementedError('_samples() must be implemented by %r' % self)
+
+
+def build_full_name(metric_type, name, namespace, subsystem, unit):
```
brian-brazil: internal functions and classes should begin with _

prometheus_client/metric_wrappers.py
```diff
@@ -16,21 +16,105 @@
...
+            raise ValueError('Invalid metric name: ' + self._name)
+
+        if registry:
+            registry.register(self)
+
+        if self._is_multi_wrapper:  # This is a multi-wrapper, so it needs additional initialization
+            self._lock = Lock()
+            self._metrics = {}
+            for l in self._labelnames:
+                if l.startswith('__'):
```
brian-brazil: Why not do this in validate_labelnames?

prometheus_client/metric_wrappers.py
```diff
@@ -16,21 +16,105 @@
...
+        self._labelnames = tuple(labelnames)
+        self._labelvalues = tuple(labelvalues)
         self._kwargs = kwargs
-        self._lock = Lock()
-        self._metrics = {}
 
-        for l in labelnames:
-            if l.startswith('__'):
-                raise ValueError('Invalid label metric name: ' + l)
+    @property
```
brian-brazil: Why not leave this as a function?

prometheus_client/__init__.py
```diff
@@ -5,18 +5,21 @@
...
-Counter = core.Counter
-Gauge = core.Gauge
-Summary = core.Summary
-Histogram = core.Histogram
-Info = core.Info
-Enum = core.Enum
+CollectorRegistry = registry.CollectorRegistry
+REGISTRY = registry.REGISTRY
+Metric = metrics.Metric
+Counter = metric_wrappers.Counter
```
brian-brazil: Users will look for metrics, not metric_wrappers - this naming can be improved.

prometheus_client/values.py
```diff
@@ -0,0 +1,108 @@
...
+        with self._lock:
+            self._value = value
+
+    def get(self):
+        with self._lock:
+            return self._value
+
+
+def MultiProcessValue(_pidFunc=os.getpid):
+    from .mmap_dict import MmapedDict, mmap_key
```
brian-brazil: Imports should be at the top level, this way has been associated with deadlocks

nazarewk:
Was about to start a discussion on the topic, then I fount this PR. Absolutely love the initiative and hope this gets merged quickly.

Ease of use is already amazing, but lack of type hints due to decorator functions magic is a nightmare.

prometheus_client/__init__.py
nazarewk: maybe rename `metrics` -> `metrics_core` and `metrics_wrappers` -> `metrics` ?

akx: That sounds better. Would you agree with that naming, @brian-brazil?

prometheus_client/metric_wrappers.py
```diff
@@ -16,21 +16,105 @@
...
+    def _is_multi_wrapper(self):
+        return (self._labelnames and not self._labelvalues)
+
+    def describe(self):
+        return [Metric(self._name, self._documentation, self._type)]
```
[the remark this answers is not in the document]
akx: It isn't in the original code either, so I didn't add it here. If the omission in the original is a bug, then sure, I'll fix that too. :)

https://github.com/prometheus/client_python/blob/a50f3e4ba38617e9c64c928087c7548b09081975/prometheus_client/core.py#L846-L847

prometheus_client/metric_wrappers.py
akx: Good question. Will fix.

prometheus_client/metric_wrappers.py
akx: Sure, if that's preferred.

prometheus_client/values.py
akx: The idea here was that the whole mmap functionality wouldn't even be imported unless a MultiProcessValue is required, but I don't mind changing it.

Curious though -- what sort of deadlocks do you mean? In what situations would those maybe happen? 😲

akx:
@brian-brazil: 

> I'm still not seeing why you need the complex class hierarchy, we only need to support labels() once.

The more complex class hierarchy makes it so that the child metrics don't actually even _have_ a `labels()` function one might be able to accidentally call.

I can simplify that and raise a runtime error within `.labels()` if one tries to do `x.labels().labels()`. Would that be better?

prometheus_client/metric_wrappers.py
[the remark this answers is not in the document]
brian-brazil: That is a bug, though it'd have no effect currently. It can be fixed in another PR - this is a complex enough one as-is.

prometheus_client/__init__.py
brian-brazil: Sure

akx:
Okay, review comments addressed and the re-refactor pushed as new commits.

I can squash/rebase as needed, but figured this might make reviewing easier.

brian-brazil:
That looks okay. Can you squash sanely?

akx:
@brian-brazil Done deal. Squashed the last 3 commits.

brian-brazil:
Thanks!

PR #346, opened 2018-11-15
Float histograms

### LINKED ISSUE #345: Histogram buckets do not follow OpenMetrics specification
OpenMetrics draft specifies the default Go %g format. 

Python is adding decimal number to the `le` field.

Like this:
```
request_duration_seconds_bucket{code="200",endpoint="/batch_read",le="1.0",method="GET",type="REST"} 545729.0
```

the `le` should be `le="1"`

This causes problems when comparing metrics with i.e. golang exported metrics

More about this https://github.com/prometheus/client_golang/issues/500

brian-brazil:
Fixes #345

### DIFF
--- a/prometheus_client/openmetrics/parser.py
+++ b/prometheus_client/openmetrics/parser.py
@@ -6,6 +6,7 @@ import math
 
 from ..metrics_core import Metric, METRIC_LABEL_NAME_RE
 from ..samples import Exemplar, Sample, Timestamp
+from ..utils import floatToGoString
 
 try:
     import StringIO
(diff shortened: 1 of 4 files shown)


tivvit:
wow cool thanks for the speed!

tivvit:
I think that it looks ok, but I do not know the codebase much and I do not have the rights to merge it. Who would be the best person to merge this @brian-brazil?

brian-brazil:
There's more work to be done here, it looks like we're going the other way on this.

brian-brazil:
This is updated now.

PR #348, opened 2018-11-16
Allow time-based metrics to conditionally include or exclude results

cpmclouth:
It is desirable to control whether metrics are captured when exceptions occur within wrapped code. @brian-brazil 

This PR allows time-based metrics to conditionally include or exclude results when exceptions occur in wrapped code 

added keyword argument observe_exceptions to Gauge.time, Summary.time,
Histogram.time, and _Timer constructor.

    observe_exceptions: None, an Exception or tuple of Exceptions, or an empty tuple
        None: always observe
        Exception or tuple of Exceptions: observe if no exception raised;
            observe if exception raised is provided or in tuple.
        empty tuple: only observe if no exception raised.

### DIFF
--- a/prometheus_client/core.py
+++ b/prometheus_client/core.py
@@ -1003,12 +1003,19 @@ class Gauge(object):
         '''
         return _InprogressTracker(self)
 
-    def time(self):
+    def time(self, observe_exceptions=None):
         '''Time a block of code or function, and set the duration in seconds.
+
+        Args:
+          observe_exceptions: None, an Exception or tuple of Exceptions, or an empty tuple
+            None: always observe
+            Exception or tuple of Exceptions: observe if no exception raised;
+                observe if exception raised is provided or in tuple.
+            empty tuple: only observe if no exception raised.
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thank you for your PR. This is considered an anti-pattern in instrumentation, as it leads to graphs of only successful latencies being used and thus missing that failing queries are slowing things down.

cpmclouth:
Thank you, for your consideration. In our case, we were not interested in recording timings of unexpected failures because we are tracking those events independently. The failures would skew the results that are important to us.

I could make the argument that the decision be left up to the library consumer, but I don't feel strongly enough about it to mount a serious challenge. 

Thanks for the library, it has proven quite useful.

brian-brazil:
There's nothing stopping you from implementing this in your own code, but I don't think it makes sense to have it in the core library.

cpmclouth:
Yes, we already did just that. As I said before, I don't have a strong opinion about it. close it as you see fit.

PR #351, opened 2018-11-26
Make prometheus_multiproc_dir uppercase to support running under Docker

mlaitinen:
Currently it's not possible to define the `prometheus_multiproc_dir` when the application is running inside a Docker container because you can't define lowercase environment variables for a container.

This fix makes the environment variable uppercase while still supporting the old lowercase `prometheus_multiproc_dir` environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -478,7 +478,7 @@ There's several steps to getting this working:
 
 **One**: Gunicorn deployment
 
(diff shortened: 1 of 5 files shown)


mlaitinen:
Nope. I was wrong.

PR #354, opened 2018-12-06
Make it possible to specify a directory path

slater-brian-john:
I noticed whilst trying to use MultiprocessCollector that I could not set a the path variable.


brian-brazil:
I'm not seeing how this changes how the code works.

brian-brazil:
The logic is still the same, so if you got an exception before you'd get one now.

slater-brian-john:
No idea what I was thinking, sorry to waste your time :)

PR #355, opened 2018-12-10
add: default arg to write_to_textfile

sugitak:
### What

Add `REGISTRY` as the default arg of `write_to_textfile`

### Why

`write_to_textfile` is one of the easiest way to bring the power of custom metrics to the users' Prometheus, so it should be very nice to have its interface simple.
Other functions like `make_wsgi_app`, `start_http_server` already have it as default value, so it seems fair to add `REGISTRY` here as default.

#### before

```
from prometheus_client import write_to_textfile, REGISTRY

while True:
    do_whatever()
    write_to_textfile(TMP_PATH, REGISTRY)
```

#### after

```
from prometheus_client import write_to_textfile

while True:
    do_whatever()
    write_to_textfile(TMP_PATH)
```

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -184,7 +184,7 @@ def start_http_server(port, addr='', registry=REGISTRY):
     t.start()
 
 
-def write_to_textfile(path, registry):
+def write_to_textfile(path, registry=REGISTRY):
     '''Write metrics to the given path.
 
     This is intended for use with the Node exporter textfile collector.


brian-brazil:
Your example is not a valid use for the textfile collector, it's something to be written once at the end of a batch job and always with a custom rather than the default registry.

sugitak:
> If you find otherwise, you should probably be scraping the process normally.

I think I got what you've meant. So having everything as HTTP server is the Google way and the philosophy is the same for Prometheus. Makes sense.

But, well, I want to keep the number of the exposed ports/servers as small as possible.  So I prefer doing this way, especially when I run this for internal network monitoring using ssh and expect.
(also, cron always makes a mess, after a few years later)

Anyway, that's just MY way and not the `prometheus/client_python` way. I deeply agree with that.
So thank you, I close this PR.

PR #361, opened 2019-01-03
Make the format strings compatible with Python 2.6

vgrigoriu:
This should restore compatibility with Python 2.6, current version fails with:
```
...
  File "/usr/lib/python2.6/site-packages/prometheus_client-0.5.0-py2.6.egg/prometheus_client/exposition.py", line 194, in write_to_textfile
    f.write(generate_latest(registry))
  File "/usr/lib/python2.6/site-packages/prometheus_client-0.5.0-py2.6.egg/prometheus_client/exposition.py", line 119, in generate_latest
    output.append(sample_line(s))
  File "/usr/lib/python2.6/site-packages/prometheus_client-0.5.0-py2.6.egg/prometheus_client/exposition.py", line 86, in sample_line
    s.name, labelstr, floatToGoString(s.value), timestamp)
  File "/usr/lib/python2.6/site-packages/prometheus_client-0.5.0-py2.6.egg/prometheus_client/utils.py", line 21, in floatToGoString
    mantissa = '{}.{}{}'.format(s[0], s[1:dot], s[dot+1:]).rstrip('0.')
ValueError: zero length field name in format
```

### DIFF
--- a/prometheus_client/utils.py
+++ b/prometheus_client/utils.py
@@ -18,6 +18,6 @@ def floatToGoString(d):
         # Go switches to exponents sooner than Python.
         # We only need to care about positive values for le/quantile.
         if d > 0 and dot > 6:
-            mantissa = '{0}.{1}{2}'.format(s[0], s[1:dot], s[dot+1:16]).rstrip('0.')
+            mantissa = '{0}.{1}{2}'.format(s[0], s[1:dot], s[dot+1:]).rstrip('0.')
             return '{0}e+0{1}'.format(mantissa, dot-1)
         return s
--- a/prometheus_client/utils.py
+++ b/prometheus_client/utils.py
@@ -18,6 +18,6 @@ def floatToGoString(d):
         # Go switches to exponents sooner than Python.
         # We only need to care about positive values for le/quantile.
         if d > 0 and dot > 6:
-            mantissa = '{}.{}{}'.format(s[0], s[1:dot], s[dot+1:]).rstrip('0.')
-            return '{}e+0{}'.format(mantissa, dot-1)
+            mantissa = '{0}.{1}{2}'.format(s[0], s[1:dot], s[dot+1:]).rstrip('0.')
+            return '{0}e+0{1}'.format(mantissa, dot-1)
         return s
--- a/prometheus_client/utils.py
+++ b/prometheus_client/utils.py
@@ -18,6 +18,6 @@ def floatToGoString(d):
         # Go switches to exponents sooner than Python.
         # We only need to care about positive values for le/quantile.
         if d > 0 and dot > 6:
-            mantissa = '{0}.{1}{2}'.format(s[0], s[1:dot], s[dot+1:]).rstrip('0.')
+            mantissa = '{0}.{1}{2}'.format(s[0], s[1:dot], s[dot+1:16]).rstrip('0.')
             return '{0}e+0{1}'.format(mantissa, dot-1)
         return s
--- a/prometheus_client/samples.py
+++ b/prometheus_client/samples.py
@@ -6,7 +6,7 @@ class Timestamp(object):
 
     def __init__(self, sec, nsec):
         if nsec < 0 or nsec >= 1e9:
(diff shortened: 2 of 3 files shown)


brian-brazil:
Can you add the DCO?

vgrigoriu:
Done.

brian-brazil:
The tests are failing on 2.6, are you sure this is broken?

vgrigoriu:
Looking into the tests failing.

vgrigoriu:
So, I think I figured things out. The tests fail in Python 2.6 because of [this line](https://github.com/prometheus/client_python/blob/5aa256d8aab3b81604b855dc03f260342fc391fb/tests/openmetrics/test_parser.py#L571).

The test tests that all those values generate `ValueError`s. This passes in Python 2.7 because `floatToGoString('1.5555555555555201e+06')` is `'1.55555555555552e+06'` -- see [this line](https://github.com/prometheus/client_python/blob/5aa256d8aab3b81604b855dc03f260342fc391fb/prometheus_client/openmetrics/parser.py#L407).

In Python 2.6, without my changes, the test passes because [this line](https://github.com/prometheus/client_python/blob/5aa256d8aab3b81604b855dc03f260342fc391fb/prometheus_client/utils.py#L21), coincidentally, throws `ValueError` (because 2.6 requires numbers for placeholders).

With my changes, this line no longer throws `ValueError` and it happens that `floatToGoString(1.5555555555555201e+06)` is `'1.5555555555555201e+06'` (the same), so the expected `ValueError` is not thrown.

The difference in the result of `floatToGoString` is that in 2.6 `repr(1.5555555555555201e+06)` is `'1555555.5555555201'` and in 2.7 it is `'1555555.55555552'`.

Obviously I don't know why this specific value was chosen for this test, but maybe we should remove it? What do you think @brian-brazil?

brian-brazil:
Odd that CI let this though.

> Obviously I don't know why this specific value was chosen for this test, but maybe we should remove it? 

It's there specifically to catch this. This sounds like https://bugs.python.org/issue7117 or https://bugs.python.org/issue7632

vgrigoriu:
> Odd that CI let this though.

This passed in CI for 2.6 because the only test path exercising [this code](https://github.com/prometheus/client_python/blob/5aa256d8aab3b81604b855dc03f260342fc391fb/prometheus_client/utils.py#L21) happens to be in [a test](https://github.com/prometheus/client_python/blob/5aa256d8aab3b81604b855dc03f260342fc391fb/tests/openmetrics/test_parser.py#L590-L591) that expects `ValueError`.

I pushed a change that makes `floatToGoString` work the same for the existing test cases in 2.6 and 2.7.

prometheus_client/utils.py
```diff
@@ -18,6 +18,6 @@ def floatToGoString(d):
         # Go switches to exponents sooner than Python.
         # We only need to care about positive values for le/quantile.
         if d > 0 and dot > 6:
-            mantissa = '{0}.{1}{2}'.format(s[0], s[1:dot], s[dot+1:]).rstrip('0.')
+            mantissa = '{0}.{1}{2}'.format(s[0], s[1:dot], s[dot+1:16]).rstrip('0.')
```
brian-brazil: I'm not sure this is correct, as it's the underlying algorithm that's not right.

vgrigoriu: Are you saying that there's no easy way to preserve compatibility with Python 2.6?

brian-brazil: Possibly not, we'd need our own floating point renderer for it.

May I ask why you're still on 2.6?

vgrigoriu: For my sins :-P

I'll close the PR then, I'm good with using 0.4.2 until we can upgrade to Python 2.7.

brian-brazil:
We should still fix the format bug on 2.6

vgrigoriu:
So what do you want to do about the test failing on 2.6?

bz2:
Can skip the test on 2.6 only with a message referencing the floating point formatting compatibility issue.

brian-brazil:
Thanks!

PR #364, opened 2019-01-08
Added metrics to MetricFamily exceptions to fix #362

### LINKED ISSUE #362: No input validation on MetricFamily classes resulting in confusing errors
Using the code from the readme with the small alteration of an incorrect value, an export results in a useless stacktrace.

The code used:
```python
from prometheus_client.core import GaugeMetricFamily, CounterMetricFamily, REGISTRY

class CustomCollector(object):
    def collect(self):
        yield GaugeMetricFamily('my_gauge', 'Help text', value=7)
        c = CounterMetricFamily('my_counter_total', 'Help text', labels=['foo'])
        c.add_metric(['bar'], None)
        c.add_metric(['baz'], 'baz')
        yield c

REGISTRY.register(CustomCollector())
```

The resulting trace:
```python
Exception happened during processing of request from ('127.0.0.1', 12345)
  File "/usr/lib/python3.6/socketserver.py", line 651, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.6/socketserver.py", line 361, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.6/socketserver.py", line 721, in __init__
    self.handle()
  File "/usr/lib/python3.6/http/server.py", line 418, in handle
    self.handle_one_request()
  File "/usr/lib/python3.6/http/server.py", line 406, in handle_one_request
    method()
  File "prometheus_client/exposition.py", line 146, in do_GET
    output = encoder(registry)
  File "prometheus_client/exposition.py", line 119, in generate_latest
    output.append(sample_line(s))
  File "/opt/lg5commander/.venv/lib/python3.6/site-packages/prometheus_client/exposition.py", line 86, in sample_line
    s.name, labelstr, floatToGoString(s.value), timestamp)
  File "test.py", line 8, in floatToGoString
    d = float(d)
TypeError: float() argument must be a string or a number, not 'NoneType'
```

I am not too familiar with the code but I wonder if it would somehow be possible to merge the `metrics_core.py` code with `metrics.py`. Or is there a reason that, for example, a `Gauge` does not use the `GaugeMetricFamily` internally (or vice versa)?

WoLpH:


brian-brazil:
That's a bit more code than I'd expect. We can't use `raise from` as we support 2.6/7, but https://stackoverflow.com/questions/9157210/how-do-i-raise-the-same-exception-with-a-custom-message-in-python/9157277 should work.

WoLpH:
That's indeed possible, but the added advantage of this method is that you can catch `MetricErrorBase` and can be certain it has a `metric` property :)
Additionally, testing is a bit easier like this and it still works if you're simply trying to catch something like `TypeError`.

brian-brazil:
We already know what the metric is, as it's in local scope.

WoLpH:
Only from within the exposition method, which is what I was trying to mitigate here. This allows for easy exception handling outside of it while still maintaining full backwards compatibility.

Can you provide an example of how you would prefer the handling otherwise? The question you linked has 11 answers so that's a little broad :P

brian-brazil:
First answer, 2nd example.

WoLpH:
I've simplified it a bit now, is that better?

tests/test_metrics_core.py
```diff
@@ -0,0 +1,83 @@
+import pytest
```
brian-brazil: Shouldn't this be in test_exposition now?

WoLpH: We could move it but it's specific to the classes from `metrics_core.py` so I'm not sure if that would be better

brian-brazil: It's testing the exposition code though.

PR #366, opened 2019-01-18
Fix SyntaxWarning due to unescaped sequence

tirkarthi:
This seemed to be trivial one so I didn't open an issue. This fixes a `SyntaxWarning` introduced on CPython master that will be released as 3.8. It's already a `DeprecationWarning` on Python 3.6. The escape sequence needs to be present inside a raw string to fix the warning.

```
prometheus_client/decorator.py:82
  /home/karthi/client_python/prometheus_client/decorator.py:82: SyntaxWarning: invalid escape sequence \s
    DEF = re.compile('\s*def\s*([_\w][_\w\d]*)\s*\(')
```

Thanks for the library :)

### DIFF
--- a/prometheus_client/decorator.py
+++ b/prometheus_client/decorator.py
@@ -79,7 +79,7 @@ def getargspec(f):
     return ArgSpec(spec.args, spec.varargs, spec.varkw, spec.defaults)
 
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

Issue #367, opened 2019-01-21
Multiprocess mode is slow


akx:
We're using uwsgi to run in production with `workers > 1`, and as such, we need to use the client's multiprocess mode.

The "this thing doesn't work in multiprocess mode" features aside, it works acceptably in the beginning. However, presenting metrics gets slower and slower as the container we run in lives on, eventually leading to the Prometheus server timing out scraping metrics. As a stopgap measure, I've just increased the timeout on the server's end, but that's not the best of solutions.

Namely, by the time I have ~6800 .db files (1 MiB each, so roughly 7 gigs altogether) in the multiproc dir, generating 341,361 bytes of metrics takes 40.75 seconds on my dev machine (script below). Yikes! Running under a profiler gives roughly the following, sorted by OwnTime:

Name | Call# | Time (ms) | OwnTime (ms)
-- | -- | -- | --
prometheus_client.mmap_dict.MmapedDict#__init__ | 6802 | 25082 | 24570
prometheus_client.multiprocess.MultiProcessCollector#merge | 1 | 44712 | 4107
_io.BufferedReader#close  | 6802 | 3038 | 3038
prometheus_client.mmap_dict.MmapedDict#_read_all_values | 673199 | 3255 | 1824
json.raw_decode | 666397 | 1848 | 1802
json.decode | 666397 | 4145 | 1055
re.Pattern.match  | 1339663 | 978 | 978
json.loads | 666397 | 5298 | 802
sorted  | 669564 | 729 | 658
prometheus_client.metrics_core.Metric#add_sample | 666397 | 1277 | 630
_struct.unpack_from  | 666397 | 504 | 504

```python
import os
import time

os.environ['prometheus_multiproc_dir'] = '/Users/akx/Desktop/roi_prometheus_multiproc'

import prometheus_client
from prometheus_client import multiprocess

t0 = time.time()
registry = prometheus_client.CollectorRegistry()
multiprocess.MultiProcessCollector(registry)
metrics_page = prometheus_client.generate_latest(registry)
print(len(metrics_page))
print(time.time() - t0)
```

Culling some of the files out makes the script linearly faster (4200 files = 25.72 seconds, 2400 files = 12.5 seconds).

I find it hard to believe other users of the library haven't bumped into this somehow, so I'm wondering if anyone has any mitigation ideas? Otherwise I'm tempted to try my hand at writing a different multiprocess backend, such as SQLite, that might be a little slower at writing but likely much faster at collating the data.

[...]

PR #368, opened 2019-01-21
Proposal: SQLite multiprocess backend

akx:
Following up on #367, I decided to try out whether it'd be feasible to add an SQLite backend for multiprocess values.

It's generally a little slower writing values, but a whole lot faster reading them out, especially when there are many PIDs! (Full disclosure, though: I'm only emulating multiple processes by hacking `os.getpid()` in my [test bench](https://gist.github.com/akx/d74f31a08382664aa43ba602ebb39b92), so these numbers could be off in a real-life situation.) See below for results.

⚠️  This hasn't yet been tested in a production situation at all, nor profiled – I'm sure there are some suboptimal bits and pieces in there. I felt like it would be better to post a work-in-progress PR anyway.
(Come to think of it, another avenue to investigate, if SQLite is considered too heavy-handed, would be some sort of append-only structure of files that can be easily collated and merged.)

prometheus_sqlite


akx:
Re the CI failures: Grump, worked on my machine... :)

brian-brazil:
> if SQLite is considered too heavy-handed, would be some sort of append-only structure of files that can be easily collated and merged.

That'd be even slower. I'd like to keep to no dependencies for this, as that makes it easier to adopt the library. Any tricks that sqllite is pulling, we can potentially pull too.

I'm suspicious of these results, I'd expect a lot more overhead from SQLite and the GIL is probably avoiding contention here.

brian-brazil:
This has gone stale, and doesn't make sense to me.

Issue #363, opened 2019-01-04
Garbage collector metrics lead to deadlock


megabotan:
Story:
We have default gc metrics, for example python_gc_collected_objects_count.
Someone tries to get metrics value(for example http handler). This leads to MutexValue.get() call. Lock is aquired inside get() method.
If garbage collector decides to run at this moment, the callback updating gc metrics is triggered. Inside this callback metrics is need to be updated, but it cant get the lock which is already taken. Deadlock.

Snippet reproducing this story: (if you try to open localhost:8000 it will fail).
``` python
import gc
import time
from threading import Lock
from prometheus_client import values
from prometheus_client import start_http_server


class MutexValue(object):
    '''A float protected by a mutex.'''

    _multiprocess = False

    def __init__(self, typ, metric_name, name, labelnames, labelvalues, **kwargs):
        self._value = 0.0
        self._lock = Lock()

    def inc(self, amount):
        with self._lock:
            self._value += amount

    def set(self, value):
        with self._lock:
            self._value = value

    def get(self):
        with self._lock:
            gc.collect()    #<------------------------------ Added only this line
            return self._value


values.ValueClass = MutexValue

start_http_server(8000)
while True:
    time.sleep(1)
```
Do we really need lock in get() method?

brian-brazil:
What version are you using? This should already be fixed.

megabotan:
prometheus-client==0.5.0

brian-brazil:
Okay, I see what's happening. As this isn't the first issue in this area, how the GC collector works should probably be reconsidered so that we're not using the client library on the critical path.

mhoyer:
We experienced the same in our application. Is there any workaround or fix in sight?

brian-brazil:
PRs and ideas welcome.

KeNaCo:
Our mitigation of this problem is simply wrapping call with timeout which kill thread. Sufficient enough to keep ThreadPool working..

PR #371, opened 2019-02-10
Deadlock on gcCollector

Krukov:
Hello!
Regading https://github.com/prometheus/client_python/issues/363

We have the same problem with asyncio using (aiohttp, single process, single thread)

As we cannot predict when `gc.collect()` will be executed (with `gc.enable`), it is not possible to use something that has locks inside the `gc.collect` callbacks. So, YES, we neet to reconsidered how the GC collector works.
 
One way it is remove GC Collector  :sunglasses:  . 
Another way in this PR - Collect metrics in simle structures witout locks and represent it in 'collect' method.
Therd way is to use `gc.stats` (since python 3.4) with Gauge (`GaugeMetricFamily`) metrics at `collect`.
Or as @megabotan said, just remove lock in `get` methof of `MutexValue`

### DIFF
--- a/prometheus_client/gc_collector.py
+++ b/prometheus_client/gc_collector.py
@@ -13,10 +13,6 @@ class GCCollector(object):
     """Collector for Garbage collection statistics."""
 
     def __init__(self, registry=REGISTRY):
-        # the GC collector is always disabled in multiprocess mode.
-        if 'prometheus_multiproc_dir' in os.environ:
-            return
-
         if not hasattr(gc, 'get_stats'):
             return
         registry.register(self)
--- a/prometheus_client/gc_collector.py
+++ b/prometheus_client/gc_collector.py
@@ -4,124 +4,48 @@ from __future__ import unicode_literals
 
 import gc
 import os
-import time
-from collections import defaultdict
 
-from .metrics_core import HistogramMetricFamily
+from .metrics_core import GaugeMetricFamily
 from .registry import REGISTRY
-from .utils import INF
 
 
 class GCCollector(object):
     """Collector for Garbage collection statistics."""
 
-    _LATENCY = 'latency'
-    _COLLECTED = 'collected'
-    _UNCOLLECTABLE = 'uncollectable'
-
     def __init__(self, registry=REGISTRY):
-        # To work around the deadlock issue described in
-        # https://github.com/prometheus/client_python/issues/322,
         # the GC collector is always disabled in multiprocess mode.
         if 'prometheus_multiproc_dir' in os.environ:
             return
 
-        if not hasattr(gc, 'callbacks'):
+        if not hasattr(gc, 'get_stats'):
             return
-
-        self._buckets = [{
-            self._LATENCY: defaultdict(lambda: 0.0),
-            self._COLLECTED: defaultdict(lambda: 0),
-            self._UNCOLLECTABLE: defaultdict(lambda: 0),
-        } for _ in [0, 1, 2]]
-
-        self._sums = [{
-            self._LATENCY: 0.0,
-            self._COLLECTED: 0,
-            self._UNCOLLECTABLE: 0,
-        } for _ in [0, 1, 2]]
-
-        times = {}
-
-        # Avoid _cb() being called re-entrantly
-        # by setting this flag and clearing it once
-        # the callback operation is complete.
-        # See https://github.com/prometheus/client_python/issues/322#issuecomment-438021132
-        self.gc_cb_active = False
-
-        def _cb(phase, info):
-            try:
-                if self.gc_cb_active:
-                    return
-                self.gc_cb_active = True
-
-                gen = info['generation']
-
-                if phase == 'start':
-                    times[gen] = time.time()
-
-                if phase == 'stop':
-                    delta = time.time() - times[gen]
-                    self._add_to_bucket(self._LATENCY, gen, delta)
-                    if 'collected' in info and info['collected'] != 0:
-                        self._add_to_bucket(self._COLLECTED, gen, info['collected'])
-                    if 'uncollectable' in info and info['uncollectable'] != 0:
-                        self._add_to_bucket(self._UNCOLLECTABLE, gen, info['uncollectable'])
-            finally:
-                self.gc_cb_active = False
-
-        gc.callbacks.append(_cb)
         registry.register(self)
 
-    def _add_to_bucket(self, bucket_name, gen, value):
-        bucket = self._buckets[gen][bucket_name]
-        self._sums[gen][bucket_name] += value
-
-        for bound in self._get_bounds(gen, bucket_name):
-            if value <= bound:
-                bucket[INF] += 1
-                bucket[bound] += 1
-                break
-
-    @staticmethod
-    def _get_bounds(gen, bucket_name):
-        if bucket_name == GC_COLLECTOR._LATENCY:
-            return .005, .01, .025, .05, .075, .1, .25, .5, .75, 1.0, 2.5, 5.0, 7.5, 10.0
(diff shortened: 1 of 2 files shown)


tests/test_gc_collector.py
```diff
@@ -1,54 +1,91 @@
...
-    def __init__(self):
-        self.callbacks = []
+        self.assertIsNone(self.registry.get_sample_value(
+                             'python_gc_collected_objects_sum',
+                             labels={"generation": "0"}))
 
-    def start_gc(self, info):
-        for cb in self.callbacks:
-            cb('start', info)
+        self.assertIsNone(self.registry.get_sample_value(
```
Krukov: I am going to remove this assert.

prometheus_client/gc_collector.py
```diff
@@ -5,15 +5,21 @@
...
+from .metrics_core import HistogramMetricFamily
 from .registry import REGISTRY
+from .utils import INF
 
 
 class GCCollector(object):
     """Collector for Garbage collection statistics."""
 
-    def __init__(self, registry=REGISTRY, gc=gc):
+    _LATENCY = 'latency'
```
brian-brazil: I assume this is needed to avoid allocs?

prometheus_client/gc_collector.py
```diff
@@ -5,15 +5,21 @@
...
+    _COLLECTED = 'collected'
+    _UNCOLLECTABLE = 'uncollectable'
+
+    def __init__(self, registry=REGISTRY):
         # To work around the deadlock issue described in
```
brian-brazil: Can we remove this now?

prometheus_client/gc_collector.py
```diff
@@ -23,28 +29,17 @@ def __init__(self, registry=REGISTRY, gc=gc):
...
 
-        collected = Histogram(
-            'python_gc_collected_objects',
-            'Objects collected during gc',
-            ['generation'],
-            buckets=[500, 1000, 5000, 10000, 50000],
-            registry=registry
-        )
+        self._buckets = [{
+            self._LATENCY: defaultdict(lambda: 0.0),
```
brian-brazil: Are default dicts threadsafe? Can they do allocs?

prometheus_client/gc_collector.py
```diff
@@ -23,28 +29,17 @@ def __init__(self, registry=REGISTRY, gc=gc):
...
+            self._COLLECTED: defaultdict(lambda: 0),
+            self._UNCOLLECTABLE: defaultdict(lambda: 0),
+        } for _ in [0, 1, 2]]
```
brian-brazil: Could there be more than 3 generations?

prometheus_client/gc_collector.py
```diff
@@ -67,15 +62,66 @@ def _cb(phase, info):
...
+            if value <= bound:
+                bucket[INF] += 1
+                bucket[bound] += 1
+                break
+
+    @staticmethod
+    def _get_bounds(gen, bucket_name):
+        if bucket_name == GC_COLLECTOR._LATENCY:
+            return .005, .01, .025, .05, .075, .1, .25, .5, .75, 1.0, 2.5, 5.0, 7.5, 10.0
+        _max = gc.get_threshold()[gen] * 2
```
brian-brazil: Could this vary over time?

It'd also be best to have the same buckets for each generation, otherwise you can't aggregate them.

prometheus_client/gc_collector.py
```diff
@@ -67,15 +62,66 @@ def _cb(phase, info):
...
 
         gc.callbacks.append(_cb)
+        registry.register(self)
+
+    def _add_to_bucket(self, bucket_name, gen, value):
+        bucket = self._buckets[gen][bucket_name]
+        self._sums[gen][bucket_name] += value
+
+        for bound in self._get_bounds(gen, bucket_name):
+            if value <= bound:
```
brian-brazil: You'd also need to add it to every bucket above this if you're taking this approach.

prometheus_client/gc_collector.py
Krukov: No it is just type error protection .

prometheus_client/gc_collector.py
Krukov: I will think about it. not yet

prometheus_client/gc_collector.py
Krukov: yes defaultdicts threadsafe but we should have locks for incrementing : (

prometheus_client/gc_collector.py
Krukov: No, gc have only 3 generations

prometheus_client/gc_collector.py
Krukov: clear

prometheus_client/gc_collector.py
Krukov: In general it is possible to change threshold.  
As a solution we can take threshold of 0 generation (the biggest one) as max border

prometheus_client/gc_collector.py
Krukov: I forget about it

brian-brazil:
I still think we'd be better off using gc.stats.

Krukov:
@brian-brazil Agree with you. I will prepare PR asap

Krukov:
@brian-brazil I remade PR to using `gc.stats`. Can you review it?

prometheus_client/gc_collector.py
```diff
@@ -4,78 +4,48 @@
...
 
 
 class GCCollector(object):
     """Collector for Garbage collection statistics."""
 
-    def __init__(self, registry=REGISTRY, gc=gc):
-        # To work around the deadlock issue described in
-        # https://github.com/prometheus/client_python/issues/322,
+    def __init__(self, registry=REGISTRY):
         # the GC collector is always disabled in multiprocess mode.
```
brian-brazil: This can be removed now

Krukov: done

prometheus_client/gc_collector.py
```diff
@@ -4,78 +4,44 @@
...
             return
+        registry.register(self)
 
-        if not hasattr(gc, 'callbacks'):
-            return
-
-        collected = Histogram(
+    def collect(self):
+        collected = GaugeMetricFamily(
             'python_gc_collected_objects',
```
brian-brazil: Aren't these all counters?

object_collected and objects_uncollectable would also be more usual.

Krukov: Please explain to me:
''object_collected'' Are you taking about variables names or about metrics names?
Is your question about choosing the type of metrics?

brian-brazil: The metric names.

Also, the types don't seem right.

Krukov: I renamed the metrics names and redid it with counter type

prometheus_client/gc_collector.py
```diff
@@ -3,79 +3,44 @@
...
+        registry.register(self)
 
-        if not hasattr(gc, 'callbacks'):
-            return
-
-        collected = Histogram(
-            'python_gc_collected_objects',
+    def collect(self):
+        collected = CounterMetricFamily(
+            'python_gc_objects_collected',
```
brian-brazil: Counters should end in _total

Krukov: `CounterMetricFamily` will add it autocratically.  You can see it in tests https://github.com/prometheus/client_python/pull/371/files#diff-b8616fecc6de945cdf630d3ad09aebb8R26

brian-brazil: Ah yes, I'd forgotten I'd already added that.

PR #372, opened 2019-02-14
Add TLS client certificate auth handler

michal0A:
Signed-off-by: Michal Altmann <michal@prowler.io>


michal0A:
@brian-brazil

brian-brazil:
I wonder if we'd be better adding the sslcontext to the default handler

brian-brazil:
This has gone stale, let me know if you're still interested in this.

Issue #378, opened 2019-02-20
GCCollector raises an exception in PyPy7.0/Python3.5


zcarlson-signifai:
When attempting to import prometheus_client in the PyPy 7.0 runtime with the Python 3.5 language implementation, the following exception is raised:

```
(prometheus-client) [zcarlson@eleusinia-forge prometheus-client]$ python
Python 3.5.3 (928a4f70d3de7d17449456946154c5da6e600162, Feb 09 2019, 11:50:43)
[PyPy 7.0.0 with GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>> import prometheus_client
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/zcarlson/Code/prometheus-client/env/pypy7.0-python3.5/site-packages/prometheus_client/__init__.py", line 5, in <module>
    from . import gc_collector
  File "/home/zcarlson/Code/prometheus-client/env/pypy7.0-python3.5/site-packages/prometheus_client/gc_collector.py", line 46, in <module>
    GC_COLLECTOR = GCCollector()
  File "/home/zcarlson/Code/prometheus-client/env/pypy7.0-python3.5/site-packages/prometheus_client/gc_collector.py", line 17, in __init__
    registry.register(self)
  File "/home/zcarlson/Code/prometheus-client/env/pypy7.0-python3.5/site-packages/prometheus_client/registry.py", line 24, in register
    names = self._get_names(collector)
  File "/home/zcarlson/Code/prometheus-client/env/pypy7.0-python3.5/site-packages/prometheus_client/registry.py", line 64, in _get_names
    for metric in desc_func():
  File "/home/zcarlson/Code/prometheus-client/env/pypy7.0-python3.5/site-packages/prometheus_client/gc_collector.py", line 37, in collect
    for generation, stat in enumerate(gc.get_stats()):
TypeError: 'GcStats' object is not iterable
```

I believe this is because pypy has a different GC from stock CPython, and thus `gc_stats()` returns different information. 

prometheus_client 0.5.0 seems to work, however, from initial testing...

brian-brazil:
Our CI should have caught that as we test on pypy, guess we also need a new test.

Krukov:
There are no CI jobs with pypy with 3,7 . Only with 2,7
https://travis-ci.org/prometheus/client_python/jobs/495991719

patallen:
[...]

PR #380, opened 2019-02-27
Disable gcCollector for pypy

Krukov:
We use `gc.get_stats` to provide metrics at `gcColector` . Pypy gc have the same method but they are very different. (https://github.com/prometheus/client_python/issues/378 http://doc.pypy.org/en/latest/gc_info.html#gc-get-stats)
I think it is better just to not provide gc metrics with pypy.

### DIFF
--- a/prometheus_client/gc_collector.py
+++ b/prometheus_client/gc_collector.py
@@ -13,7 +13,7 @@ class GCCollector(object):
     """Collector for Garbage collection statistics."""
 
     def __init__(self, registry=REGISTRY):
-        if not hasattr(gc, 'get_stats') or platform.python_implementation() == 'PyPy':
+        if not hasattr(gc, 'get_stats') or platform.python_implementation() != 'CPython':
             return
         registry.register(self)
 
--- a/prometheus_client/gc_collector.py
+++ b/prometheus_client/gc_collector.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 import gc
+import platform
 
 from .metrics_core import CounterMetricFamily
 from .registry import REGISTRY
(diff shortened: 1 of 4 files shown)


prometheus_client/gc_collector.py
```diff
@@ -12,7 +13,7 @@ class GCCollector(object):
     """Collector for Garbage collection statistics."""
 
     def __init__(self, registry=REGISTRY):
-        if not hasattr(gc, 'get_stats'):
+        if not hasattr(gc, 'get_stats') or platform.python_implementation() == 'PyPy':
```
brian-brazil: Wouldn't it be better to limit this to CPython?

Krukov: You right.

brian-brazil:
Thanks!

hynek:
Any chance of a release? This seems to be breaking prometheus-async's test suite on both PyPys:

```
______________________ ERROR collecting tests/test_aio.py ______________________
tests/test_aio.py:24: in <module>
.tox/pypy3/site-packages/prometheus_client/__init__.py:5: in <module>
    from . import gc_collector
.tox/pypy3/site-packages/prometheus_client/gc_collector.py:46: in <module>
    GC_COLLECTOR = GCCollector()
.tox/pypy3/site-packages/prometheus_client/gc_collector.py:17: in __init__
    registry.register(self)
.tox/pypy3/site-packages/prometheus_client/registry.py:24: in register
    names = self._get_names(collector)
.tox/pypy3/site-packages/prometheus_client/registry.py:64: in _get_names
    for metric in desc_func():
.tox/pypy3/site-packages/prometheus_client/gc_collector.py:37: in collect
    for generation, stat in enumerate(gc.get_stats()):
E   TypeError: 'GcStats' object is not iterable
```

brian-brazil:
It's on my todo list. I'm expecting some OpenMetrics draft spec changes next week that I also want to include, so it should happen then.

PR #384, opened 2019-03-15
Fix instrumentation link in README

zhulongcheng:
Maybe the link to `instrumentation best practices` should be

https://prometheus.io/docs/practices/instrumentation/#counter-vs-gauge-summary-vs-histogram


brian-brazil:
Thanks!

PR #386, opened 2019-03-20
debug MetricWrapperBase.remove function

xiaoluwwwxiaolu:
Hello，I am using client_python for developing our exporter，thanks for so convenient tool.

In our project, we don't want some of our metrics to be captured, so we should use MetricWrapperBase's remove function.But when self._metrics is empty or "labelvalues" not in self._metrics, our exporter will stop because of "key error". As a user of the tool,I want some more information not just key error.what's worse,I don't want my exporter to be stoped. It would be better if just some warnings, because we have to be sure that some values of labels are not captured, but we cannot be sure that whether the value-tuples are in self._metrics or not.


brian-brazil:
This sort of printlining does not belong in a library like this. I'd suggest not putting in data you don't want to be there in the first place.

xiaoluwwwxiaolu:
yes,you are right.

PR #387, opened 2019-03-21
start_http_server: returns thread and server

romuald:
This may allow client to join the thread or stop the server manually if needed


brian-brazil:
See #128

PR #390, opened 2019-03-31
Histogram samples should not accumulate buckets

timthesinner:
@brian-brazil it appears that the histogram is accumulating values across all buckets instead of reporting the sampled value for each bucket.


brian-brazil:
This behavior is correct.

PR #393, opened 2019-04-09
Passing arguments to callback function (in set_function)

parsaeian:
@brian-brazil

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -359,7 +359,7 @@ class Gauge(MetricWrapperBase):
         """
         return Timer(self.set)
 
-    def set_function(self, f):
+    def set_function(self, f, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


brian-brazil:
You can use a lambda if you want to do this, there's no need to expand the API.

parsaeian:
> You can use a lambda if you want to do this, there's no need to expand the API.

Suppose this function exists: 
```python
def check_service(service_name):
    report = requests.post(URL, data={ 'service_name': service_name})
    json_data = json.loads(report.text)
    data = json_data['sub_count']
    return data
```
There is a Gauge :
```python
g = Gauge('service_check', 'Check service ', ['service_name'])
```
I need to use like this:
```python
for service in SERVICES:
     g.labels(service_name=service).set_function(check_service, service)
```
How can I use lambda in this case?

PR #395, opened 2019-04-09
Fix bug of multiprocess mode by `path` argument

Sparkycz:
`path` argument at `multiprocess.MultiProcessCollector.__init__()` is only for merging multiprocess files but it doesn't work for processing metrics and creating that files (without set `os.environ.['prometheus_multiproc_dir']`). Everybody has to activate the mode by the environment variable `prometheus_multiproc_dir`.

works: 
```
import os
from prometheus_client import REGISTRY

os.environ.['prometheus_multiproc_dir'] = '/tmp'

MultiProcessCollector(REGISTRY)
```

doesn't work:
```
from prometheus_client import REGISTRY

MultiProcessCollector(REGISTRY, path='/tmp')
```

### DIFF
--- a/prometheus_client/multiprocess.py
+++ b/prometheus_client/multiprocess.py
@@ -13,7 +13,7 @@ from .samples import Sample
 from .utils import floatToGoString
 
 
-PATH = None
+PATH = os.environ.get('prometheus_multiproc_dir')
 
 
 class MultiProcessCollector(object):
@@ -24,8 +24,6 @@ class MultiProcessCollector(object):
 
         if path is not None:
             PATH = path
-        elif 'prometheus_multiproc_dir' in os.environ:
-            PATH = os.environ['prometheus_multiproc_dir']
 
         if not PATH or not os.path.isdir(PATH):
             raise ValueError('Any prometheus multiprocess dictionary is not set or not a directory.')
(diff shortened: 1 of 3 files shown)


brian-brazil:
In case you want to point it at a different directory for some reason. Your change doesn't make sense overall I'm afraid, the path for collectors needs to be configured before the process starts.

Sparkycz:
I'm trying another change.. 

It sets the global variable during the process starts from the environment variable or it's overwritten by the `path` argument. 
The exposing metrics from another dictionary you're still able to because the function `prometheus_generate_latest()` has the argument for injecting the collector you want to expose metrics from.

iScrE4m:
The whole root of this pull request is our misunderstanding of the `path` argument in `MultiprocessCollector`. We assumed that it was an alternative to the environment variable - which I'd guess can be a common mistake for people that skip over the sentence `The prometheus_multiproc_dir environment variable must be set` in README and start experimenting right away instead. My humble suggestion would be to remove the argument, as it gives false 'hope' that there is an alternative...

Then there's the attempt to 'improve' this and make the environmental variable an alternative rather than a requirement. After @Sparkycz sent me the link to this PR we went together through the code and realized that due to the singleton nature of this library it's quite clearly not worth it to implement that and any attempt would just break the current behavior that some might rely on already.

Issue #185, opened 2017-08-21
Including timestamp in metric


jclcntl:
Hello,

I'd like to include the timestamp in the metric according to the Prometheus documentation: https://prometheus.io/docs/instrumenting/exposition_formats/

```
metric_name [
  "{" label_name "=" `"` label_value `"` { "," label_name "=" `"` label_value `"` } [ "," ] "}"
] value [ timestamp ]
```

I'd like that feature for the following reason: Some exporters take a while to gather certain information. Therefore I would like to run them asynchronously, not only when the endpoint is scraped. To guarantee that the metrics are as accurate as possible, I'd like to include the timestamp with the metric.

Currently, I don't think this is possible.

Thanks

brian-brazil:
This isn't currently possible, but a patch to add it would be accepted. #176 is related.

Note that using the timestamp is only recommended when pulling data from another monitoring system that has timestamps, and not when caching.

erikdsjostrom:
Has there been any change in regards to this issue? I'd also like to parse timestamps.

jclcntl:
What's the timeframe for the openmetrics work? I see that on the roadmap it says near future, but can you share more details? Thank you!

brian-brazil:
It'll likely be quite soon. However based on how you have described your use case, you should not use timestamps.

jclcntl:
> Note that using the timestamp is only recommended when pulling data from another monitoring system that has timestamps, and not when caching.

LukeDefeo:
Hi i could use this too. my use case is using https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd to push metrics in prometheus format to stackdriver. Strackdriver needs to know what time range the counters are from / to for them to work

hollodotme:
I have question regarding the timestamp value type. The [prometheus docs](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information) say:

>The timestamp is an int64 (milliseconds since epoch, i.e. 1970-01-01 00:00:00 UTC, excluding leap seconds), represented as required by Go's ParseInt() function.

Looking at the [python parser and its tests](https://github.com/prometheus/client_python/blob/master/tests/openmetrics/test_parser.py#L367) floats (I think microseconds) are used for the timestamps. 

What is the right format / type? Am I missing something?

brian-brazil:
It's seconds for openmetrics.

rabbl:
[...]

PR #396, opened 2019-04-10
Add support for timestamp at MetricFamily creation.

vladvasiliu:
Somewhat related to #185 

This allows to create a MetricFamily with a single metric and a timestamp in one go instead of having to separately call `add_metric`.

### DIFF
--- a/prometheus_client/metrics_core.py
+++ b/prometheus_client/metrics_core.py
@@ -64,7 +64,7 @@ class UnknownMetricFamily(Metric):
     For use by custom collectors.
     """
 
-    def __init__(self, name, documentation, value=None, labels=None, unit=''):
+    def __init__(self, name, documentation, value=None, labels=None, unit='', timestamp=None):
         Metric.__init__(self, name, documentation, 'unknown', unit)
         if labels is not None and value is not None:
             raise ValueError('Can only specify at most one of value and labels.')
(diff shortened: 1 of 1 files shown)


brian-brazil:
Adding a timestamp should be relatively rare, so I think having to do an extra line of code is okay.

vladvasiliu:
Seeing how MetricFamily metrics are mainly used to proxy metrics from other systems I don't think it's *that* rare, and however rare that may be I'm not sure that's a reason for that code to be more verbose, especially since this change doesn't change existing behavior.

brian-brazil:
Even then, custom collectors rarely produce timestamps. There's only a handful of exporters that do it currently.

vladvasiliu:
I understand how that would be a reason to not spend time attempting to implement this, but I'm not sure why the library would actively make it hard for people to use something legitimate, however rare.

brian-brazil:
The number of people who incorrectly think they need this is much higher than the number that actually need it.

vladvasiliu:
I understand that, but the current situation doesn't prevent those people from using it wrongly.
I'm not sure that making it harder for people who actually need it solves the fundamental problem, which is people doing something they should.
As the functionality is already present and is only a function call away, this posture seems a bit moralizing. People are going to use a library in inappropriate ways even if they have to call an extra function. I think making it harder for legitimate uses does nothing more than just that, sacrificing convenience for everyone.

brian-brazil:
My stance remains, I don't see a need to make things easier for use cases that are almost entirely inappropriate.

PR #402, opened 2019-05-06
optimize openmetrics text parsing (~4x perf)

ahmed-mez:
This PR optimizes the openmetrics parser using the logic introduced in https://github.com/prometheus/client_python/pull/282 to optimize the prometheus parser.

Here are some benchmark using `timeit`:

```
call (x100000): _parse_sample('simple_metric 1.513767429e+09')

Simple example with prometheus parser: 0.2489180564880371
Simple example with openmetrics parser: 1.1144659519195557
Simple example with the optimized openmetrics parser: 0.5948491096496582

call (x100000): _parse_sample('kube_service_labels{label_app="kube-state-metrics",label_chart="kube-state-metrics-0.5.0",label_heritage="Tiller",label_release="ungaged-panther",namespace="default",service="ungaged-panther-kube-state-metrics"} 1')

KSM metric example with prometheus parser: 1.6796550750732422
KSM metric example openmetrics parser: 6.6183180809021
KSM metric example optimized openmetrics parser: 2.0289480686187744
```

### DIFF
--- a/prometheus_client/openmetrics/parser.py
+++ b/prometheus_client/openmetrics/parser.py
@@ -175,9 +175,10 @@ def _parse_labels_with_state_machine(text):
 
 def _parse_labels(text):
     labels = {}
-    # Return if we don't have valid labels
-    if "=" not in text:
-        return labels
+
+    # Raise error if we don't have valid labels
+    if text and "=" not in text:
+        raise ValueError
 
     # Copy original labels
     sub_labels = text
@@ -190,14 +191,16 @@ def _parse_labels(text):
             sub_labels = sub_labels[value_start + 1:]
 
             # Check for missing quotes 
-            if sub_labels[0] != '"':
+            if not sub_labels or sub_labels[0] != '"':
                 raise ValueError
 
             # The first quote is guaranteed to be after the equal
             value_substr = sub_labels[1:]
 
             # Check for extra commas
-            if label_name[0] == ',' or value_substr[-1] == ',':
+            if not label_name or label_name[0] == ',':
+                raise ValueError
+            if not value_substr or value_substr[-1] == ',':
                 raise ValueError
 
             # Find the last unescaped quote
@@ -255,7 +258,7 @@ def _parse_sample(text):
         label = text[label_start + 1:label_end]
         labels = _parse_labels(label)
     else:
-        # Line contains an exemplar
+        # Line potentially contains an exemplar
         # Fallback to parsing labels with a state machine
         labels, labels_len = _parse_labels_with_state_machine(text[label_start + 1:])
         label_end = labels_len + len(name)      
--- a/prometheus_client/openmetrics/parser.py
+++ b/prometheus_client/openmetrics/parser.py
@@ -104,7 +104,7 @@ def _parse_timestamp(timestamp):
 
 def _is_character_escaped(s, charpos):
     num_bslashes = 0
-    while (charpos > 0 and
+    while (charpos > num_bslashes and
            s[charpos - 1 - num_bslashes] == '\\'):
         num_bslashes += 1
     return num_bslashes % 2 == 1
@@ -205,7 +205,7 @@ def _parse_labels(text):
             i = 0
             while i < len(value_substr):
                 i = value_substr.index('"', i)
-                if not _is_character_escaped(value_substr, i):
+                if not _is_character_escaped(value_substr[:i], i):
                     break
                 i += 1
 
--- a/prometheus_client/openmetrics/parser.py
+++ b/prometheus_client/openmetrics/parser.py
@@ -3,8 +3,9 @@
 from __future__ import unicode_literals
 
 import math
+import re
 
-from ..metrics_core import Metric, METRIC_LABEL_NAME_RE
+from ..metrics_core import Metric
 from ..samples import Exemplar, Sample, Timestamp
 from ..utils import floatToGoString
 
@@ -22,6 +23,24 @@ def text_string_to_metric_families(text):
     """
     for metric_family in text_fd_to_metric_families(StringIO.StringIO(text)):
         yield metric_family
+
+
+ESCAPE_SEQUENCES = {
+    '\\\\': '\\',
+    '\\n': '\n',
+    '\\"': '"',
+}
+
+
+def replace_escape_sequence(match):
+    return ESCAPE_SEQUENCES[match.group(0)]
+
+
+ESCAPING_RE = re.compile(r'\\[\\n"]')
+
+
+def _replace_escaping(s):
+    return ESCAPING_RE.sub(replace_escape_sequence, s)
 
 
 def _unescape_help(text):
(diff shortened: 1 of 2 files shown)


prometheus_client/openmetrics/parser.py
```diff
@@ -24,6 +25,24 @@ def text_string_to_metric_families(text):
...
 
 
+ESCAPE_SEQUENCES = {
+    '\\\\': '\\',
+    '\\n': '\n',
+    '\\"': '"',
+}
+
+
+def replace_escape_sequence(match):
```
brian-brazil: internal functions and constants should begin with _

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
 
 
-def _parse_labels(it, text):
-    # The { has already been parsed.
-    state = 'startoflabelname'
-    labelname = []
-    labelvalue = []
+def _is_character_escaped(s, charpos):
+    num_bslashes = 0
+    while (charpos > num_bslashes and
```
brian-brazil: Shouldn't this comparison be against 0?

This is also going to be n^2 overall if there's many backslashes.

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
+
+    # Copy original labels
+    sub_labels = text
+    try:
+        # Process one label at a time
+        while sub_labels:
+            # The label name is before the equal
+            value_start = sub_labels.index("=")
+            label_name = sub_labels[:value_start]
+            sub_labels = sub_labels[value_start + 1:].lstrip()
```
brian-brazil: Why are you lstipping here?

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
+
+            # Find the first quote after the equal
+            quote_start = sub_labels.index('"') + 1
+            value_substr = sub_labels[quote_start:]
+
+            # Check for extra commas
+            if label_name[0] == ',' or value_substr[len(value_substr)-1] == ',':
```
brian-brazil: What if label_name is zero length?

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
+                raise ValueError
+
+            # Find the last unescaped quote
+            i = 0
+            while i < len(value_substr):
+                i = value_substr.index('"', i)
+                if not _is_character_escaped(value_substr, i):
```
brian-brazil: Yeah, this is n^2. Work from the start in one loop.

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
+                    break
+                i += 1
+
+            # The label value is inbetween the first and last quote
+            quote_end = i + 1
+            label_value = sub_labels[quote_start:quote_end]
+            # Replace escaping if needed
+            if escaping:
```
brian-brazil: Couldn't you check each value, rather than the whole string?

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
 
 def _parse_sample(text):
-    name = []
-    value = []
+    # Detect the labels in the text
+    try:
+        # `index` and `rindex` methods raise a ValueError with
+        # `substring not found` message if text doesn't contain label braces
+        label_start = text.index("{")
+        label_end = text.rindex("}", 0, text.find(" # "))  # ignore exemplar label braces
```
brian-brazil: What if " # " is part of a label value?

You should also add a testcase for this so it doesn't trip up someone else.

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
+        value, timestamp, exemplar = _parse_remaining_text(remaining_text)
+        return Sample(name, labels, value, timestamp, exemplar)
+
+    except ValueError as e:
+        if str(e).startswith("substring not found"):
+            # We don't have labels
+            # Detect what separator is used
+            separator = " "
+            if separator not in text:
+                separator = "\t"
```
brian-brazil: Tabs aren't supported.

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
+        # We ignore the starting curly brace
+        label = text[label_start + 1:label_end]
+        labels = _parse_labels(label)
+        # Parse the remaining text after the label end
+        remaining_text = text[label_end + 2:]
+        value, timestamp, exemplar = _parse_remaining_text(remaining_text)
+        return Sample(name, labels, value, timestamp, exemplar)
+
+    except ValueError as e:
+        if str(e).startswith("substring not found"):
```
brian-brazil: This is not a clean way to do this. Use find instead.

prometheus_client/openmetrics/parser.py
```diff
@@ -83,95 +102,126 @@ def _parse_timestamp(timestamp):
...
-    name = []
-    value = []
+    # Detect the labels in the text
+    try:
+        # `index` and `rindex` methods raise a ValueError with
+        # `substring not found` message if text doesn't contain label braces
+        label_start = text.index("{")
+        label_end = text.rindex("}", 0, text.find(" # "))  # ignore exemplar label braces
+        # The name is before the labels
+        name = text[:label_start].strip()
```
brian-brazil: Why the strip?

ahmed-mez:
Thank you for reviewing the PR, I've made the requested changes and added a test case, the parsing logic is more solid now.
I'll be looking forward to get a second review. Thanks!

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +170,126 @@ def _parse_labels(it, text):
...
+            # Find the first quote after the equal
+            quote_start = sub_labels.index('"') + 1
+            value_substr = sub_labels[quote_start:]
+
+            # Check for empty label name
+            if len(label_name) == 0:
+                raise ValueError
+
+            # Check for extra commas
+            if label_name[0] == ',' or value_substr[len(value_substr) - 1] == ',':
```
brian-brazil: `value_substr[-1]` is more succinct

prometheus_client/openmetrics/parser.py
brian-brazil: This is still n^2

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +170,126 @@ def _parse_labels(it, text):
...
+            label_value = sub_labels[quote_start:quote_end]
+            # Replace escaping if needed
+            if "\\" in label_value:
+                label_value = _replace_escaping(label_value)
+            labels[label_name.strip()] = label_value
+
+            # Remove the processed label from the sub-slice for next iteration
+            sub_labels = sub_labels[quote_end + 1:]
+            next_comma = sub_labels.find(",") + 1
+            sub_labels = sub_labels[next_comma:].lstrip()
```
brian-brazil: Why the lstrip?

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +170,126 @@ def _parse_labels(it, text):
...
+            # Fallback to parsing labels with a state machine
+            labels, labels_len = _parse_labels_with_state_machine(text[label_start + 1:])
+            label_end = labels_len + len(name) + 3  # We count the braces
+        # Parse the remaining text
+        remaining_text = text[label_end + 2:]
+        value, timestamp, exemplar = _parse_remaining_text(remaining_text)
+        return Sample(name, labels, value, timestamp, exemplar)
+
+    except ValueError as e:
+        if str(e).find("substring not found") > -1:
```
brian-brazil: This is still here, don't look inside error strings

prometheus_client/openmetrics/parser.py
ahmed-mez: it's n^2 in the worst case yes when we're escaping too many characters, I've mad a change by calling the function with smaller arguments `_is_character_escaped(value_substr[:i], i)` instead of `_is_character_escaped(value_substr, i)` 
Let me know if you're thinking of a possible better solution to optimize  👍

prometheus_client/openmetrics/parser.py
ahmed-mez: it's n^2 in the worst case yes when we're escaping too many characters, I've mad a change by calling the function with smaller arguments `_is_character_escaped(value_substr[:i], i)` instead of `_is_character_escaped(value_substr, i)`
Let me know if you're thinking of a possible better solution to optimize 👍

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +170,121 @@ def _parse_labels(it, text):
...
+    try:
+        # Process one label at a time
+        while sub_labels:
+            # The label name is before the equal
+            value_start = sub_labels.index("=")
+            label_name = sub_labels[:value_start]
+            sub_labels = sub_labels[value_start + 1:]
+
+            # Find the first quote after the equal
+            quote_start = sub_labels.index('"') + 1
```
brian-brazil: This is guaranteed to be right after the equals.

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +170,121 @@ def _parse_labels(it, text):
...
+            value_substr = sub_labels[quote_start:]
+
+            # Check for empty label name
+            if len(label_name) == 0:
```
brian-brazil: Shouldn't the MetricFamily code catch this already?

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +170,121 @@ def _parse_labels(it, text):
...
+            quote_end = i + 1
+            label_value = sub_labels[quote_start:quote_end]
+            # Replace escaping if needed
+            if "\\" in label_value:
+                label_value = _replace_escaping(label_value)
+            labels[label_name] = label_value
+
+            # Remove the processed label from the sub-slice for next iteration
+            sub_labels = sub_labels[quote_end + 1:]
+            next_comma = sub_labels.find(",") + 1
```
brian-brazil: This is guaranteed to be after the ", if present.

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +170,121 @@ def _parse_labels(it, text):
...
+    except ValueError:
+        raise ValueError("Invalid labels: " + text)
 
 
 def _parse_sample(text):
-    name = []
-    value = []
+    # Detect the labels in the text
+    try:
+        # `index` method raises a ValueError with
```
brian-brazil: Use find instead of index

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +170,121 @@ def _parse_labels(it, text):
...
+        # `substring not found` message if text 
+        # doesn't contain label braces
+        label_start = text.index("{")
+        # The name is before the labels
+        name = text[:label_start]
+        seperator = " # "
+        if not name.endswith("_bucket") or text.count(seperator) == 0:
+            # Line doesn't contain an exemplar
```
brian-brazil: What if it (incorrectly) does?

ahmed-mez: changed it to `if text.count(seperator) == 0:`
it should garantee that 👍

ahmed-mez:
Here is the benchmark after the changes we made in this PR, **the perfs got even better now ~3.9x**

```
call (x100000): _parse_sample('simple_metric 1.513767429e+09')

Simple example with prometheus parser: 0.24088597297668457
Simple example with openmetrics parser: 1.116285800933838
Simple example with the optimized openmetrics parser: 0.48735499382019043

call (x100000): _parse_sample('kube_service_labels{label_app="kube-state-metrics",label_chart="kube-state-metrics-0.5.0",label_heritage="Tiller",label_release="ungaged-panther",namespace="default",service="ungaged-panther-kube-state-metrics"} 1')

KSM metric example with prometheus parser: 1.608799934387207
KSM metric example openmetrics parser: 6.636054039001465
KSM metric example optimized openmetrics parser: 1.7176191806793213
```

brian-brazil:
That looks about right. Could you expand the unittests to ensure we're covering everything for both the old and new way of parsing labels? Also, it'd be great if you could add tests for any things this PR had incorrect at any point as if you've made this mistake others likely will too and the tests are going to be used as the openmetrics test suite.

ahmed-mez:
Added test cases, I guess the PR is ready for a final review :)

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +169,120 @@ def _parse_labels(it, text):
...
+    name = text[:label_start]
+    seperator = " # "
+    if text.count(seperator) == 0:
+        # Line doesn't contain an exemplar
+        # We can use `rindex` to find `label_end`
+        label_end = text.rindex("}")
+        label = text[label_start + 1:label_end]
+        labels = _parse_labels(label)
+    else:
+        # Line contains an exemplar
```
brian-brazil: potentially contains

prometheus_client/openmetrics/parser.py
```diff
@@ -141,37 +169,120 @@ def _parse_labels(it, text):
...
                 raise ValueError("Invalid line: " + text)
-    return labels
+        labels_len += 1
+    return labels, labels_len
+
+
+def _parse_labels(text):
+    labels = {}
+    # Return if we don't have valid labels
+    if "=" not in text:
```
brian-brazil: How would this handle something like `{a}`  ?

tests/openmetrics/test_parser.py
```diff
@@ -374,6 +374,48 @@ def test_timestamps(self):
...
+""")
+        hfm = HistogramMetricFamily("a", "help")
+        hfm.add_sample("a_bucket", {"le": "1.0", "foo": "bar # "}, 0.0, None, Exemplar({"a": "b", "foo": "bar # bar"}, 0.5))
+        hfm.add_sample("a_bucket", {"le": "2.0", "foo": "bar # "}, 2.0, None, Exemplar({"a": "c", "foo": "bar # bar"}, 0.5))
+        hfm.add_sample("a_bucket", {"le": "+Inf", "foo": "bar # "}, 3.0, None, Exemplar({"a": "d", "foo": "bar # bar"}, 4))
+        self.assertEqual([hfm], list(families))
+
+    @unittest.skipIf(sys.version_info < (3, 3), "Test requires Python 3.3+.")
+    def test_fallback_to_state_machine_label_parsing(self):
+        from unittest.mock import patch
```
brian-brazil: As these tests will become the official regression suite for OpenMetrics, it'd be best to test a full line rather than just a function.

prometheus_client/openmetrics/parser.py
ahmed-mez: made the required changes and added some use cases like that

tests/openmetrics/test_parser.py
ahmed-mez: added multiple test cases to assert what function are called 👍

brian-brazil:
Thanks!

ahmed-mez:
Great! @brian-brazil  :) any plans to release soon?

brian-brazil:
I'll add it to my todo list.

PR #403, opened 2019-05-08
Compact types.MethodType between 2 and 3

Hanaasagi:
In Python2, the `MethodType` constructor requires the obj‘s class to be passed. Compact code is copied from `six` https://github.com/benjaminp/six/blob/8da94b8a153ceb0d6417d76729ba75e80eaa75c1/six.py#L545-L556


Hanaasagi:
Is this ok?

brian-brazil:
Thanks!

PR #408, opened 2019-05-16
Fixes #407 by catching AttributeError in the exception

### LINKED ISSUE #407: except ImportError: not working as expected since python 3.3
I am running into some issues with prometheus_client trying import the resource package since I'm on a Windows machine. It looks like there was a previous fix #28 . However, that fix no longer works properly for some of us. I looks like Since Python 3.3+ you can import an empty folder and therefore when it try's to import the resource package it doesn't always raise an ImportError. It does however raise an AttributeError when it try's to get the page size.

I'll submit a pull request since it's a simple fix, but I figured I'd submit an issue for tracking first.

tyleracorn:
Python 3.3+ allows empty folders to be imported, so the ImportError doesn't always get raised. However the resource.getpagefilesize() will raise an AttributeError if you don't have the resource package

### DIFF
--- a/prometheus_client/process_collector.py
+++ b/prometheus_client/process_collector.py
@@ -11,7 +11,7 @@ try:
     import resource
 
     _PAGESIZE = resource.getpagesize()
-except(ImportError, AttributeError):
+except ImportError:
     # Not Unix
     _PAGESIZE = 4096
 
--- a/prometheus_client/process_collector.py
+++ b/prometheus_client/process_collector.py
@@ -11,7 +11,7 @@ try:
     import resource
 
     _PAGESIZE = resource.getpagesize()
-except ImportError:
+except(ImportError, AttributeError):
     # Not Unix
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you add the DCO, and a comment?

tyleracorn:
Ok not sure what I did there. I've never done the DCO stuff before and didn't realize I needed the signoff before. I seem to have possibly made it worse, I thought I fixed the signoff with the `git commit --amend --signoff` but then when I tried the push I received this error.
image

Do you know what I did wrong? I've kind of figured out at the least the signoff part, I think, so if it would be easier I could submit a new pull request.

brian-brazil:
If a new request is easier for you, go ahead.

brian-brazil:
Did you get around to this?

brian-brazil:
This has gone stale, let us know if you're still interested in this.

PR #411, opened 2019-05-21
Add remove files at worker exit

kimkkikki:
When Gunicorn is used as multiple processes,
Added code to delete the actual generated Guage file when the worker exits.

guage_all_{pid}.db
histogram_{pid}.db
counter_{pid}.db


brian-brazil:
The current behaviour is correct, you don't want counters going backwards.

kimkkikki:
Every time the worker restarts, the PID changes and a new guage file is created with the changed PID. The more the number of guage files in multiprocs_directory, the slower the metrics collection rate when collecting.
This PR has solved this problem for me, but I'm sorry it was not what you want. :)

brian-brazil:
Removing these files would break your metrics.

PR #413, opened 2019-05-22
Added _delta in Histogram and Summary

ratanboddu:
@brian-brazil 
Added a new field in Summary and Histogram named "_delta". As, "_sum" is cumulative, it's difficult to get the value("amount" in this case) by which the graph has increased and use those values for plotting a graph. Hence, "_delta" will contain the "amount" which can be used for plotting a graph and gaining much better insights.

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -412,12 +412,15 @@ class Summary(MetricWrapperBase):
         self._count = values.ValueClass(self._type, self._name, self._name + '_count', self._labelnames,
                                         self._labelvalues)
         self._sum = values.ValueClass(self._type, self._name, self._name + '_sum', self._labelnames, self._labelvalues)
+        self._delta = 0.0
         self._created = time.time()
+
 
     def observe(self, amount):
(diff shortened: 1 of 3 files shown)


brian-brazil:
If you want this you can use a Gauge. This is not generally useful, as it's only a arbitrary observation.

ratanboddu:
I actually had a constraint and had to use Summary. So, was looking for a way to to do it using Summary and couldn't find one, hence I've made the following changes.

ratanboddu:
Is there a workaround via which I'll be able to get all the Summary details as well as get the delta ?

brian-brazil:
Use a Gauge for this, though it's of very little use statistically.

ratanboddu:
Yes, Gauge is not as useful as Summary. So, was wondering if there could be a way to add something to Summary so as to be able to get much better analysis of the data at hand. Adding the delta field to Summary could very well help a user with their analysis.

brian-brazil:
This isn't something that'd be generally useful, and data model changes like this aren't something that can just be changed in one client.

Issue #412, opened 2019-05-22
Histogram.observe is slow O(n)


belm0:
Recording an observation is a slow O(n) on the number of bins.  By "slow" I mean the loop is in Python (as opposed to something like `sum()` where the loop is in C within the interpreter.

https://github.com/prometheus/client_python/blob/3cb4c9247f3f08dfbe650b6bdf1f53aa5f6683c1/prometheus_client/metrics.py#L525-L531

It should be straightforward to use the builtin bisect module for bin lookup.  It's O(log n) and implemented in C.

brian-brazil:
You're only going to have ~10 buckets, so it doesn't make a difference. In addition bisect may be less efficient due to branch misprediction.

belm0:
Given the default buckets in prometheus, bisect is 2x faster than linear search with a manual loop.

I would think that a top objective of a monitoring package is to not interfere with client performance.

Furthermore, given static bins, many applications will need a higher bin count to achieve reasonable accuracy over a few orders of magnitude (e.g. see conclusion in https://amplitude.com/blog/2014/08/06/optimal-streaming-histograms).

Really, prometheus could relieve users from knowing / guessing what their bins should be by using an approximate histogram which have the property of keeping as much info possible about the distribution given a constant number of bins-- but that's an issue for another day.

```python
from math import inf
from random import random
from collections import defaultdict
from bisect import bisect_right

buckets = (.005, .01, .025, .05, .075, .1, .25, .5, .75, 1.0, 2.5, 5.0, 7.5, 10.0, inf)
buckets2 = [i / 5 for i in range(50)] + [inf]

def linear_search(l, amount): 
    for i, bound in enumerate(l): 
        if amount <= bound:
            foo[i] += 1
            break

def bisect_search(l, amount):
    foo[bisect_right(l, amount)] += 1
    
%timeit linear_search(buckets, random() * 10)
%timeit linear_search(buckets2, random() * 10)
%timeit bisect_search(buckets, random() * 10)
%timeit bisect_search(buckets2, random() * 10)
```
```
1.08 µs ± 15.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
1.65 µs ± 10.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
515 ns ± 5.39 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
[...]

PR #415, opened 2019-05-24
use bisect for Histogram.observe()

belm0:
For #412.

Rationale:  less code; not slower for default buckets; benefit to anyone trying to use a large bucket count.


skudriashev:
Nice change!

brian-brazil:
Per the issue, I don't think this complexity is worth the tiny relative potential improvement and I'd worry about increased branch mispredictions.

PR #421, opened 2019-06-04
Multiprocess exposition speed boost

akx:
Hey there!

We at @valohai are still (#367, #368) bumping into situations where a long-lived multiproc (uWSGI) app's metric exposition steadily gets slower and slower, until it starts clogging up all workers and bumping into Prometheus's deadlines, and things break in a decidedly ungood way.

I figured I could take a little look at what exactly is taking time there, and I'm delighted to say I managed to eke out a roughly 5.8-fold speed increase in my test case (2857 files in the multiproc dir, totaling 2.8 GiB).

By far the largest boost here came from actually not using `mmap()` at all (f0319fa) when we're only reading the file; instead, simply reading the file fully into memory and parsing things from the memory buffer is much, much faster. Given each file (in my case anyway) is about 1 meg a pop, it shouldn't cause too much momentary memory pressure either.

Another decent optimization (0d8b870) came from looking at `vmprof`'s output (and remembering python-babel/babel#571); it said a lot of time was spent in small, numerous memory allocations within Python, and the trail led to `json.loads()`. Since the JSON blobs in the files are written with `sort_keys=True`, we can be fairly certain that there's going to be plenty of string duplication. Simply adding an unbounded `lru_cache()` to where the JSON strings are being parsed into (nicely immutable!) objects gave a nice speed boost and probably also reduced memory churn since the same objects get reused.  Calling `cache_info()` on the lru_cache validated the guess of duplication: `CacheInfo(hits=573057, misses=3229, maxsize=None, currsize=3229)`

A handful of other micro-optimizations brought the speed up a little more still.

My benchmark code was essentially 5 iterations of

```python
registry = prometheus_client.CollectorRegistry()
metrics_page = prometheus_client.generate_latest(registry)
assert metrics_page
```

and on my machine it took 22.478 seconds on 5132fd2 and 3.885 on 43cc95c91. 🎉

### DIFF
--- a/prometheus_client/mmap_dict.py
+++ b/prometheus_client/mmap_dict.py
@@ -37,9 +37,11 @@ class MmapedDict(object):
     def __init__(self, filename, read_mode=False):
         self._f = open(filename, 'rb' if read_mode else 'a+b')
         self._fname = filename
-        if os.fstat(self._f.fileno()).st_size == 0:
+        capacity = os.fstat(self._f.fileno()).st_size
+        if capacity == 0:
             self._f.truncate(_INITIAL_MMAP_SIZE)
-        self._capacity = os.fstat(self._f.fileno()).st_size
+            capacity = _INITIAL_MMAP_SIZE
+        self._capacity = capacity
         self._m = mmap.mmap(self._f.fileno(), self._capacity,
                             access=mmap.ACCESS_READ if read_mode else mmap.ACCESS_WRITE)
 
--- a/prometheus_client/mmap_dict.py
+++ b/prometheus_client/mmap_dict.py
@@ -81,8 +81,13 @@ class MmapedDict(object):
     @staticmethod
     def read_all_values_from_file(filename):
         with open(filename, 'rb') as infp:
-            data = infp.read()
-        return _read_all_values(data)
+            # Read the first block of data, including the first 4 bytes which tell us
+            # how much of the file (which is preallocated to _INITIAL_MMAP_SIZE bytes) is occupied.
+            data = infp.read(65535)
+            used = _unpack_integer(data, 0)[0]
+            if used > len(data):  # Then read in the rest, if needed.
+                data += infp.read(used - len(data))
+        return _read_all_values(data, used)
 
     def _init_value(self, key):
         """Initialize a value. Lock must be held by caller."""
--- a/prometheus_client/mmap_dict.py
+++ b/prometheus_client/mmap_dict.py
@@ -81,7 +81,6 @@ class MmapedDict(object):
         # on every loop iteration
         used = self._used
         data = self._m
-        unpack_from = struct.unpack_from
 
         while pos < used:
             encoded_len = _unpack_integer(data, pos)[0]
@@ -90,11 +89,11 @@ class MmapedDict(object):
                 msg = 'Read beyond file size detected, %s is corrupted.'
                 raise RuntimeError(msg % self._fname)
             pos += 4
-            encoded = unpack_from(('%ss' % encoded_len).encode(), data, pos)[0]
+            encoded_key = data[pos : pos + encoded_len]
             padded_len = encoded_len + (8 - (encoded_len + 4) % 8)
             pos += padded_len
             value = _unpack_double(data, pos)[0]
-            yield encoded.decode('utf-8'), value, pos
+            yield encoded_key.decode('utf-8'), value, pos
             pos += 8
 
     def read_all_values(self):
--- a/prometheus_client/mmap_dict.py
+++ b/prometheus_client/mmap_dict.py
@@ -22,6 +22,29 @@ def _pack_integer(data, pos, value):
     data[pos:pos + 4] = _pack_integer_func(value)
 
 
+def _read_all_values(data, used=0):
+    """Yield (key, value, pos). No locking is performed."""
+
+    if used <= 0:
+        # If not valid `used` value is passed in, read it from the file.
+        used = _unpack_integer(data, 0)[0]
+
+    pos = 8
+
+    while pos < used:
+        encoded_len = _unpack_integer(data, pos)[0]
+        # check we are not reading beyond bounds
+        if encoded_len + pos > used:
+            raise RuntimeError('Read beyond file size detected, file is corrupted.')
+        pos += 4
+        encoded_key = data[pos : pos + encoded_len]
+        padded_len = encoded_len + (8 - (encoded_len + 4) % 8)
+        pos += padded_len
+        value = _unpack_double(data, pos)[0]
+        yield encoded_key.decode('utf-8'), value, pos
+        pos += 8
+
+
 class MmapedDict(object):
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks, taking a quick peek this looks okay.

I'm surprised that mmap is slower, would all the files have been in page cache?
Do we have to worry about a memory leak in the LRU cache?

akx:
> I'm surprised that mmap is slower, would all the files have been in page cache?

I'm guessing the slower perf is because of small reads going through whatever wrappers Python has for mmaps, but who knows. (`vmprof` output was a little unreliable there.)

> Do we have to worry about a memory leak in the LRU cache?

I don't think so. The cache is [decorated here](https://github.com/prometheus/client_python/pull/421/commits/1d452b2f36cac3702e551cb2e53d0288781dcb5c#diff-52e98fdb8931e88f7d37feb48ba6ded2R56) to the lifetime of the `_parse_key()` function, which is an inner function of `_read_metrics`. That is to say, it's only live in memory for the duration of `MultiprocessCollector` reading the files off disk.

akx:
Another small optimization that occurred to me on the bus to work: The bulk of the mmap files is likely to be pretty empty (I mean, those 2.8 GiB of files gzcompressed down to 9.2 megs), so we can only read `used` bytes from them instead of the full megabyte (which is the initial mmap size), if we peek at the value first. (d8910d3)

brian-brazil:
> I'm guessing the slower perf is because of small reads going through whatever wrappers Python has for mmaps, but who knows. (vmprof output was a little unreliable there.)

It should only be hitting memory.

> The bulk of the mmap files is likely to be pretty empty (I mean, those 2.8 GiB of files gzcompressed down to 9.2 megs),

At most half should be empty, once it has expanded.

prometheus_client/multiprocess.py
```diff
@@ -82,21 +82,22 @@ def _parse_key(key):
     def _accumulate_metrics(metrics, accumulate):
         for metric in metrics.values():
             samples = defaultdict(float)
-            buckets = {}
+            buckets = defaultdict(lambda: defaultdict(float))
+            samples_setdefault = samples.setdefault
             for s in metric.samples:
-                name, labels, value = s.name, s.labels, s.value
+                name, labels, value, timestamp, exemplar = s
```
brian-brazil: timestamp and exemplar aren't used

prometheus_client/multiprocess.py
```diff
@@ -7,11 +7,23 @@
 import json
 import os
 
+try:
+    from functools import lru_cache
+except ImportError:
+    # `lru_cache` is not available on Python 2;
+    # we simply don't get the caching benefits on Python 2, then.
+    def lru_cache(maxsize=128):
```
brian-brazil: Considering this is not actually LRU, hand implementing it is fine and will give us 2.x support

svanscho:
Nice work! I honestly at first couldn't believe the current implementation cannot handle collecting and aggregating the metrics in a stable way over time. 

So two questions:
1. this seems like a lot faster, which is very nice and something we can definitely use asap
2. wouldn't a compaction of the data be better - do we need to keep all data when the pid are no longer active?

prometheus_client/multiprocess.py
akx: True. This is another micro-optimization: unpacking a slice of `s` would be slower than unpacking everything.

prometheus_client/multiprocess.py
akx: Sure thing. I'll do that when I get back to the office tomorrow. I suspect using a simple dict cache will be faster on 3.x too, I just wanted to use the included batteries to begin with.

akx:
> It should only be hitting memory.

If I'm reading https://github.com/python/cpython/blob/142566c028720934325f0b7fe28680afd046e00f/Modules/mmapmodule.c#L837-L867 and 
https://github.com/python/cpython/blob/142566c028720934325f0b7fe28680afd046e00f/Objects/bytesobject.c#L126 right, slicing an mmap object involves allocation and `memcpy`. 😞 That'd do it...

> At most half should be empty, once it has expanded.

When some of the worker processes don't live for very long at all, I was seeing files with `used` ranging in some tens of kilobytes. (Perhaps I should tune the initial size variable for this use case...)

akx:
> 2. wouldn't a compaction of the data be better - do we need to keep all data separately when the pid are no longer active?

It'd have to involve some interesting locking, since any process may be writing into their metrics file while we're collating them. Of course, if compaction is an opt-in step and you're willing to risk dropping a measurement or few while it's happening, it'd be an interesting next step for this. :)

svanscho:
Compaction can be done for all dead/inactive pids. It will avoid calculating the same sums etc over and over again without locking. The existing mechanism could then still be used for the active pids. No?

akx:
> Compaction can be done for all dead/inactive pids. It will avoid calculating the same sums etc over and over again without locking. The existing mechanism could then still be used for the active pids. No?

Sure, if you can know when the PID becomes inactive. With uWSGI, it's hard to know which pids are still active; at least I don't know about a hook for that...

prometheus_client/multiprocess.py
akx: Done in 5d0f877.

brian-brazil:
A PID may also become active again when a new process starts.

prometheus_client/mmap_dict.py
```diff
@@ -81,8 +81,14 @@ def __init__(self, filename, read_mode=False):
...
         with open(filename, 'rb') as infp:
-            data = infp.read()
-        return _read_all_values(data)
+            # Read the first block of data, including the first 4 bytes which tell us
+            # how much of the file (which is preallocated to _INITIAL_MMAP_SIZE bytes) is occupied.
+            data = infp.read(65535)
+            used = _unpack_integer(data, 0)[0]
+            if used > len(data):  # Then read in the rest, if needed.
+                data += infp.read(used - len(data))
+            assert len(data) >= used, ('short read for %s, file likely truncated' % filename)
```
brian-brazil: I don't think this can ever happen

akx:
> A PID may also become active again when a new process starts.

Which begs the question: does the `pid` have to necessarily be the real PID of the process? Could it just be an UUID that gets generated at process start?

prometheus_client/mmap_dict.py
akx: Sure, I can get rid of it. It was originally there to make sure my read length arithmetic was correct.

brian-brazil:
That isn't going to help with churn ;)

brian-brazil:
Thanks!

svanscho:
Yeah, I figured gunicorn has hooks on dead/inactive workers but not all implementations might have/use this. Anyway, awesome fix, thanks a lot! @brian-brazil when do you think we can make this into a release?

brian-brazil:
Tomorrow is the plan.

svanscho:
Awesome plan, thanks guys!

PR #422, opened 2019-06-05
very small change to fix errors when handling datetime object from aws

wuarthur:
Hi, I am trying to export CreateDate using the[ aws-prometheus-exporter ](https://github.com/movio/aws-prometheus-exporter) the exporter fails because of an error in [exposition.py ]

there seems to be an type error in this function, when v is expected to be a string, however sometimes aws return datas as an datetime object, which cause `v.replace('\\', r'\\')` errors out.
image

i found that this can be fixed by minimally changing line 78 in exposition.py to `k, str(v).replace('\\', r'\\').replace('\n', r'\n').replace('"', r'\"'))`

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -75,7 +75,7 @@ def generate_latest(registry=REGISTRY):
         if line.labels:
             labelstr = '{{{0}}}'.format(','.join(
                 ['{0}="{1}"'.format(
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you add a unittest for this?

brian-brazil:
This has gone stale, let us know if you still want to work on this.

PR #428, opened 2019-06-17
Only test twisted on 3.7 (latest minor)

xavfernandez:


tox.ini
```diff
@@ -29,7 +29,7 @@ deps =
 [testenv]
 deps =
     {[base]deps}
-    {py27,py34,py35,py36,pypy,pypy3}: twisted
+    {py27,py36,py37,pypy,pypy3}: twisted
```
brian-brazil: We don't need 3.6 either

xavfernandez:
@brian-brazil all good ?

brian-brazil:
Thanks!

PR #430, opened 2019-06-18
using worker_id instead of pid will produce fewer db files

fangjianfeng:
As the program is restarted frequently, more and more db files will be generated. These files are cumbersome to manage, and aggregate metrics also consume resources.

If the worker id is used as the db file naming method, as long as the number of wokers is determined and the metric type is unchanged, the number of generated db files will not increase, but the values ​​in the file are changing.

image

this will slow down the prometheus scrape duration.

### DIFF
--- a/prometheus_client/values.py
+++ b/prometheus_client/values.py
@@ -95,6 +95,12 @@ def MultiProcessValue(_pidFunc=os.getpid):
 
     return MmapedValue
 
+def _get_worker_id():
+    try:
+        import uwsgi
+        return uwsgi.worker_id()
+    except:
+        return 0
 
 def get_value_class():
     # Should we enable multi-process mode?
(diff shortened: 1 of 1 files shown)


brian-brazil:
This would break everyone not using uwsgi. Doing an import on every metric declaration would also be unwise.

fangjianfeng:
> This would break everyone not using uwsgi. Doing an import on every metric declaration would also be unwise.

uwsgi is used in my project, in order to avoid generating too many db files, modify it to this. Indeed, every time read metric, it will be imported once. I will consider how to optimize.

thanks

QuicksilverMachine:
Maybe the function for getting the ID could be configurable from the outside, so the default behavior could stay as it is if it's not set, but would allow for overrides for each individual case.

fangjianfeng:
> Maybe the function for getting the ID could be configurable from the outside, so the default behavior could stay as it is if it's not set, but would allow for overrides for each individual case?

There is no problem get it outside. 

and

If modify this method to use a class implementation, it is better to support user-defined functions that pass in <_pidFunc>

fangjianfeng:
> This would break everyone not using uwsgi. Doing an import on every metric declaration would also be unwise.

can we aggregate the metrics of non-current pid files into a specific file(count_0.db/histogram_0.db), so that the total number of pid files can be controlled without increasing.

brian-brazil:
This has gone stale, and would still break many users.

PR #441, opened 2019-07-18
Feature: user defined pidprovider

AMekss:
Similarly as in https://github.com/prometheus/client_python/pull/430 we are hitting problem when monitoring [RQ](https://python-rq.org/) workers. By design RQ worker forks new process for every new job, while it always execute one job at a time. The churn of process pids is insane and our worker pods are running out of memory in a few days of running because of this. 

Would be cool to have a possibility to override PID provider for such edge cases. This PR is an attempt to do so in less obtrusive way I managed to find. I'm not seasoned Python developer nor very familiar with all the ins and outs of this project, so feel free to give me hints on how proposed solution can be improved. 

This PR makes possible the following:
```
import uwsgi
from prometheus_client.pidprovider import Pidprovider
Pidprovider.source = uwsgi.worker_id

# or in case of RQ
Pidprovider.source = os.getppid
```

### DIFF
--- a/prometheus_client/mmap_dict.py
+++ b/prometheus_client/mmap_dict.py
@@ -37,7 +37,7 @@ def _read_all_values(data, used=0):
         if encoded_len + pos > used:
             raise RuntimeError('Read beyond file size detected, file is corrupted.')
         pos += 4
-        encoded_key = data[pos : pos + encoded_len]
+        encoded_key = data[pos: pos + encoded_len]
         padded_len = encoded_len + (8 - (encoded_len + 4) % 8)
         pos += padded_len
         value = _unpack_double(data, pos)[0]
--- a/prometheus_client/pidprovider.py
+++ b/prometheus_client/pidprovider.py
@@ -1,0 +1,9 @@
+import os
+
+
+class Pidprovider(object):
+    source = os.getpid
+
+    @classmethod
+    def getpid(cls):
+        return cls.source()
--- a/prometheus_client/values.py
+++ b/prometheus_client/values.py
@@ -4,7 +4,7 @@ import os
 from threading import Lock
 
 from .mmap_dict import mmap_key, MmapedDict
-
+from .pidprovider import Pidprovider
 
 class MutexValue(object):
     """A float protected by a mutex."""
@@ -27,8 +27,7 @@ class MutexValue(object):
         with self._lock:
             return self._value
 
-
-def MultiProcessValue(_pidFunc=os.getpid):
+def MultiProcessValue(_pidFunc=Pidprovider.getpid):
     files = {}
     values = []
     pid = {'value': _pidFunc()}
--- a/prometheus_client/values.py
+++ b/prometheus_client/values.py
@@ -6,6 +6,7 @@ from threading import Lock
 from .mmap_dict import mmap_key, MmapedDict
 from .pidprovider import Pidprovider
 
+
 class MutexValue(object):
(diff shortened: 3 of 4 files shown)


AMekss:
Looking to fix broken build..

brian-brazil:
I don't see how this works, any code like this would need to be called before anything that might instantiate a metric is.

brian-brazil:
That'd still have files for all the original pids though, so it'd make the
number of files higher

AMekss:
Sure, the correct way of setting this up is before importing any application logic and metrication code. I'm trying to figure out the worst case scenario.. where even if that would mistakenly happen afterwards it would not mess the metrics and be handled by pid change logic.

AMekss:
After playing around with python I figured out a way of how to achieve the same within a current implementation, leaving it here for anyone else tackling the same issue. Before any importing application code this snippet can customise PID provider function:
```
import os
from prometheus_client import values
values.ValueClass = values.MultiProcessValue(os.getppid)
```
It works for us while it feels a bit of a hack to me so would be cool have some official and better way of dealing with this edge case. Anyways closing this PR for now as it doesn't bring any value.

fangjianfeng:
> After playing around with python I figured out a way of how to achieve the same within a current implementation, leaving it here for anyone else tackling the same issue. Before importing any application code this snippet can customise PID provider function:
> 
> ```python
> import os
> from prometheus_client import values
> values.ValueClass = values.MultiProcessValue(os.getppid)
> ```
> 
> It works for us while it feels a bit of a hack to me so would be cool have some official and better way of dealing with this edge case. Anyways closing this PR for now as it doesn't bring any value.

can we aggregate the metrics of non-current pid files into a specific file(count_0.db/histogram_0.db), so that the total number of pid files can be controlled without increasing.

fangjianfeng:
> > After playing around with python I figured out a way of how to achieve the same within a current implementation, leaving it here for anyone else tackling the same issue. Before importing any application code this snippet can customise PID provider function:
> > ```python
> > import os
> > from prometheus_client import values
> > values.ValueClass = values.MultiProcessValue(os.getppid)
> > ```
> > 
> > 
> > It works for us while it feels a bit of a hack to me so would be cool have some official and better way of dealing with this edge case. Anyways closing this PR for now as it doesn't bring any value.
> 
> can we aggregate the metrics of non-current pid files into a specific file(count_0.db/histogram_0.db), so that the total number of pid files can be controlled without increasing.

The purpose of this is to control the number of pid files and not let it continue to grow. So, I think it's a viable way to aggregate all the db files from a period of time ago into a specific db file.

fangjianfeng:
> After playing around with python I figured out a way of how to achieve the same within a current implementation, leaving it here for anyone else tackling the same issue. Before importing any application code this snippet can customise PID provider function:
> 
> ```python
> import os
> from prometheus_client import values
> values.ValueClass = values.MultiProcessValue(os.getppid)
> ```
> 
> It works for us while it feels a bit of a hack to me so would be cool have some official and better way of dealing with this edge case. Anyways closing this PR for now as it doesn't bring any value.

The purpose of this is to control the number of pid files and not let it continue to grow. So, I think it's a viable way to aggregate all the db files from a period of time ago into a specific db file.

AMekss:
> can we aggregate the metrics of non-current pid files into a specific file(count_0.db/histogram_0.db), so that the total number of pid files can be controlled without increasing.

This could be the way, however I have no idea how efficient would be to merge files once pid got changed

PR #442, opened 2019-07-25
Support slashes in pgw grouping keys.

brian-brazil:

@beorn7


prometheus_client/exposition.py
```diff
@@ -361,6 +361,14 @@ def _use_gateway(method, gateway, job, registry, grouping_key, timeout, handler)
     )()
 
 
+def _escape_grouping_key(k, v):
+    if '/' in v:
+        # Added in Pushgateway 0.9.0.
+        return k + "@base64", base64.b64encode(v.encode("utf-8")).decode("utf-8")
```
beorn7: You need to use `base64.urlsafe_b64encode` to avoid `/` and `+` in the result.

brian-brazil: Fixed

PR #448, opened 2019-08-09
except OSError

Stelminator:
I'm trying to use this on Google App Engine standard (python 2.7) just to export some custom metrics.  This line throws an OSError.

I also had some trouble with the multiprocess mmap stuff.  Just dropping something into sys.modules['prometheus_client.mmap_dict'] seems to work well enough to solve that one, though.


brian-brazil:
Can you add the DCO?

Stelminator:
> Can you add the DCO?

updated.

brian-brazil:
Thanks!

PR #453, opened 2019-08-13
Add target_info to registries

brian-brazil:
This allows target labels as needed by push-based systems to be
provided, in a way that doesn't mess up Prometheus's own
top-down pull based approach to SD.

@SuperQ @RichiH @robskillington @sumeer Here's roughly what I'm thinking. Needs unittests etc.

@beorn7 Your input would be appreciated on the pgw front. https://github.com/OpenObservability/OpenMetrics/issues/63 is the general background. Pgw could continue to do this out of band for grouping labels (potentially with the client library using this as a default), or have this as an option.

### DIFF
--- a/prometheus_client/registry.py
+++ b/prometheus_client/registry.py
@@ -12,11 +12,13 @@ class CollectorRegistry(object):
     exposition formats.
     """
 
-    def __init__(self, auto_describe=False):
+    def __init__(self, auto_describe=False, target_info=None):
         self._collector_to_names = {}
         self._names_to_collectors = {}
         self._auto_describe = auto_describe
         self._lock = Lock()
+        self._target_info = {}
+        self.set_target_info(target_info)
 
     def register(self, collector):
         """Add a collector to the registry."""
@@ -69,8 +71,13 @@ class CollectorRegistry(object):
     def collect(self):
         """Yields metrics from the collectors in the registry."""
         collectors = None
+        ti = None
         with self._lock:
             collectors = copy.copy(self._collector_to_names)
+            if self._target_info:
+                ti = self._target_info_metric()
+        if ti:
+            yield ti
         for collector in collectors:
             for metric in collector.collect():
                 yield metric
@@ -87,11 +94,14 @@ class CollectorRegistry(object):
         Experimental."""
         names = set(names)
         collectors = set()
+        metrics = []
         with self._lock:
+            if 'target_info' in names and self._target_info:
+                metrics.append(self._target_info_metric())
+                names.remove('target_info')
             for name in names:
                 if name in self._names_to_collectors:
                     collectors.add(self._names_to_collectors[name])
-        metrics = []
         for collector in collectors:
             for metric in collector.collect():
                 samples = [s for s in metric.samples if s[0] in names]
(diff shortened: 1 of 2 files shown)


beorn7:
I don't see a blocker from the Pushgateway side. There are many options how the PGW could incorporate this. If it's not the client library taking the target_info as a default for the grouping key, it could even be done by the PGW itself.

prometheus_client/registry.py
```diff
@@ -106,6 +114,19 @@ def collect(self):
...
+        with self._lock:
+            self._target_info = labels
+
+    def get_target_info(self):
+        with self._lock:
+            return self._target_info
+
+    def _target_info_metric():
+        m = Metric('target', 'Target metadata', 'info')
+        m.add_sample('target_info', self._target_info, 1)
```
robskillington: I see, would it be required to also emit an actual value (i.e. in this case 1) or could we avoid that perhaps and special case it in the parser that metrics without a value are info metrics?

brian-brazil: This is a standard info metric, so the value of 1 is required.

brian-brazil:
I've added unittests, and a basic check that we're not exporting two target_infos.

prometheus_client/registry.py
```diff
@@ -106,6 +115,25 @@ def collect(self):
...
+        with self._lock:
+            if labels:
+                if not self._target_info and 'target_info' in self._names_to_collectors:
+                    raise ValueError('CollectorRegistry already contains a target_info metric')
+                self._names_to_collectors['target_info'] = None
+            elif self._target_info:
+                self._names_to_collectors.pop('target_info', None)
+            self._target_info = labels
+
+    def get_target_info(self):
```
robskillington: Is `get_target_info` meant to be used from anywhere?  Or mainly for the public API of get/set?

prometheus_client/registry.py
```diff
@@ -106,6 +115,25 @@ def collect(self):
...
         return RestrictedRegistry()
 
+    def set_target_info(self, labels):
+        with self._lock:
+            if labels:
+                if not self._target_info and 'target_info' in self._names_to_collectors:
+                    raise ValueError('CollectorRegistry already contains a target_info metric')
+                self._names_to_collectors['target_info'] = None
+            elif self._target_info:
+                self._names_to_collectors.pop('target_info', None)
```
robskillington: Since the result of `.pop(...)` is not used, does it make sense to not pass an argument at all for that param rather than `None`?

prometheus_client/registry.py
```diff
@@ -106,6 +115,25 @@ def collect(self):
 
         return RestrictedRegistry()
 
+    def set_target_info(self, labels):
+        with self._lock:
+            if labels:
+                if not self._target_info and 'target_info' in self._names_to_collectors:
+                    raise ValueError('CollectorRegistry already contains a target_info metric')
+                self._names_to_collectors['target_info'] = None
```
robskillington: I'm a little confused about this one, from my rudimentary understanding (or extreme rudimentary understanding should I say) - every value in `_names_to_collectors` should be a collector yeah?

If None is a value in `_names_to_collectors` and then it gets added to `collectors` var in `restricted_registry`:
```python
        names = set(names)
        collectors = set()
        metrics = []
        with self._lock:
            for name in names:
                if name in self._names_to_collectors:
                    collectors.add(self._names_to_collectors[name])
            if 'target_info' in names and self._target_info:
                metrics.append(self._target_info_metric())
        for collector in collectors:
                samples = [s for s in metric.samples if s[0] in names]
                if samples:
                    m = Metric(metric.name, metric.documentation, metric.type)
                    m.samples = samples
                    metrics.append(m)
```

Wouldn't the `collector.collect()` be a call to `None` from the line `for metric in collector.collect()` if `names` happens to contain `target_info`?

robskillington:
Hm, I was able to reproduce the issue I described in the comment I left:
```
>>> import prometheus_client
>>> registry = prometheus_client.CollectorRegistry()
>>> registry.set_target_info({'foo':'bar'})
>>> results = registry.restricted_registry(['target_info'])
  File "prometheus_client/registry.py", line 105, in restricted_registry
AttributeError: 'NoneType' object has no attribute 'collect'
```

brian-brazil:
Right you are, fixed and unittest added.

Issue #454, opened 2019-08-19
'Counter' object has no attribute '_value'


AlexPadron:
Hi all, 

I'm running into an issue when creating an exception counter out of a standard counter without passing label names. Specifically, it does not provide an informative error if a user incorrectly passes labels when calling `count_exceptions`. Code to replicate:

```
from prometheus_client import Counter


c = Counter('counter', documentation='A sample counter', labelnames=('label',))

with c.count_exceptions():
    raise RuntimeError('panic!')
```

Expected behavior: `c.count_exceptions()` should raise an informative error message because the incorrect labels are passed.

Actual behavior: `AttributeError: 'Counter' object has no attribute '_value'` occurs when the exception is raised, because we are trying to increment an non observable metric.

Proposed fix: Check `self._is_observable()` in the `count_exceptions` method, and raise an error if the user is constructing an `ExceptionCounter` out of a non observable `Counter`.

brian-brazil:
Do you want to send a PR?

AlexPadron:
sure!

AlexPadron:
I think I need access to the repo, not currently able to push to a remote branch

brian-brazil:
Fork it in github, and push your branch there.

AlexPadron:
https://github.com/prometheus/client_python/pull/455

PR #455, opened 2019-08-19
Add observability check to counters

AlexPadron:
This is for issue https://github.com/prometheus/client_python/issues/454

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -362,6 +362,7 @@ class Gauge(MetricWrapperBase):
         Increments the gauge when the code is entered,
         and decrements when it is exited.
         """
+        self._raise_if_not_observable()
         return InprogressTracker(self)
 
     def time(self):
@@ -369,6 +370,7 @@ class Gauge(MetricWrapperBase):
 
         Can be used as a function decorator or context manager.
         """
+        self._raise_if_not_observable()
         return Timer(self.set)
 
     def set_function(self, f):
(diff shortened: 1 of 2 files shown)


prometheus_client/metrics.py
```diff
@@ -58,6 +58,16 @@ def _is_observable(self):
         # * the child of a labelled metric.
         return not self._labelnames or (self._labelnames and self._labelvalues)
 
+    def _raise_if_not_observable(self):
+        # Functions that mutate the state of the metric, for example incrementing
+        # a counter, will fail if the metric is not observable, because only if a
+        # metric is observable will the value be initialized.
+        if not self._is_observable():
+            raise ValueError(
+                '%s metric is not observable. This is likely because the metric requires '
```
AlexPadron: Thinking if we want to make this more detailed, for example we could use inspect and get the method name that is being invoked. Also, we may want to use this in other metrics as well. I added the check to counters, but the same issue exists in gauges as well for example

prometheus_client/metrics.py
```diff
@@ -239,8 +249,11 @@ def _metric_init(self):
 
     def inc(self, amount=1):
         """Increment counter by the given amount."""
+        self._raise_if_not_observable()
```
brian-brazil: This is a hot path, we shouldn't add code here.

prometheus_client/metrics.py
brian-brazil: I'm not sure "observable" is the right word to use here, it'll confuse users. Missing label values would be more accurate.

brian-brazil:
Other metrics types also need this, e.g. the time() functions.

AlexPadron:
> Other metrics types also need this, e.g. the time() functions.

I think I got all of these (at least in metrics.py). As far as I can tell by grepping this is the only location where the metrics are defined

brian-brazil:
Thanks!

PR #460, opened 2019-09-05
Add checks for various other float formats.

brian-brazil:
Check for new stuff Go added in 1.13, and also
make sure leading 0s are permitted as some users
want to be able to do fixed-size output.

@SuperQ


brian-brazil:
Tests caught that Python 3 also supported some of this, so fixed that too.

PR #462, opened 2019-09-07
Unbundle decorator

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


brian-brazil:
We purposefully avoid dependencies in the core. Is this causing you an issue?

ignatenkobrain:
I can definitely carry this patch downstream in Fedora, but why dependenices are a problem?

brian-brazil:
It increases the risk of versioning conflicts etc. I want this library to be as risk free to use as practical.

brian-brazil:
This has gone stale, and I don't see a need for this change.

PR #466, opened 2019-09-13
docs: update docs connected with gauge for better understanding

Arfey:


brian-brazil:
Thanks for your PR. The current example is self contained, this doesn't seem to make much of a difference.

Arfey:
When I read documentation I didn't understand how it's work.

brian-brazil:
This has gone stale, and I'm still not seeing this as an improvement.

PR #467, opened 2019-09-14
Fixed some handler issues in mmap_dict

Jallasia:
Unit test from test_multiprocess were not working on windows, it was because shutil.rmtree were ending up with a Permission error because of some remaining mmap file handlers.
Added:
* __del__ that call close on mmap
* a retry_after_close_handler method for testing to make sure we can indeed delete tmp resources that display a warning in case some handler were closed

### DIFF
--- a/prometheus_client/mmap_dict.py
+++ b/prometheus_client/mmap_dict.py
@@ -136,6 +136,9 @@ class MmapedDict(object):
             self._f.close()
             self._f = None
 
+    def __del__(self):
+        self.close()
+
 
 def mmap_key(metric_name, name, labelnames, labelvalues):
     """Format a key for use in the mmap file."""
--- a/tests/test_multiprocess.py
+++ b/tests/test_multiprocess.py
@@ -5,6 +5,9 @@ import os
 import shutil
 import sys
 import tempfile
+import warnings
+
+import gc
 
 from prometheus_client import mmap_dict, values
 from prometheus_client.core import (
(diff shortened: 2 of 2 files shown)


Jallasia:
Note that there might be something to improve because even after deleting all created objets there are still refs on the mmap objet (old in the cache)

brian-brazil:
This seems a bit hacky, how about ensuring we're calling close() where we should be?

Jallasia:
Well __del__ is call by the garbage collector when object if not referenced anymore.
meaning that this will be call only if nothing else can access it.
But it's not mandatory anyway, do you want me just to add the fix of checks ?

brian-brazil:
It seems a bit odd to be depending on GC for this, I think something more explicit would be better

brian-brazil:
Are you still working on this?

Jallasia:
well the fix is done, do you need something else before integration ?

brian-brazil:
I'd like a solution that doesn't require poking around inside GC internals, or being aware of GC at all.

Jallasia:
Ok, as long as it's making the test work on windows, that's fine for me.

Jallasia:
Do you want me to close the PR ?

brian-brazil:
I'd prefer to find a better way to fix the tests on Windows.

Jallasia:
ok when do you think you can take care of it ?

brian-brazil:
I don't use Windows, so someone like you would need to propose a different solution.

Jallasia:
ok i will wait for that.

PR #469, opened 2019-09-28
Adding uppercase PROMETHEUS_MULTIPROC_DIR

andersbogsnes:
Allowed both prometheus_multiproc_dir and PROMETHEUS_MULTIPROC_DIR
as valid environment variables

It's been a common point of confusion for our users,  as most environment variables are uppercase  they believe they have set the variable, when in fact they forgot to lowercase it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ dist
 .tox
 .*cache
 htmlcov
+.idea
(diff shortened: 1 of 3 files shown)


brian-brazil:
I think one way to do it is simplest, and less likely to cause confusion overall.

PR #478, opened 2019-10-24
Add support for Python 3.8

asherf:
Python 3.8 was released recently. This runs tests and adds the classifer to indicate Python 3.8 is supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 sudo: false
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -22,11 +23,15 @@ matrix:
       env: TOXENV=py36
     - python: "3.7"
       env: TOXENV=py37
-      dist: xenial
       sudo: true
     - python: "3.7"
       env: TOXENV=py37-nooptionals
```
brian-brazil: We only need this for the latest.

asherf: @brian-brazil I'm not sure I understand...
I made this dist (xenial) the default for everything (I think Python 3.8 needs it too).
and just changed the dist for 2.6 to trusty since python 2.6 doesn't work in xenial...

brian-brazil: We only need the nooptionals for the latest

asherf: ah! got it.
will move it to 3.8 then!

brian-brazil:
Thanks!

PR #481, opened 2019-10-26
Add __repr__ method to metric objects, make them debug friendly.

asherf:
before:
Screenshot 2019-10-25 18 44 10

after:
Screenshot 2019-10-25 18 43 40

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -80,9 +80,12 @@ class MetricWrapperBase(object):
             metric.add_sample(self._name + suffix, labels, value)
         return [metric]
 
+    def __str__(self):
+        return "{0}:{1}".format(self._type, self._name)
+
     def __repr__(self):
         metric_type = type(self)
-        return "{0}.{1}:{2}".format(metric_type.__module__, metric_type.__name__, self._name)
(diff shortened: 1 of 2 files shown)


prometheus_client/metrics.py
```diff
@@ -80,6 +80,10 @@ def collect(self):
             metric.add_sample(self._name + suffix, labels, value)
         return [metric]
 
+    def __repr__(self):
+        metric_type = type(self)
+        return "{0}.{1}:{2}".format(metric_type.__module__, metric_type.__name__, self._name)
```
brian-brazil: A repr() should in principle allow instantiation of the object. This would be a `__str__`. Also _type covers you more here.

asherf: As far as using the _type, I added an `__str__` implementation that uses it.
I think having the full type path is useful in repr output which what is used in debuggers.
I updated the `__repr__` implementation to use parenthesis instead of a column to be more compatible with the ability to [pass eval](https://stackoverflow.com/a/1436756/38265)
However, I feel that making it a 100% compatible with that (i.e. an eval(repr(metric)) that produces an identical object) will result in a string that will be to noisy and thus less useful when debugging.

brian-brazil:
Thanks!

PR #482, opened 2019-10-26
Inject readme.md to long_description so it will show up on the pypi project page.

asherf:
See: https://packaging.python.org/guides/making-a-pypi-friendly-readme/
https://packaging.python.org/guides/making-a-pypi-friendly-readme/#including-your-readme-in-your-package-s-metadata


brian-brazil:
Thanks!

Issue #461, opened 2019-09-07
Difference in parser behaviour compared to prometheus/common/expfmt 


zegelin:
The behaviour of the Python client's built-in text format parser differs to that of the parser found inside Prometheus itself.

Given the following (poorly formed, but apparently still valid) exposition output:

```
test_metric_a {hello="world"} 0 0
test_metric_b {hello="world"} 0 0
test_metric_a {hello="universe"} 0 0
test_metric_b {hello="universe"} 0 0
```

The Go expfmt `TextParser.TextToMetricFamilies(...)` function produces:

```
(map[string]*io_prometheus_client.MetricFamily) (len=2) {
 (string) (len=13) "test_metric_a": (*io_prometheus_client.MetricFamily)(0xc0000c4320)(name:"test_metric_a" type:UNTYPED metric:<label:<name:"hello" value:"world" > untyped:<value:0 > timestamp_ms:0 > metric:<label:<name:"hello" value:"universe" > untyped:<value:0 > timestamp_ms:0 > ),
 (string) (len=13) "test_metric_b": (*io_prometheus_client.MetricFamily)(0xc0000c4370)(name:"test_metric_b" type:UNTYPED metric:<label:<name:"hello" value:"world" > untyped:<value:0 > timestamp_ms:0 > metric:<label:<name:"hello" value:"universe" > untyped:<value:0 > timestamp_ms:0 > )
}
```

whereas the Python client's `parser.text_fd_to_metric_families(...)` function produces:

```
[
    Metric(test_metric_a, , unknown, , [Sample(name='test_metric_a', labels={'hello': 'world'}, value=0.0, timestamp=None, exemplar=None)]),
    Metric(test_metric_b, , unknown, , [Sample(name='test_metric_b', labels={'hello': 'world'}, value=0.0, timestamp=None, exemplar=None)]),
    Metric(test_metric_a, , unknown, , [Sample(name='test_metric_a', labels={'hello': 'universe'}, value=0.0, timestamp=None, exemplar=None)]),
    Metric(test_metric_b, , unknown, , [Sample(name='test_metric_b', labels={'hello': 'universe'}, value=0.0, timestamp=None, exemplar=None)])
]
```

As you can see the Go parser groups the samples into two untyped metric families with two samples each whereas the Python parser creates a new metric family for each sample.

brian-brazil:
That's not the parser inside Prometheus itself. Is there a problem that this is causing you?

asherf:
[...]

PR #483, opened 2019-10-26
Add support for parsing timestamps.

asherf:
Closes: https://github.com/prometheus/client_python/issues/461

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -273,7 +273,7 @@ a{foo="b\\\\a\\z"} 2
         metric_family.add_metric(["b\\a\\z"], 2)
         self.assertEqualMetrics([metric_family], list(families))
 
-    def test_timestamps_discarded(self):
+    def test_timestamps(self):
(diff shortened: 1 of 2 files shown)


tests/test_parser.py
```diff
@@ -282,8 +282,9 @@ def test_timestamps_discarded(self):
 b 2  1234567890
 """)
         a = CounterMetricFamily("a", "help", labels=["foo"])
-        a.add_metric(["bar"], 1)
-        b = CounterMetricFamily("b", "help", value=2)
+        a.add_metric(["bar"], 1, timestamp=0)
+        b = CounterMetricFamily("b", "help")
+        b.add_metric([], 2, timestamp=1234567890)
```
brian-brazil: This isn't right, Prometheus timestamp are in milliseconds whereas the Python client is using the OpenMetrics data model so is working in seconds.

brian-brazil:
Thanks!

Issue #457, opened 2019-08-28
python wheel


gjcarneiro:
Would it be possible, from now on, to also generate and upload Python wheels?  It will speed up installation of this package a lot...

brian-brazil:
This package is pure python, would it make a difference?

gjcarneiro:
Yes, it makes a huge difference, installation from Python wheel, even for pure Python, is _so_ much faster, as installation is little more than downloading a zip file and uncompressing it directly to the final destination.  A normal package installation, not only needs uncompressing to a temporary src directory, then it needs to execute `python setup.py install`, which is a lot slower.

https://pythonwheels.com

brian-brazil:
Can you quantify "a lot"? 200kB doesn't take a lot to process.

xavfernandez:
It mainly saves pip from spawning a process to launch `python setup.py bdist_wheel` and automagically create the wheel :)

 ```
(tmp-f8893c0a100797) [xfernandez@xaferiba tmp-f8893c0a100797]$ pip uninstall prometheus-client -y
Uninstalling prometheus-client-0.7.1:
  Successfully uninstalled prometheus-client-0.7.1
(tmp-f8893c0a100797) [xfernandez@xaferiba tmp-f8893c0a100797]$ time pip install --no-cache-dir --no-index ~/Downloads/prometheus_client-0.7.1.tar.gz
Processing /home/xfernandez/Downloads/prometheus_client-0.7.1.tar.gz
Building wheels for collected packages: prometheus-client
  Building wheel for prometheus-client (setup.py) ... done
  Created wheel for prometheus-client: filename=prometheus_client-0.7.1-cp37-none-any.whl size=41402 sha256=66668319af083af049349c7e8e81f7def84260e36c9e0968b753e27fefd5c421
  Stored in directory: /tmp/pip-ephem-wheel-cache-sh8aaupe/wheels/c3/4f/4e/65ac82a9f5c3008dee932dbf792e6cf5f9a3ae9de9031735a9
Successfully built prometheus-client
Installing collected packages: prometheus-client
Successfully installed prometheus-client-0.7.1

real    0m0.760s
user    0m0.677s
sys     0m0.083s
(tmp-f8893c0a100797) [xfernandez@xaferiba tmp-f8893c0a100797]$ pip uninstall prometheus-client -y
Uninstalling prometheus-client-0.7.1:
  Successfully uninstalled prometheus-client-0.7.1
[...]

PR #487, opened 2019-11-14
Update setup.py to produce universal wheels

xavfernandez:
To ease the transition to #457:

    `python setup.py sdist bdist_wheel`

will produce the universal wheel in addition to the sdist.


brian-brazil:
Thanks!

PR #489, opened 2019-11-18
Fix some typos

asherf:


prometheus_client/openmetrics/parser.py
```diff
@@ -214,7 +214,7 @@ def _parse_labels(text):
                     break
                 i += 1
 
-            # The label value is inbetween the first and last quote
```
brian-brazil: inbetween is a word

brian-brazil:
Thanks!

PR #492, opened 2019-12-06
Return thread object from start_http_server

mgalgs:
Callers of start_http_server might want to know when the server has exited.
Return the thread object so that it can be join()ed for this purpose.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -194,6 +194,7 @@ def start_http_server(port, addr='', registry=REGISTRY):
     t = threading.Thread(target=httpd.serve_forever)
     t.daemon = True
     t.start()
+    return t
 
 
 def write_to_textfile(path, registry):


brian-brazil:
Per the previous issues around this, this function is to get you up and running quickly. If you have more complex needs, you should create your own http server.

Also, this server should never exit.

mgalgs:
@brian-brazil 

> Per the previous issues around this, this function is to get you up and running quickly. If you have more complex needs, you should create your own http server.

Of course, but if the goal is to help people get up and running quickly why would we want them to copy/paste this function just so that they can keep a hold of the thread?  I don't have any more complex needs, I am just trying to get up and running quickly*, I just need to block and would prefer not to put a `while True: time.sleep(...)` in my code...  What does it hurt to return the thread object?  The code looks weird to my eyes, firing off a thread and then just letting go of all references to it.

Call it my OCD, call me a `time.sleep()` bigot, but when I call a function that fires off a thread I expect to get a reference to that puppy for safekeeping.

\* although if that were really the case I'd have copy/pasted this long ago ;) Now I'm just thinking of my fellow developers who might have the same issue.

> Also, this server should never exit.

Best practice when you fire off a thread is to join it, even if it "should" never exit.

brian-brazil:
> why would we want them to copy/paste this function just so that they can keep a hold of the thread? 

They shouldn't need to care about the thread, and generally they'll already have a http server they wish to use.

> What does it hurt to return the thread object?

That's defining an api, and we could never change it to some other object - and if you're going down that path better to write something yourself. It's only a few lines of code.

PR #493, opened 2019-12-06
Get rid of a confusing comment about time going backwards

gmossessian:
Use the monotonic timer to avoid issues with time going backwards, and a confusing comment.

### DIFF
--- a/prometheus_client/context_managers.py
+++ b/prometheus_client/context_managers.py
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 
-from timeit import default_timer
+from time import monotonic
 
 from .decorator import decorate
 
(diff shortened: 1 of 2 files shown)


brian-brazil:
We explicitly avoid 3rd party dependencies, to reduce the changes of bad interactions with code that users have. The comment is correct.

gmossessian:
@brian-brazil OK thank you, closing PR unless I find another workaround.

brian-brazil:
Is there an actual problem with how the code is currently behaving?

gmossessian:
No. We were adding some timing decorators to get function-level processing time data into logs and looked into prometheus' implementation to see whether there were any gotchas we had missed. The comment about time going backwards bothered me and I wanted to get rid of it. It could still be done without the 3rd party library, but you the maintainers probably don't want to deal with the extra complexity for the sake a smattering of requests that will erroneously report 0 time.

PR #503, opened 2020-01-22
Fix pushgateway example url

k0t3n:
urllib requires a full URL


brian-brazil:
If this doesn't work then it's a regression to be fixed.

k0t3n:
@brian-brazil sorry, I didn’t quite understand what you mean.

brian-brazil:
I've confirmed that the existing example is correct.

PR #506, opened 2020-01-30
Travis CI: The sudo: tag is now fully deprecated

cclauss:
__sudo:__ is now deprecated in Travis CI and it has been removed from the Travis docs.

Also: Travis CI has replaced the __matrix:__ keyword with __jobs:__

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,11 @@
 dist: xenial
-sudo: false
 cache:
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you add the DCO?

brian-brazil:
We're moving off Travis, so this is obselete.

Issue #504, opened 2020-01-29
gauge.set_function() doesn't work in multiprocess mode


MatthewMaclean:
Multiprocess mode's [collect()](https://github.com/prometheus/client_python/blob/470b2d3849fe069a6d390f0818b91b75c9fde79e/prometheus_client/multiprocess.py#L147) reads the registry files and aggregates metrics that have been written to `prometheus_multiproc_dir`.

This doesn't work with `gauge.set_function()` which does not record its value. The provided function is just [called during collection](https://github.com/prometheus/client_python/blob/470b2d3849fe069a6d390f0818b91b75c9fde79e/prometheus_client/registry.py#L82).

That means with the following code:
```
registry = CollectorRegistry()
Gauge("test", "test", registry=self._registry).set_function(lambda: 100)
multiprocess.MultiProcessCollector(registry)
```

The output will be:
```
# HELP test test
# TYPE test gauge
test 100.0
# HELP test Multiprocess metric
# TYPE test gauge
test{pid="10705"} 0.0
```

Current side effects:
- If any mode other than `all` or `liveall` is used, the `pid` tag won't be included. This results in duplicate metrics being reported to Prometheus. Prometheus currently only uses the first metric it reads, which is non-deterministic due to iteration over the [registry's dictionary](https://github.com/prometheus/client_python/blob/470b2d3849fe069a6d390f0818b91b75c9fde79e/prometheus_client/registry.py#L81).
- If `registry=None` to avoid double reporting, only the default value of `0.0` is reported.
- Current way to work around it is to use the mode `all` and to ignore gauges in Prometheus and with the tag `pid`.

Proposal:
I'm not sure how you could incorporate `set_function` into the multiprocess registry and I'm not convinced how useful of a feature it would be. Is it reasonable to add a new `multiprocess_mode: exclude` which would prevent the incorrect 0.0 value being reported? Or would it be better to just add documentation to recommend using two independent registries?

brian-brazil:
This classifies as a custom collector, which it's not possible to make work with multiprocess mode.

MatthewMaclean:
Makes sense, I'll look/ask about ways to work around that in the mailing list.

hdost:
Given that multiprocess is getting some traction is it worth re-opening this?

csmarchbanks:
[...]

PR #507, opened 2020-01-30
gauge.set_function is a custom collector

MatthewMaclean:
Clarifying in the Multiprocess Mode instructions that gauge.set_function is also a custom collector.
https://github.com/prometheus/client_python/issues/504


brian-brazil:
Can you add the DCO?

PR #511, opened 2020-02-12
include tests on pypi distribution

pacoesteban:
I'm trying to update the OpenBSD port of this software and add tests in the process.

In the packaging process the regression tests are used in order to make sure that updates
work properly and that an update of a dependency doesn't break a package.
Having the regression tests in the PyPI tarball would make that much easier.


brian-brazil:
Thanks!

PR #512, opened 2020-02-14
Added ASGI application

Skeen:
This PR introduces an ASGI counterpart to the current WSGI app.

The ASGI code itself is moved into a seperate module `asgi.py`, which is conditionally included as it is only valid in Python 3 (due to async/await calls).

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,8 +311,6 @@ start_wsgi_server(8000)
 To use Prometheus with [ASGI](http://asgi.readthedocs.org/en/latest/), there is
 `make_asgi_app` which creates an ASGI application.
 
-Save the snippet below in a `myapp.py` file
-
 ```python
 from prometheus_client import make_asgi_app
 
@@ -320,17 +318,6 @@ app = make_asgi_app()
 ```
 Such an application can be useful when integrating Prometheus metrics with ASGI
 apps.
-
-The app can be used to serve the metrics through an ASGI implementation, such
-as [daphne](https://github.com/django/daphne) or
-[uvicorn](https://www.uvicorn.org/).
-```bash
-# Install daphne if you do not have it
-pip install daphne
-daphne myapp:app
-```
-
-Visit http://localhost:8000/ to see the metrics
 
 #### Flask
 
--- a/README.md
+++ b/README.md
@@ -306,6 +306,32 @@ from prometheus_client import start_wsgi_server
 start_wsgi_server(8000)
 ```
 
+#### ASGI
+
+To use Prometheus with [ASGI](http://asgi.readthedocs.org/en/latest/), there is
+`make_asgi_app` which creates an ASGI application.
+
+Save the snippet below in a `myapp.py` file
+
+```python
+from prometheus_client import make_asgi_app
+
+app = make_asgi_app()
+```
+Such an application can be useful when integrating Prometheus metrics with ASGI
+apps.
+
+The app can be used to serve the metrics through an ASGI implementation, such
+as [daphne](https://github.com/django/daphne) or
+[uvicorn](https://www.uvicorn.org/).
+```bash
+# Install daphne if you do not have it
+pip install daphne
+daphne myapp:app
+```
+
+Visit http://localhost:8000/ to see the metrics
+
 #### Flask
 
 To use Prometheus with [Flask](http://flask.pocoo.org/) we need to serve metrics through a Prometheus WSGI application. This can be achieved using [Flask's application dispatching](http://flask.pocoo.org/docs/latest/patterns/appdispatch/). Below is a working example.
--- a/prometheus_client/asgi.py
+++ b/prometheus_client/asgi.py
@@ -25,7 +25,7 @@ def make_asgi_app(registry=REGISTRY):
                     "type": "http.response.start",
                     "status": int(status.split(' ')[0]),
                     "headers": [
-                        (x.encode('utf8') for x in header)
+                        tuple(x.encode('utf8') for x in header)
                     ]
                 }
             )
--- a/prometheus_client/asgi.py
+++ b/prometheus_client/asgi.py
@@ -9,12 +9,15 @@ def make_asgi_app(registry=REGISTRY):
 
     async def prometheus_app(scope, receive, send):
         assert scope.get("type") == "http"
+        # Prepare parameters
         params = parse_qs(scope.get('query_string', b''))
         accept_header = "Accept: " + ",".join([
             value.decode("utf8") for (name, value) in scope.get('headers')
             if name.decode("utf8") == 'accept'
         ])
-        status, headers, output = _bake_output(registry, accept_header, params)
+        # Bake output
+        status, header, output = _bake_output(registry, accept_header, params)
+        # Return output
         payload = await receive()
         if payload.get("type") == "http.request":
             await send(
@@ -22,8 +25,7 @@ def make_asgi_app(registry=REGISTRY):
                     "type": "http.response.start",
                     "status": int(status.split(' ')[0]),
                     "headers": [
-                        [key.encode('utf8'), value.encode('utf8')]
-                        for (key,value) in headers
+                        (x.encode('utf8') for x in header)
                     ]
                 }
             )
--- a/prometheus_client/asgi.py
+++ b/prometheus_client/asgi.py
@@ -1,0 +1,34 @@
+from urllib.parse import parse_qs
+
+from .exposition import choose_encoder
+from .registry import REGISTRY
+
+
+def make_asgi_app(registry=REGISTRY):
+    """Create a ASGI app which serves the metrics from a registry."""
+
+    async def prometheus_app(scope, receive, send):
+        assert scope.get("type") == "http"
+        params = parse_qs(scope.get('query_string', b''))
+        r = registry
+        accept_header = "Accept: " + ",".join([
+            value.decode("utf8") for (name, value) in scope.get('headers')
+            if name.decode("utf8") == 'accept'
+        ])
+        encoder, content_type = choose_encoder(accept_header)
+        if 'name[]' in params:
+            r = r.restricted_registry(params['name[]'])
+        output = encoder(r)
+
+        payload = await receive()
+        if payload.get("type") == "http.request":
+            await send(
+                {
+                    "type": "http.response.start",
+                    "status": 200,
+                    "headers": [[b"Content-Type", content_type.encode('utf8')]],
+                }
+            )
+            await send({"type": "http.response.body", "body": output})
+
+    return prometheus_app
--- a/prometheus_client/asgi.py
+++ b/prometheus_client/asgi.py
@@ -1,6 +1,6 @@
 from urllib.parse import parse_qs
 
-from .exposition import choose_encoder
+from .exposition import _bake_output
 from .registry import REGISTRY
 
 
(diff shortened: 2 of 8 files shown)


brian-brazil:
I'd prefer not to have to maintain yet another copy of the exposition http handling for a subtly difference interface. Is there not some interface all of these can share?

Skeen:
I understand your hesitance to adapt yet another interface. The README file is already quite crowded.

As for whether there is a common interface all of these can share, I wanna say yes, there is the PEP-3333 standard WSGI and it's spiritual successor ASGI.
- Supporting just these two should be enough, as any web-framework should be able to integrate with one or the other, depending on whether it's a synchronous or asynchronous framework.

My last though is that WSGI and ASGI should not be considered equal to the frameworks. As they are gateway interfaces and thus more general than specific frameworks.

Skeen:
Note: I have closed the PR for the Quart extension of the README file, I included it solely for completeness to have a flask-equivalent ASGI framework example.

brian-brazil:
It doesn't matter to me why this exists, it matters that we're already duplicating this code 3 times for various things that were popular at various times.

Skeen:
How would you feel about cleaning up, by deprecating the twisted code, removed flask and python http documentation and going solely with the code+examples for the gateway interfaces?

The gateway interfaces produces production quality endpoints in standard defined ways, and the integrations are simple in most languages, either via [routers](https://channels.readthedocs.io/en/latest/topics/routing.html#urlrouter), [middleware](https://pgjones.gitlab.io/hypercorn/dispatch_apps.html) or [adapters](https://twisted.readthedocs.io/en/twisted-15.1.0/web/howto/web-in-60/wsgi.html).

For [Flask](https://flask.palletsprojects.com/en/1.1.x/patterns/appdispatch/#combining-applications)/[Django](https://docs.djangoproject.com/en/3.0/howto/deployment/wsgi/#applying-wsgi-middleware) it's common practice to utilize [WSGI middlewares](https://www.python.org/dev/peps/pep-0333/#middleware-components-that-play-both-sides), for [Quart](https://pgjones.gitlab.io/quart/middleware.html)/[Django](https://docs.djangoproject.com/en/3.0/howto/deployment/asgi/#applying-asgi-middleware) it's common practice to utliize [ASGI middleware](https://channels.readthedocs.io/en/latest/asgi.html#composability). Twisted has a [WSGIResource](https://twistedmatrix.com/documents/10.1.0/api/twisted.web.wsgi.WSGIResource.html) which can call any WSGI app, thus I don't see the need for the [MetricsResource](https://github.com/prometheus/client_python/blob/master/prometheus_client/twisted/_exposition.py#L8) for [twisted](https://twistedmatrix.com/trac/).

brian-brazil:
Is it possible to get twisted to use one of the *sgi apis? That way it could be replaced by docs.

Skeen:
Yea, that is very much possible.

To use the `make_wsgi_app` as a twisted resource, simply replace `MetricsResource()`, with: `WSGIResource(reactor, reactor.getThreadPool(), make_wsgi_app())`.
Or use [txasgiresource](https://github.com/JohnDoee/txasgiresource) to do the same thing with the `make_asgi_app`, by replacing `MetricsResource()` with `ASGIResource(make_asgi_app())`.

The example in the README example becomes:
```
from prometheus_client import make_asgi_app
from txasgiresource import ASGIResource
from twisted.web.server import Site
from twisted.web.resource import Resource
from twisted.internet import reactor

root = Resource()
root.putChild(b'metrics', ASGIResource(make_asgi_app())

factory = Site(root)
reactor.listenTCP(8000, factory)
reactor.run()
```
And all the framework specific code for twisted can be deprecated/removed.

brian-brazil:
If we can do that then and avoid duplicating the parameter handing code, that sounds good.

Skeen:
I will do it when I can get around to it.

How do you feel about the [MetricsHandler](https://github.com/prometheus/client_python/blob/master/prometheus_client/exposition.py#L141), and the [start_http_server + _ThreadingSimpleServer](https://github.com/prometheus/client_python/blob/master/prometheus_client/exposition.py#L190)?

We could eliminate `_ThreadingSimpleServer` by making `start_http_server` an alias of `start_wsgi_server`. If we did that we could additionally deprecate `MetricsHandler`, and potentially rewrite it to wrap and call the WSGI app from `make_wsgi_app`? - How do you feel about it this?

brian-brazil:
MetricsHandler is used by users, so it should continue to work. I'm okay with switching around how start_http_server works as long as it can still handle concurrent scrapes.

Skeen:
What do you think about this?

README.md
```diff
@@ -306,6 +306,32 @@ from prometheus_client import start_wsgi_server
...
+
+```python
+from prometheus_client import make_asgi_app
+
+app = make_asgi_app()
+```
+Such an application can be useful when integrating Prometheus metrics with ASGI
+apps.
+
+The app can be used to serve the metrics through an ASGI implementation, such
```
brian-brazil: This seems unnecessary. These are the client_python docs, not an ASGI intro.

prometheus_client/asgi.py
```diff
@@ -0,0 +1,34 @@
+from urllib.parse import parse_qs
+
+from .exposition import _bake_output
+from .registry import REGISTRY
+
+
+def make_asgi_app(registry=REGISTRY):
+    """Create a ASGI app which serves the metrics from a registry."""
+
+    async def prometheus_app(scope, receive, send):
```
brian-brazil: Will having these new keywords in the codebase cause problems for those still on Python 2?

tox.ini
```diff
@@ -30,6 +32,7 @@ deps =
 deps =
     {[base]deps}
     {py27,py37,pypy,pypy3}: twisted
+    {py35,py36,py37,py38,pypy3}: asgiref
```
brian-brazil: We only test the latest Python 3, which is 3.7 here.

tests/test_wsgi.py
```diff
@@ -0,0 +1,64 @@
...
+            super().assertIn(item, iterable)
+        except:  # Python < 2.7
+            self.assertTrue(
+                item in iterable,
+                msg="{item} not found in {iterable}".format(
+                    item=item, iterable=iterable
+                )
+            )
+
+    @parameterized.expand([
```
brian-brazil: Please keep this explicit, rather than adding a dependency.

prometheus_client/exposition.py
```diff
@@ -143,18 +161,15 @@ class MetricsHandler(BaseHTTPRequestHandler):
...
-        except:
-            self.send_error(500, 'error generating metric output')
-            raise
-        self.send_response(200)
-        self.send_header('Content-Type', content_type)
+        # Bake output
+        status, header, output = _bake_output(registry, accept_header, params)
+        # Return output
+        self.send_response(int(status.split(' ')[0]))
+        self.send_header('Content-Type', header[1])
```
brian-brazil: It seems odd that we're using only the 2nd item in header.

README.md
Skeen: I simplified the documentation.

prometheus_client/asgi.py
Skeen: No, it should not cause any issues.

The `asgi.py` file is conditionally included in `exposition.py`, and hence will only be succesfully loaded in python3.6 and higher.
This was the primary argument for introducing this file, rather than just including this code directly in `exposition.py`.

tox.ini
Skeen: I'm not sure I understand.

It does seem like Travis runs through several Python 3 versions here: https://travis-ci.org/prometheus/client_python/builds/652096731 ?

prometheus_client/exposition.py
Skeen: I agree, I cleaned it up. It now calls by `self.send_header(*header)`.

tests/test_wsgi.py
Skeen: Done

tox.ini
brian-brazil: Not for things that add dependencies, we only test them on the latest.

Why does this pull in a dependency?

Skeen: The dependency is to get `asgiref.testing`; https://github.com/django/asgiref/blob/master/asgiref/testing.py

It's the reference support library for assisting in testing asyncronous applications within syncronous test-frameworks.

brian-brazil: You still need to trim down the line here.

Skeen: Done

brian-brazil:
Thanks!

pquentin:
@Skeen Thanks for this! It works really well, and I was apple to add it to a FastAPI app like this:

```
from fastapi import FastAPI
import prometheus_client

app = FastAPI()
metrics_app = prometheus_client.make_asgi_app()
app.mount("/metrics", metrics_app)
```

@brian-brazil Can you please consider releasing a new version to PyPI? Thanks!

denisSurkov:
Hi! 

ASGI is really cool feature. Especially for fastapi app (see comment above). When are you going to release asgi version to PyPI?

whg517:
I can't wait to use Prometheus in fastapi. But I noticed that prometheus_client has not released a new version. So I chose to [starlette_exporter](https://github.com/stephenhillier/starlette_exporter). At least for now, there are no functional issues. You can try it, too.
@denisSurkov @pquentin

pquentin:
@brian-brazil Thank you for the release! 🎉

aLowMagic:
> @Skeen Thanks for this! It works really well, and I was able to add it to a FastAPI app like this:
> 
> ```
> from fastapi import FastAPI
> import prometheus_client
> 
> app = FastAPI()
> metrics_app = prometheus_client.make_asgi_app()
> app.mount("/metrics", metrics_app)
> ```
> 
> @brian-brazil Can you please consider releasing a new version to PyPI? Thanks!

could you please remain me that, how can i add a count metrics in the asgi mode? Thanks alot

Skeen:
> could you please remain me that, how can i add a count metrics in the asgi mode? Thanks alot

The same way you'd do it for non-asgi:
```
c = Counter('my_failures', 'Description of counter')
c.inc()     # Increment by 1
c.inc(1.6)  # Increment by given value
```

aLowMagic:
> > could you please remain me that, how can i add a count metrics in the asgi mode? Thanks alot
> 
> The same way you'd do it for non-asgi:
> 
> ```
> from prometheus_client import Counter
> c = Counter('my_failures', 'Description of counter')
> c.inc()     # Increment by 1
> c.inc(1.6)  # Increment by given value
> ```

Thanks a lot!

aLowMagic:
> > could you please remain me that, how can i add a count metrics in the asgi mode? Thanks alot
> 

> The same way you'd do it for non-asgi:
> 
> ```
> from prometheus_client import Counter
> c = Counter('my_failures', 'Description of counter')
> c.inc()     # Increment by 1
> c.inc(1.6)  # Increment by given value
> ```

Hello mogul, it's me again~
I found that, in the multiprocess mode, if two fastapi(asgi) workers exists, their are two counter exists, and the data will like:
prome_pic

and then i need calculate the qps like:
```
sum(increase(context_app_calls_total{app_name=~".*", endpoint="/test", instance="0.0.0.0:8000", job="8000_metrics", method="get", status_code="200"}[1m])/60)
```
will i get correct result like this way? Thanks a lot

Skeen:
Hi @aLowMagic 

I don't think this PR is the right place to ask for help like this, or to have these discussions, as it's unrelated to ASGI itself.
The challenge your facing is a more generic one of combining multiple parallel data-series, irrelevant to how the data series are collected, or which programming language is used to do that collection.

I would suggest you have a look at: https://prometheus.io/community/ - And seek help there :)

aLowMagic:
> Hi @aLowMagic
> 
> I don't think this PR is the right place to ask for help like this, or to have these discussions, as it's unrelated to ASGI itself.
> The challenge your facing is a more generic one of combining multiple parallel data-series, irrelevant to how the data series are collected, or which programming language is used to do that collection.
> 
> I would suggest you have a look at: https://prometheus.io/community/ - And seek help there :)

Thanks a lot~

Issue #509, opened 2020-02-10
Allow customizing the suffix function in multiprocess mode (instead of os.getpid())


xavfernandez:
One frequent issue with using `prometheus_client` in multiprocess mode is that you can end up with a lot of files which can severely impact the collecting performance ending with scraping timeouts in the worth cases.

This comes from the fact that to prevent concurrent usage, `prometheus_client` uses `os.getpid()` to name its metrics file and ensure that each process only touches its own files: it's a perfectly good and simple design.
However for long running processes spawning new processes every few minutes you can end up with several thousands files after a few months.

But in some cases, you have access to other methods to ensure that each process is uniquely identified (and ensure that each process only touches uniquely identified metrics files).

**Would it be possible to allow customizing (& document as public API) the function generating metrics file suffix ?
(Of course, this would still be an use-at-your-own-risk escape hatch).**

I'm speaking of:
https://github.com/prometheus/client_python/blob/e58301600c47d5c93985b83490bc57903af2918e/prometheus_client/values.py#L31

For instance, using uWSGI, we might want to use `uwsgi.worker_id()`:
```
from prometheus_client import values
import uwsgi

values.ValueClass = values.MultiProcessValue(uwsgi.worker_id)
```

Or when each worker has its own task and you can ensure that only one worker at a time is dealing with this task (thanks to a postgres advisory lock in my case).

```
def prometheus_pid_func():
    return os.getenv("PROMETHEUS_WORKER_TASK_ID", str(os.getpid()))


from prometheus_client import values
values.ValueClass = values.MultiProcessValue(prometheus_pid_func)
```

brian-brazil:
The problem is that that isn't safe if the process then later forks, it took a while to get the current logic right and I'm not sure how correctly it'll work with anything that's not actually a pid.

xavfernandez:
> The problem is that that isn't safe if the process then later forks, it took a while to get the current logic right and I'm not sure how correctly it'll work with anything that's not actually a pid.

Yes, hence the "use-at-your-own-risk escape hatch".
This would be clearly documented as a "I know what I'm doing" hook but I think that in most cases the user has a fairly good understanding whether its code will spawn a subprocess or not (typically web server workers don't).

brian-brazil:
I don't think this is something that should go in the readme, we already have issues with users failing to follow the docs that are there and this would be handing them a chainsaw to juggle. I am open to code changes that'd make this easier to do for those who know what they're up to though.

[...]

PR #514, opened 2020-02-15
values: add a docstring to MultiProcessValue

xavfernandez:
To document the method as semi-supported.

Adresses #509

### DIFF
--- a/prometheus_client/values.py
+++ b/prometheus_client/values.py
@@ -28,10 +28,17 @@ class MutexValue(object):
             return self._value
 
 
-def MultiProcessValue(_pidFunc=os.getpid):
+def MultiProcessValue(process_identifier=os.getpid):
+    """Returns a MmapedValue class based on a process_identifier function.
+
+    The 'process_identifier' function MUST comply with this simple rule:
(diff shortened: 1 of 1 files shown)


prometheus_client/values.py
```diff
@@ -28,10 +28,17 @@ def get(self):
             return self._value
 
 
-def MultiProcessValue(_pidFunc=os.getpid):
+def MultiProcessValue(process_identifier=os.getpid):
+    """Returns a MmapedValue class based on a process_identifier function.
+
+    The 'process_identifier' function MUST comply with this simple rule:
```
brian-brazil: Don't mix must and should.

xavfernandez: Updated

brian-brazil:
Thanks!

tyler-8:
By the way, this contained a breaking change in some cases as it wasn't just a docstring update technically. `process_identifier` used to be named `_pidFunc` so for anyone passing that as an explicit kwarg will run in to errors.

brian-brazil:
The Python convention is that names beginning with an underscore are internal, thus changing or removing them is not breaking.

Issue #443, opened 2019-07-26
Aggregate expired pid db files, control the number of files and improve scrape effectiveness


fangjianfeng:
history related issues:
https://github.com/prometheus/client_python/pull/441
https://github.com/prometheus/client_python/pull/430

Can we aggregate all the db files from a period of time ago and non-current pid into a total db file, to control the number of pid files ？

I have realized this idea with golang，here are some details：

Project deploy info：
gunicorn django
128 workers
gunicorn max_requests：10000（create a new pid file almost every minute）

1. I can't solve the problem that the pid file has been growing, and it can reach 6,000 in four days；
2. Try to delete the expired pid regularly in the code, but it will cause the figure to drop with grafana；
3. The time to request metric is getting longer as the program runs.

Improve scrape efficiency:
I used golang to rewrite the logic of python aggregate metrics（generate metric still using python）. After rewriting, each scrape time is less than 1 second. 

Solve the growing pid files:
Aggregate all the db files from a period of time ago and non-current pid into a total db file. Then delete these files. When calculating metric, history total db + curent pid = current pid db. (i do it every hour)

Now, num of pid files is <200 in my project. if we can do change this, it would be a big strengthen. Just like prometheus will also aggregate historical data

fangjianfeng:
I just raised this idea to see if it is necessary. and sorry，i used Golang to rewrite aggregate metric, just because I was trying to improve the scrape time.

Issue #204, opened 2017-10-13
Poor performance in MultiProcessCollector with frequently changing PIDs


manics:
I'm using [django-prometheus](https://github.com/korfuri/django-prometheus) in multiprocess mode, and I've noticed the time to fetch metrics increases the longer the server has been running. Currently I've got 5414 `*.db` files in `prometheus_multiproc_dir`.

[`MultiProcessCollector.collect`](https://github.com/prometheus/client_python/blob/v0.0.21/prometheus_client/multiprocess.py#L21) reads all db files which I suspect is the bottleneck.

[`mark_process_dead`](https://github.com/prometheus/client_python/blob/v0.0.21/prometheus_client/multiprocess.py#L89) only removes gauge files.

Do you think it'd be feasible to remove all files by copying the contents of `type_{pid}.db` files into a single `type_old.db` file? Or is this something that should be solved in django-prometheus?

brian-brazil:
It's not safe to compact, as there's no locking and a pid might be reused.

How long is the scrape taking?

manics:
django-prometheus was taking around 10s causing the scrape to timeout. There's a total of 55MB .db files in prometheus_multiproc_dir.

I've looked at further into this and can see a few options:
- Increase `scrape_timeout`
- Change our gunicorn `max_requests` to prevent workers turning over so quickly

brian-brazil:
You're going to have to either burn cpu somewhere on locking, or reduce the churn. When I designed this I only had maybe 100 pids in mind. Increasing the interval+timeout and reducing the churn are both options.

manics:
Thanks. If I've understood the code correctly the db files are re-read on each scrape, and would grow without bound assuming the PIDs don't change. Is there a maximum number of metrics or maximum db size beyond which you'd advise restarting the process?

brian-brazil:
This is a relatively new feature, so there's no specific guidance yet. I'd say balance the disruption of a restart against the cost of handling more data.

groner:
I'm running into this issue with metrics I started collecting last week.  Our average scrape time been going up a couple seconds every day.

I'm interested in the possibility of adding a compacting operation to mark_process_dead, if you're open to discussing it @brian-brazil.

svanscho:
We are running into the same issue - is there a fix already?

PR #518, opened 2020-02-28
merge history db file into a common db file to avoid db file booming

lee3164:


brian-brazil:
I don't believe this is safe, nothing stops something else with that pid coming along while this is running. In addition there's nothing stopping this being run concurrently.

lee3164:
> I don't believe this is safe, nothing stops something else with that pid coming along while this is running. In addition there's nothing stopping this being run concurrently.

1. I think this method should be called when a process is atexit, so, nobody should called it at runtime.
for uwsgi and gunicorn, they both have a hook to run method at worker exit
2. for safe, if you insist on it, we can use a file lock, but I think process exit is a low probability  event.

lee3164:
#443 this issue has the same problem with me. I used multi process mode a few days and I found the endpoint for scrape  latency was becoming more and more slow, so I checked the prometheus_multiproc_dir, and found there were 1000+ db files, because sometimes the worker would exit, so the number of db files increased dramatically

brian-brazil:
We can't make any assumptions about how this code is used, and given it's in a multi-process environment it must avoid any race conditions.

How would this work for gauges?

lee3164:
> We can't make any assumptions about how this code is used, and given it's in a multi-process environment it must avoid any race conditions.
> 
> How would this work for gauges?

As you can see in the code, we use the same method to merge metrics as they are being scraped. I have use this commit in my production environment,  and it works well, the latency of scrape endpoint has been reduced significantly. We can use file lock to avoid race condition.

lee3164:
We should point out that this code should only be used in the multiprocess mode

lee3164:
I have added process lock code which is from django.core.files.locks

prometheus_client/multiprocess.py
```diff
@@ -157,3 +176,43 @@ def mark_process_dead(pid, path=None):
...
+        merge_file = os.path.join(path, '{0}_merge.db'.format(file_prefix))
+        if merge_file not in merge_files:
+            merge_files.append(merge_file)
+
+        # if not exist merge_file, create and init it
+        if not os.path.exists(merge_file):
+            MmapedDict(merge_file).close()
+
+    # do merge, here we use the same method to merge
+    metrics = MultiProcessCollector.merge(files + merge_files, accumulate=False)
```
lee3164: @brian-brazil  here I use the same method to merge metrics

brian-brazil:
What happens if a process exists, sets a gauge, exits, and a new process comes along with the same pid?

lee3164:
> What happens if a process exists, sets a gauge, exits, and a new process comes along with the same pid?
```python
            for s in metric.samples:
                name, labels, value, timestamp, exemplar = s
                if metric.type == 'gauge':
                    without_pid_key = (name, tuple([l for l in labels if l[0] != 'pid']))
                    if metric._multiprocess_mode == 'min':
                        current = samples_setdefault(without_pid_key, value)
                        if value < current:
                            samples[without_pid_key] = value
                    elif metric._multiprocess_mode == 'max':
                        current = samples_setdefault(without_pid_key, value)
                        if value > current:
                            samples[without_pid_key] = value
                    elif metric._multiprocess_mode == 'livesum':
                        samples[without_pid_key] += value
                    else:  # all/liveall
                        samples[(name, labels)] = value
```
1. We have 5 mode for Gauge in multi process mode, all, min, max, liveall, livesum. In livesum and liveall mode,  we will delete the died process's db file, in min and max mode,  we will compare the old and new value and chose the correct value, in all mode, if this event happend, no matter old code or my commit, the new value will cover the old value. 
2. Process exit and get the same pid is a low probability event.

brian-brazil:
> Process exit and get the same pid is a low probability event.

It's a certainty though at scale.

lee3164:
> > Process exit and get the same pid is a low probability event.
> 
> It's a certainty though at scale.

Yes，Gauge is really a problem in multi process mode, developer should think it carefully. Do you have a good idea to reduce the latency in multiprocess mode when the nums of db files increase?

brian-brazil:
Currently the main approach is avoiding creating lots of files in the first place.

PierreF:
I'm also hitting this issue (uWSGI restarted often, resulting in lots of files which cause /metrics endpoints to become too slow). I'm interested and willing to help to see this PR being merged.

With the exception of test that fail due to:
* nested lock acquisition - mark_process_dead and merge took the lock
* and labels pid=merge added on gauge in *addition* to pid=$original_pid. So we could not add this duplicated pid labels or ensure it's added "before".

What block this PR ?
* locking seems correct. Merge is only performed on own pid files (no pid reuse issue) and with the lock file
* gauges are merged as they already are. We just reads less files.
   * On gauges, this PR indeed don't fix the fact that number of gauge metrics will increase... but it's not a regression.

brian-brazil:
> locking seems correct. Merge is only performed on own pid files (no pid reuse issue)

That's not the case, mark_process_dead runs in a different process.

lee3164:
I' m using this feature in my production env now and there is no problem at present

lee3164:
> > locking seems correct. Merge is only performed on own pid files (no pid reuse issue)
> 
> That's not the case, mark_process_dead runs in a different process.

mark_process_dead runs when process is going to exit, and it only do its own merge work, using a lock will make this progress serialized.

PierreF:
In uWSGI case, mark_process_dead will probably run at process exit (so with own PID). At least it's own I did it (using Python atexit handler). But in Gunicorn, mark_process_dead will be called by master process for another PID.

This indeed open a (small) race-condition: if mark_process_dead is run with the PID of a terminated worker AND a new worker start using the same PID. The lock won't help there, since the new worker kept an open file descriptor to the file that mark_process_dead will remove.

This should however be quite rare as it require the OS to re-assign the PID of a recently terminated process. But nothing forbid this to happen.

PierreF:
BTW, this race-condition is now new. Gauge livesum and liveall already had this issue (new worker open the file, then mark_process_dead run by master remove them). Or did I miss something that avoid this race condition which could be reused ?

lee3164:
I think we can use pid+process_create_time to name the file, this could avoid 2 process with same pid to have same file name

brian-brazil:
> I think we can use pid+process_create_time to name the file, this could avoid 2 process with same pid to have same file name

There's nothing stopping two processes being created and terminated in the same second with the same PID.

> (new worker open the file, then mark_process_dead run by master remove them).

True, we should look at fixing that.

lee3164:
image


> There's nothing stopping two processes being created and terminated in the same second with the same PID.
> 
> >

yes, but I think this is a very very very low probability event. In psutil, I also find a comments related to this problem.

PierreF:
In psutil, the create_time has more precision that in our case (it's a float number of seconds). But I agree that this event look like a very low probability (and in my opinion it's acceptable. maybe the merge should be an opt-in option ?)

The best solution I can think of that don't have "low probability" error is to add a new method that must be called by worker itself at exit, not by the master. It will merge files. User may or not register this method if the merge is wanted or not.

brian-brazil:
> to add a new method that must be called by worker itself at exit, not by the master.

I doubt that our users will have that level of control. This needs to work with all forms of multiprocess setups, including gunicorn and multiprocessing - and handling things like a segfault of a child correctly.

lee3164:
> In psutil, the create_time has more precision that in our case (it's a float number of seconds). But I agree that this event look like a very low probability (and in my opinion it's acceptable. maybe the merge should be an opt-in option ?)
> 
> The best solution I can think of that don't have "low probability" error is to add a new method that must be called by worker itself at exit, not by the master. It will merge files. User may or not register this method if the merge is wanted or not.

why gunicorn worker cannot register this method to ateixt?

brian-brazil:
Atexit is not reliably generally, plus things might happen before it got registered.

indrif:
I still dont understand how to handle the growing number of db files. When is it safe to remove them? The documentation states that the files should be cleared on every deploy which suggests that they can be removed without any special hesitation unless a process is currently writing to them. So what happens if I remove files? Do I risk losing data? Then how come it works to purge the files on a hard restart? Wouldnt an easy solution be to just have some kind of cleanup cron that just removes old files?

brian-brazil:
The documentation says that you should delete them when the process restarts, there's nothing more to it. Anything else is unsafe.

This issue is also not the place for usage questions, please take those to the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users).

prometheus_client/multiprocess.py
mhumpula: I believe this will cause deadlock: `mark_process_dead` is already holding lock and calling `merge` will try to acquire the same lock again.

mhumpula (commented):
`def collect(self)` should be guarded by lock too. The list and subsequent merge can have different view on directory content otherwise.

vearutop:
The issue from 2017 https://github.com/prometheus/client_python/issues/204 is still not resolved, suggesting this client library is designed to serve metrics with multi-second latency.

I appreciate @brian-brazil is so careful about unlikely race conditions that may happen, but it means this library is in no way production-ready.

prometheus_client/multiprocess.py
chawco: I don't believe it will -- it looks like lock acquisition via `flock` is reentrant safe, but unlocking via `flock` _isn't_, so I think this will prematurely release the lock after the `merge` call here, at least on POSIX systems. I don't know what the locking behaviour on NT will be.

stefanor:
> Have you guys looked into a shared database as storage? For example, local redis or memcached?

Implemented shared Redis in #1170

Issue #490, opened 2019-12-02
Multiprocessing collector doesn't work


Sayed-Soroush-Hashemi:
### version
prometheus_client==0.7.1

### How to reproduce
```python
import json
import os
import time
from multiprocessing import Pool
from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor 

from prometheus_client import Counter, CollectorRegistry, multiprocess, start_http_server, generate_latest

os.environ['prometheus_multiproc_dir'] = './tmp_test' 

registry = CollectorRegistry()
multiprocess.MultiProcessCollector(registry)

counter_1 = Counter('name_counter_1', '', ['l1', 'l2'], registry=registry)

def f(i):
	while True:
		counter_1.labels(f'l1_val_{i}').inc()
		print(f'process {i}')
		print(generate_latest(registry).decode('utf-8'))
		time.sleep(i+1)

def test_multiprocessing_pool():
	with Pool(2) as p:
		p.map(f, [1, 2])

def test_process_pool_executor():
	with ProcessPoolExecutor(2) as p:
		p.map(f, [1, 2])

def test_thread_pool_executor():
	with ThreadPoolExecutor(2) as p:
		p.map(f, [1, 2])


if __name__ == '__main__':
	# test_multiprocessing_pool()
	test_process_pool_executor()
	# test_thread_pool_executor()

```
### what I see
Each process collects only its own metrics when using `ProcessPollExecuter` and `multiprocessing.Pool`, but it works correctly when using `ThreadPoolExecutor`.
```text
process 1
# HELP name_counter_1_total 
# TYPE name_counter_1_total counter
name_counter_1_total{l="l_val_1"} 2.0
# TYPE name_counter_1_created gauge
name_counter_1_created{l="l_val_1"} 1.575303633967081e+09

process 2
# HELP name_counter_1_total 
# TYPE name_counter_1_total counter
name_counter_1_total{l="l_val_2"} 2.0
# TYPE name_counter_1_created gauge
name_counter_1_created{l="l_val_2"} 1.5753036339671488e+09
```
### what I expect
Each process should be able to collect all the metrics in all the processes.

brian-brazil:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

[...]

Issue #382, opened 2019-03-13
Question: how to instrument a multiprocessing program?


moio:
Specifically: I am trying to instrument a program that uses `multiprocessing` directly, without Gunicorn or any other framework/server.

What are the basic assumptions? What I am currently doing (in the parent process) is:

```python
prometheus_dir = "/tmp/my-dir"
shutil.rmtree(prometheus_dir, ignore_errors=True)
os.mkdir(prometheus_dir)
os.environ["prometheus_multiproc_dir"] = prometheus_dir

registry = CollectorRegistry()
multiprocess.MultiProcessCollector(registry)
start_http_server(8004, registry=registry)
```

Then code I want to instrument lives in child processes, where I am using:

```python
REQUEST_TIME = Summary('request_processing_seconds', 'Time spent in request')

[...]

@REQUEST_TIME.time()
@tornado.gen.coroutine
def method(self, load):
  [...]
```

Expected behavior: `request_processing_seconds_count` and `request_processing_seconds_sum` grow as `method` gets called.

Actual behavior: `request_processing_seconds_count` and `request_processing_seconds_sum` stay at 0.0.

I am surely doing something wrong, but I could not figure out what from documentation or internal code so far - if anybody could explain that would be very very nice.

Thanks in advance!

brian-brazil:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

moio:
Thanks, sorry.

PR #524, opened 2020-03-03
Improve multiprocessing documentation

Sinclert:
The current multiprocessing documentation is short, and can lead to some confusion.

This PR addresses the following aspects:

- Setting up `prometheus_multiproc_dir` env. var but still fail to collect metrics, usually due to a wrong setting of the variable (see issues https://github.com/prometheus/client_python/issues/490 &  https://github.com/prometheus/client_python/issues/382).

- Separating _mandatory steps_ (1 and 2) from _optional steps_ (3), depending on the usage of `Gauge`. This is because the `mark_process_dead` function does nothing related to `Count`, `Summary` or `Histogram` metrics, as seen in [its definition](https://github.com/prometheus/client_python/blob/master/prometheus_client/multiprocess.py#L152).

cc @brian-brazil

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,13 +495,13 @@ The `prometheus_multiproc_dir` environment variable must be set to a directory
 that the client library can use for metrics. This directory must be wiped
 between Gunicorn runs (before startup is recommended).
 
-This environment variable must be set **from a start-up shell script**,
-and not directly from Python (otherwise it will not propagate to child processes).
+This environment variable should be set from a start-up shell script,
+and not directly from Python (otherwise it may not propagate to child processes).
 
 **2. Metrics collector**:
 
-The application must initialize a new metrics `CollectorRegistry`,
-and save it within the multi-process collector.
+The application must initialize a new `CollectorRegistry`,
+and store the multi-process collector inside.
 
 ```python
 from prometheus_client import multiprocess
@@ -521,10 +521,9 @@ def app(environ, start_response):
     return iter([data])
 ```
 
-**3. Metrics mode and configuration (Gauge)**:
+**3. Gunicorn configuration**:
 
-When `Gauge` metrics are used, additional tuning needs to be performed.
-First of all, the `gunicorn` configuration file need to include the following function:
+The `gunicorn` configuration file needs to include the following function:
 
 ```python
 from prometheus_client import multiprocess
@@ -533,8 +532,10 @@ def child_exit(server, worker):
     multiprocess.mark_process_dead(worker.pid)
 ```
 
-Second of, metrics instrumentation. Gauges have several modes they can run in,
-which can be selected with the `multiprocess_mode` parameter.
+**4. Metrics tuning (Gauge)**:
+
+When `Gauge` metrics are used, additional tuning needs to be performed.
+Gauges have several modes they can run in, which can be selected with the `multiprocess_mode` parameter.
 
 - 'all': Default. Return a timeseries per process alive or dead.
 - 'liveall': Return a timeseries per process that is still alive.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -489,21 +489,60 @@ This comes with a number of limitations:
...
-**One**: Gunicorn deployment
+**1. Gunicorn deployment**:
 
 The `prometheus_multiproc_dir` environment variable must be set to a directory
 that the client library can use for metrics. This directory must be wiped
 between Gunicorn runs (before startup is recommended).
 
-Put the following in the config file:
+This environment variable must be set **from a start-up shell script**,
+and not directly from Python (otherwise it will not propagate to child processes).
```
brian-brazil: may not

It is possible to do it right, but unlikely

README.md
```diff
@@ -489,21 +489,60 @@ This comes with a number of limitations:
...
 
-**One**: Gunicorn deployment
+**1. Gunicorn deployment**:
 
 The `prometheus_multiproc_dir` environment variable must be set to a directory
 that the client library can use for metrics. This directory must be wiped
 between Gunicorn runs (before startup is recommended).
 
-Put the following in the config file:
+This environment variable must be set **from a start-up shell script**,
```
brian-brazil: should

No need for the bold

README.md
```diff
@@ -489,21 +489,60 @@ This comes with a number of limitations:
...
+and not directly from Python (otherwise it will not propagate to child processes).
+
+**2. Metrics collector**:
+
+The application must initialize a new metrics `CollectorRegistry`,
+and save it within the multi-process collector.
```
brian-brazil: Object ownership is the other way around

README.md
```diff
@@ -489,21 +489,60 @@ This comes with a number of limitations:
...
 that the client library can use for metrics. This directory must be wiped
 between Gunicorn runs (before startup is recommended).
 
-Put the following in the config file:
+This environment variable must be set **from a start-up shell script**,
+and not directly from Python (otherwise it will not propagate to child processes).
+
+**2. Metrics collector**:
+
+The application must initialize a new metrics `CollectorRegistry`,
```
brian-brazil: s/metrics//

README.md
```diff
@@ -489,21 +489,60 @@ This comes with a number of limitations:
...
+Second of, metrics instrumentation. Gauges have several modes they can run in,
+which can be selected with the `multiprocess_mode` parameter.
+
+- 'all': Default. Return a timeseries per process alive or dead.
+- 'liveall': Return a timeseries per process that is still alive.
+- 'livesum': Return a single timeseries that is the sum of the values of alive processes.
+- 'max': Return a single timeseries that is the maximum of the values of all processes, alive or dead.
+- 'min': Return a single timeseries that is the minimum of the values of all processes, alive or dead.
+
+Complete example:
```
brian-brazil: This is now duplicating the above.

Sinclert: I thought that having the complete example would make sense.

I will change it if you think it is not necessary.

Sinclert:
Suggestions applied ✅

README.md
```diff
@@ -527,19 +521,35 @@ def app(environ, start_response):
     return iter([data])
 ```
 
-**Three**: Instrumentation
+**3. Metrics mode and configuration (Gauge)**:
 
-Counters, Summarys and Histograms work as normal.
+When `Gauge` metrics are used, additional tuning needs to be performed.
+First of all, the `gunicorn` configuration file need to include the following function:
```
brian-brazil: Everyone should do this, not just those using gauges.

Sinclert: What benefit does it provide to non `Gauge` metrics, if [its implementation](https://github.com/prometheus/client_python/blob/master/prometheus_client/multiprocess.py#L152) has nothing to do with them?

brian-brazil: The current implementation doesn't, that doesn't mean it won't in future. In general exposition should be independent of instrumentation anyway, so you should do it in case a dependency uses gauges.

Sinclert: I disagree with the first point:

If the implementation changes, affecting other metric types, the `README.md` should reflect that change.

Defining the `mark_process_dead` function just because _"who knows, right? maybe it changes in the future..."_ does not seem like a strong point to me. That same argument could be applied to almost the entire codebase as _"any function may change in the future... right?"_

---
However, I see your second point: a dependency _may_ use a Gauge.

Sinclert:
Suggestions v2 applied ✅ 

Let me know if there is anything else you want to change. cc/ @brian-brazil

README.md
```diff
@@ -527,19 +521,36 @@ def app(environ, start_response):
     return iter([data])
 ```
 
-**Three**: Instrumentation
+**3. Gunicorn configuration**:
+
+The `gunicorn` configuration file need to include the following function:
```
brian-brazil: needs

brian-brazil:
Thanks!

Issue #527, opened 2020-03-12
Eagerly convert value to float in *MetricFamily.add_metric()


mhansen:
Hi there, an experience report and improvement suggestion.

I have a [pretty simple exporter](https://github.com/mhansen/bom_exporter/blob/master/bom_exporter.py) which queries a backend URL for JSON, grabs the JSON, stuffs it into prometheus metric families, which it then yields.

Sometimes it errors with stacktraces only when the request is finishing, with a stack traces that points at `finish_request` rather than the code that added the bad float:

```
TypeError: ("float() argument must be a string or a number, not 'NoneType'", Metric(bom_wind_speed, Wind speed (km/h) from the Bureau of Meterology, gauge, , [Sample(name='bom_wind_speed', labels={'location': 'Sydney Airport'}, value=None, timestamp=None, exemplar=None), Sample(name='bom_wind_speed', labels={'location': 'Sydney - Observatory Hill'}, value=20, timestamp=None, exemplar=None)]))
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/socketserver.py", line 625, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/local/lib/python3.5/socketserver.py", line 354, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/local/lib/python3.5/socketserver.py", line 681, in __init__
    self.handle()
  File "/usr/local/lib/python3.5/http/server.py", line 422, in handle
    self.handle_one_request()
  File "/usr/local/lib/python3.5/http/server.py", line 410, in handle_one_request
    method()
  File "/usr/local/lib/python3.5/site-packages/prometheus_client/exposition.py", line 152, in do_GET
    output = encoder(registry)
  File "/usr/local/lib/python3.5/site-packages/prometheus_client/openmetrics/exposition.py", line 56, in generate_latest
    floatToGoString(s.value),
  File "/usr/local/lib/python3.5/site-packages/prometheus_client/utils.py", line 8, in floatToGoString
    d = float(d)
TypeError: ("float() argument must be a string or a number, not 'NoneType'", Metric(bom_wind_speed, Wind speed (km/h) from the Bureau of Meterology, gauge, , [Sample(name='bom_wind_speed', labels={'location': 'Sydney Airport'}, value=None, timestamp=None, exemplar=None), Sample(name='bom_wind_speed', labels={'location': 'Sydney - Observatory Hill'}, value=20, timestamp=None, exemplar=None)]))
[...]

PR #530, opened 2020-03-19
Eagerly verify Sample values are float-convertible

mhansen:
Addresses #527.

There's a test setting GaugeHistogramMetricFamily.gsum_value = None, so
I've preserved that behaviour, by not appending and crashing if gsum_value
is None. That meant that this test case "([('spam', 0), ('eggs', 0)], None, TypeError)" doesn't throw any more, so I've deleted the test case.

I was a bit unsure about this bit:
        assert isinstance(exception.args[-1], core.Metric)

I'm not sure what exception.args[-1] is, the python docs for TypeError and ValueError
don't explain. I've removed the assertion.

Not sure if removing these assertions is a great idea - would love feedback on this.


prometheus_client/samples.py
```diff
@@ -32,12 +32,15 @@ def __gt__(self, other):
...
-# Value can be an int or a float.
+# Value must be a float.
 # Timestamp can be a float containing a unixtime in seconds,
 # a Timestamp object, or None.
 # Exemplar can be an Exemplar object, or None.
-Sample = namedtuple('Sample', ['name', 'labels', 'value', 'timestamp', 'exemplar'])
-Sample.__new__.__defaults__ = (None, None)
+sample = namedtuple('Sample', ['name', 'labels', 'value', 'timestamp', 'exemplar'])
+
+# Wrap the namedtuple to provide eage type-checking that value is a float.
```
brian-brazil: eager

prometheus_client/samples.py
```diff
@@ -32,12 +32,15 @@ def __gt__(self, other):
...
+def Sample(name, labels, value, timestamp = None, exemplar = None):
```
brian-brazil: Watch your whitespace

tests/test_exposition.py
```diff
@@ -373,7 +362,12 @@ def _expect_metric_exception(registry, expected_error):
 def test_basic_metric_families(registry, MetricFamily, value, error):
     metric_family = MetricFamily(MetricFamily.__name__, 'help')
     registry.register(Collector(metric_family, value))
-    _expect_metric_exception(registry, error)
+    try:
+        generate_latest(registry)
+    except error as exception: 
+        # Got a valid error as expected, return quietly
```
brian-brazil: Can we check the type of exception is correct?

brian-brazil:
> I've preserved that behaviour, by not appending and crashing if gsum_value
is None.

That's the right thing to do. gsum is technically optional.

prometheus_client/samples.py
mhansen: removed whitespace around equals and added two spaces around top-level members - is this ok?

prometheus_client/samples.py
mhansen: done

tests/test_exposition.py
mhansen: Unless I'm misunderstanding, we are: `error` is a type which pytest is parametrizing to be `TypeError` or `ValueError`

mhansen:
Thanks for the quick review. I pushed a new version, hope you don't mind me marking the comments as resolved, please reopen if any further questions on them.

tests/test_exposition.py
brian-brazil: Right you are

prometheus_client/samples.py
```diff
@@ -32,12 +32,17 @@ def __gt__(self, other):
 
 
 # Timestamp and exemplar are optional.
-# Value can be an int or a float.
+# Value must be a float.
```
brian-brazil: Actually this is a problem, we must support ints too. If someone passes in an int, it needs to stay as an int.

mhansen:
Could you elaborate? It seemed to me like whatever we pass to value, it’s
converted to a float for output in floatToGoString. What constraints mean
we have to keep ints ints?

mhansen:
Oh, is it so you can +1 your counter and have that work even when you run
out of floating point precision to represent a +1 step? Yeah that’s
important

brian-brazil:
Yes, OpenMetrics has int support for that so we need it.

mhansen:
FYI keeping ints as ints isn't the behaviour today: it looks like the
current openmetrics exposition.py always outputs ints as floats today (uses
floatToGoString), see this test that inputs 17 and outputs 17.0:
https://github.com/prometheus/client_python/blob/master/tests/openmetrics/test_exposition.py#L54

I can change it to keep ints as ints, but could you spell out the
requirement for me so I can add a test case for it?
- is it to output openmetrics "17" when input "17"?
- is it for avoiding precision issues with floats when you increment a
large float and it rounds down to the same float?

brian-brazil:
> - is it to output openmetrics "17" when input "17"?

Yes

> is it for avoiding precision issues with floats when you increment a
large float and it rounds down to the same float?

This isn't something this client does currently, for direct instrumentation everything is a float.

prometheus_client/samples.py
mhansen: Had a go at this, split over two commits, what do you think?

mhansen:
Had a go at preserving integer formatting in the openmetrics exposition.

prometheus_client/samples.py
```diff
@@ -36,8 +36,17 @@ def __gt__(self, other):
 # Timestamp can be a float containing a unixtime in seconds,
 # a Timestamp object, or None.
 # Exemplar can be an Exemplar object, or None.
-Sample = namedtuple('Sample', ['name', 'labels', 'value', 'timestamp', 'exemplar'])
-Sample.__new__.__defaults__ = (None, None)
+sample = namedtuple('Sample', ['name', 'labels', 'value', 'timestamp', 'exemplar'])
+
+
+# Wrap the namedtuple to provide eager type-checking that value is a float.
```
brian-brazil, line 42: Convertable to a float?

mhansen:
Hi, sorry for delay, this is still on my radar, but I haven't had a lot of
time lately -- still intending to finish this.

PR #538, opened 2020-04-22
Update canonical floats per openmetrics latest.

brian-brazil:
We now only require a specific set of very common le/quantile
labels to be in a canonical form.


brian-brazil:
@SuperQ 
Test failure is unrelated.

PR #540, opened 2020-05-06
Fix CI failures for pypy3.

brian-brazil:


brian-brazil:
@Skeen Something has broken here and I haven't the knowledge of asgi/twisted to figure out what. Would you mind taking a look?

Issue #542, opened 2020-05-07
Setting metric unit to "total"


nritsche:
a4dd93bcc6a0422e10cfa585048d1813909c6786 enforced counters ending in `_total`, but the `unit=` parameter doesn't seem to interact well with that. If I pass `unit="total"` to a `Counter`, I get a `_total_total` suffix (but only in some places?).
```
$ python3.6
Python 3.6.8 (default, Aug  7 2019, 17:28:10) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from prometheus_client import Counter
>>> c = Counter('my_failures', 'Description of counter', unit="total")
>>> c.inc(6.6)
>>> from prometheus_client import start_http_server, Summary
>>> start_http_server(8000)
```
```
$ curl localhost:8000/metrics
...
# HELP my_failures_total_total Description of counter
# TYPE my_failures_total_total counter
my_failures_total 6.6
my_failures_created 1.5888939194938827e+09
```

This is with version `0.7.1`.

brian-brazil:
That behaviour sounds correct to me, however `total` is not a good unit. What are you trying to do?

nritsche:
Oh, I see. I didn't read this careful enough: https://prometheus.io/docs/practices/naming/. The `_total` goes behind the unit. Anyways, note that the exposed metric doesn't match it's "HELP" and "TYPE".

brian-brazil:
That would be a bug, though not something that should arise in practice.

PR #543, opened 2020-05-09
Cleaning up name before appending unit on name

knabben:
Unit append in the name were being removed when *_total* exists. This moves the clean up before appending the unit. 

Refs https://github.com/prometheus/client_python/issues/542

@brian-brazil Should this "repeated" unit raise an ValueError exception instead since this suffix is already append when dumping on WSGI.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -527,6 +527,11 @@ class TestMetricWrapper(unittest.TestCase):
         self.assertRaises(ValueError, Info, 'foo', 'help', unit="x")
         self.assertRaises(ValueError, Enum, 'foo', 'help', unit="x")
 
+    def test_name_cleanup_before_unit_append(self):
+        self.assertEqual(self.counter._name, 'c')
+        self.c = Counter('c_total', 'help', unit="total", labelnames=['l'], registry=self.registry)
+        self.assertEqual(self.c._name, 'c_total')
+
 
 class TestMetricFamilies(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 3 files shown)


tests/test_core.py
```diff
@@ -527,6 +527,11 @@ def test_no_units_for_info_enum(self):
         self.assertRaises(ValueError, Info, 'foo', 'help', unit="x")
         self.assertRaises(ValueError, Enum, 'foo', 'help', unit="x")
 
+    def test_name_cleanup_before_unit_append(self):
```
brian-brazil, line 530: Can we also add a test that this is exposed correctly somewhere?

knabben: Added on tests/test_asgi.py and tests/test_wsgi.py

brian-brazil: Those are additional handlers, text_exposition would be the best place to put it.

knabben: PTAL

brian-brazil:
Thanks!

Sebadst:
Hello,

is this change intended?
I have just updated my prometheus-client from `0.7.1` to `0.8.0` and, as a result, all my metrics are broken.
Basically, if I had a metric like this:
```
JOBS = Counter(
    namespace=PROJECT,
    subsystem=ORCHESTRATOR,
    name="jobs",
    unit="total",
    documentation="Total number of jobs generated by the orchestrator.",
)
```
I was looking at it with the following:
`sum(irate(project_orchestrator_jobs_total[5m]))`.
Now I need to do:
`sum(irate(project_orchestrator_jobs_total_total[5m]))`.
It seems weird that such a change is from version `0.7.x` to `0.8.x` as it looks breaking.
Could you please let me know your thoughts on that?
Thank you.

Issue #122, opened 2016-12-01
how to instrument a bottle app


sebgoa:
Hi, this a newbie question. How can I instrument a python WSGI  `bottle` or flask app that has its own server ?

two threads ?

```
import sys

from bottle import route, run
from prometheus_client import start_http_server, Summary
import random
import time

REQUEST_TIME = Summary('request_processing_seconds','Time Spent processing request')

@route('/')
@REQUEST_TIME.time()
def handler():
    print "Hello world"
    time.sleep(random.random)

run(host='0.0.0.0', port=8080)
```

sebgoa:
I managed to do it with tornado, but is there a best practice *without* tornado:

```
from tornado.wsgi import WSGIContainer
from tornado.ioloop import IOLoop
from tornado.web import FallbackHandler, RequestHandler, Application
from prometheus_client import start_http_server, Summary

from flask import Flask
from prometheus_client import Counter

app = Flask(__name__)

REQUEST_TIME = Summary('request_processing_seconds','Time Spent processing request')

@app.route('/')
@REQUEST_TIME.time()
def handler():
    return "hello world"

class MainHandler(RequestHandler):
    def get(self):
        self.write("This message comes from Tornado")

tr = WSGIContainer(app)

application = Application([
                          (r"/tornado", MainHandler),
                          (r".*", FallbackHandler, dict(fallback=tr)),
                          ])

if __name__ == "__main__":
    application.listen(5000)
    start_http_server(9999)
    IOLoop.instance().start()
```

drawks:
Here is a working example of instrumenting a few routes in a basic bottle application and making the metrics route available an attached to the application. This also uses the `MultiProcessCollector` since it would be fairly atypical (at all but the smallest scale) to run a wsgi application as a single process. This requires that the `prometheus_multiproc_dir` variable be set. This can typically be accomplished in your wsgi runner, for example in uwsgi you could add the following to your config

`env = prometheus_multiproc_dir=/tmp/helloworld.stats`

```
from bottle import route, run, template, Bottle
from prometheus_client import multiprocess
from prometheus_client import generate_latest, CollectorRegistry, Gauge, Counter

application = Bottle()
registry = CollectorRegistry()
multiprocess.MultiProcessCollector(registry)

IN_PROGRESS = Gauge("inprogress_requests", "help", multiprocess_mode='livesum')
REQUESTS = Counter('http_requests_total', 'Description of counter', ['method', 'endpoint'])


[...]

PR #544, opened 2020-05-10
Add usage with Bottle framework

ymyzk:
There was an issue which asks how to use client_python with Bottle framework (#122) but it wasn't documented in `README.md`. For convenience of users, I'm proposing to add an example with Bottle to README. This example is different from what was proposed in #122 and closer to one for Flask.


brian-brazil:
I'm looking to reduce the number of frameworks mentioned, not expand them, as each requires ongoing maintenance. I've also honestly never heard of this framework.

I'd suggest documenting this somewhere else, such as a blog post.

PR #559, opened 2020-06-19
generate_latest: regularly sleep to release the GIL

gjcarneiro:
This is to avoid blocking asyncio or other threads so much while exposing lots
of metrics.

The reason is that I measured the time it takes for generate_latest to complete in my application, in it often takes between 100ms and 250ms.  

It already runs in an thread separate from asyncio main loop, but it's possible that CPU bound work like this might keep hold of the GIL for too long and cause latency issues in the rest of the application.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -6,6 +6,7 @@ import os
 import socket
 import sys
 import threading
+import time
 from wsgiref.simple_server import make_server, WSGIServer, WSGIRequestHandler
 
 from .openmetrics import exposition as openmetrics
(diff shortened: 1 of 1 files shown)


brian-brazil:
This doesn't seem right at all. The GIL is used for very granular correctness, not coarse scheduling.

If generating metrics is taking that long then I'd suggest that the issue is too high cardinality.

gjcarneiro:
It's just those damn histograms, expanding to so much...

But, researching some more, you're right.  I'm jumping the gun a bit here.  I found this: https://opensource.com/article/17/4/grok-gil

> If a thread runs uninterrupted for 1000 bytecode instructions in Python 2, or runs 15 milliseconds in Python 3, then it gives up the GIL and another thread may run.

So, it sounds like Python is already doing the right thing.  15ms is very reasonable.

Never mind!

PR #564, opened 2020-07-03
Ignore utf-8 errors in generate_latest

vmarkovtsev:
This little f*cker has almost cost us draining our monthly Sentry events limit in the beginning of the month. There was an incoming label or other text with broken Unicode, which resulted in

```
'utf-8' codec can't encode character '\udca0' in position 2210: surrogates not allowed
```

(the position is random, never repeats)

We `generate_latest()` every few seconds as the main part of server health checks. Python 3.8. Once it started erroring, it never stopped on multiple servers. So stressful.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -144,7 +144,7 @@ def generate_latest(registry=REGISTRY):
                                                        metric.documentation.replace('\\', r'\\').replace('\n', r'\n')))
             output.append('# TYPE {0}{1} gauge\n'.format(metric.name, suffix))
             output.extend(lines)
-    return ''.join(output).encode('utf-8')
+    return ''.join(output).encode('utf-8', errors='ignore')
 
 
 def choose_encoder(accept_header):


vmarkovtsev:
OMG we support py2.6 here

vmarkovtsev:
Reviewers: I can conditionally enable this on py 2.7+ if the feature policy is graceful degradation, otherwise happy to patch locally and close the PR.

brian-brazil:
Can you explain exactly what the issue is here? Silently ignoring encoding errors does not seem wise, and if you've invalid UTF-8 coming in I'd suggest fixing it up before it gets to this library as valid UTF-8 is required.

vmarkovtsev:
I cannot explain this - it happened in our prod and the actual string is not logged (apparently - it fails to encode).

There are a hundred places where the metrics are recorded, and (apparently again) all the prometheus_client APIs entry points consume strings, not raw bytes. Sorry, but trying to encode every incoming string in an API wrapper is not something I am looking forward to regarding the performance overhead.

OK, so if anybody googles this error and lands here, just know that one day your prod may blow up for no visible reason if you don't apply this workaround.

What would help to turn this (suspicious, impure, unexplainable) raw code patch to at least a monkey patch is abstracting out `''.join(output).encode('utf-8')` to a new function. Say, named `_assemble_output(List[str]) -> bytes`. How about that?

vmarkovtsev:
Not to seem completely proveless, here is a Sentry screenshot:
image

brian-brazil:
> (apparently again) all the prometheus_client APIs entry points consume strings, not raw bytes. 

Yes, the APIs expect valid UTF-8 strings. Even with raw bytes, we'd still have to check that they are valid UTF-8 strings as either the Python client lets you know that you have invalid data or Prometheus will when it fails the scrape. I think having the error surfaced sooner is better.

> How about that?

The challenge is that if we munge the inputs then that could lead to collisions in label values, which will cause its own problems as an explicit failure is being changed to a silent data corruption. The only safe thing we can do on invalid data is to raise an exception. Maybe there's a way to have a better exception here to help you track down the culprit?

brian-brazil:
Closing per above, this is not safe or correct.

PR #572, opened 2020-08-10
Multiproc with multiple containers

ChuckHend:
We starting running into some issues with multiproc when running multiple Docker containers on a single host with a shared volume between the containers writing metrics and a Flask app Collector.

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,6 +550,44 @@ from prometheus_client import Gauge
 IN_PROGRESS = Gauge("inprogress_requests", "help", multiprocess_mode='livesum')
 ```
 
+**5. Multi-container configuration with Docker**:
+
+Given the docker-compose configuration assumed to deploy with `docker-compose up --scale worker=5`:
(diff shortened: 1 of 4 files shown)


brian-brazil:
Thanks for your PR. This does not sound like a sane deployment model, and certainly is not something to be encouraged by mentioning it in the readme. It's multi-process, not multi-container or multi-machine.

I would recommend running one much larger container, and if not give each container its own directory.

ChuckHend:
We're looking to keep container size down and want to have the ability to the scale number of container replicas up/down. I think if we need to go with separate directories then we will also need be deploying another MultiProcessCollector() (another Flask App) for each additional container, right?

brian-brazil:
Yes, one per container

If you're all on one machine and already multi-processs, there's not too much point in having many containers on that machine. That's just adding overhead. That sort of scaling is useful across machines, not within a machine.

PR #577, opened 2020-09-10
Catch name clashes in the OM parser.

brian-brazil:
Make sure we're catching all the cases in the client too.
This means you can't have a foo counter and a foo_created gauge.

@SuperQ Per the most recent discussions.


brian-brazil:
@SuperQ Updated as per the meeting just there.

SuperQ:
Looks like this needs a rebase.

brian-brazil:
Rebased.

robskillington (approved):
LGTM

PR #589, opened 2020-10-21
Add example for initializing metrics with labels

mpitt:
I know metrics should be initialized, but I wasn't able to find how to do it if they have labels in this doc. I eventually found the answer in [Brian's blog](https://www.robustperception.io/existential-issues-with-metrics), which is always an amazing source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ c.labels(method='post', endpoint='/submit').inc()
 ```
 
 Metrics with labels are not initialized when declared, because the client can't
-know what values the label can have. It is reccommended to initialize the label
+know what values the label can have. It is recommended to initialize the label
 values by calling the `.label()` method alone:
 
 ```python
--- a/README.md
+++ b/README.md
@@ -222,17 +222,13 @@ c.labels(method='post', endpoint='/submit').inc()
 
 Metrics with labels are not initialized when declared, because the client can't
 know what values the label can have. It is recommended to initialize the label
-values by calling the `.label()` method alone:
+values by calling the `.labels()` method alone:
 
 ```python
 from prometheus_client import Counter
 c = Counter('my_requests_total', 'HTTP Failures', ['method', 'endpoint'])
-# initialize the label values
 c.labels('get', '/')
 c.labels('post', '/submit')
-# then use the metrics as normal
-c.labels('get', '/').inc()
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,6 +207,11 @@ Taking a counter as an example:
 ```python
 c = Counter('my_requests_total', 'HTTP Failures', ['method', 'endpoint'])
+# when using labels, metrics are not initialized at creation
```
brian-brazil, line 210: This doesn't really fit here, as a less common use case is being wedged in front of the simple normal usage.

This would be better as a separate thing down the send of this section.

mpitt: Fair enough, I thought it was a strong enough recommendation to highlight it in the normal use case. Pushed a revision.

README.md
```diff
@@ -220,6 +220,21 @@ c.labels(method='get', endpoint='/').inc()
...
+know what values the label can have. It is recommended to initialize the label
+values by calling the `.label()` method alone:
+
+```python
+from prometheus_client import Counter
+c = Counter('my_requests_total', 'HTTP Failures', ['method', 'endpoint'])
+# initialize the label values
+c.labels('get', '/')
+c.labels('post', '/submit')
+# then use the metrics as normal
```
brian-brazil, line 233: You don't need from this bit onwards.

README.md
```diff
@@ -220,6 +220,21 @@ c.labels(method='get', endpoint='/').inc()
 c.labels(method='post', endpoint='/submit').inc()
 ```
 
+Metrics with labels are not initialized when declared, because the client can't
+know what values the label can have. It is recommended to initialize the label
+values by calling the `.label()` method alone:
```
brian-brazil, line 225: labels

README.md
```diff
@@ -220,6 +220,21 @@ c.labels(method='get', endpoint='/').inc()
...
+
+```python
+from prometheus_client import Counter
+c = Counter('my_requests_total', 'HTTP Failures', ['method', 'endpoint'])
+# initialize the label values
```
brian-brazil, line 230: This comment isn't needed.

README.md
mpitt: oops

brian-brazil:
Thanks!

PR #591, opened 2020-11-03
Add few code fixes.

rayandas:
This PR fixes few code qualities.


brian-brazil:
Thanks for your PR. These appears to be Python 3 specific changes, however this repo also supports Python 2.

rayandas:
@brian-brazil Ah! Right. closing it.

PR #592, opened 2020-11-05
Convert CI to CircleCI

SuperQ:
Replace Travis with CircleCI.
* Update Python version list.

Signed-off-by: Ben Kochie <superq@gmail.com>


brian-brazil:
This doesn't seem to be doing quite the same thing. Some supported versions have been lost from testing, and we're no longer testing that things still work when none of the optional modules are installed.

SuperQ:
Some of those old versions aren't supported by CircleCI (2.6, 3.4 for example).

brian-brazil:
There's a 3.4: https://hub.docker.com/layers/circleci/python/3.4/images/sha256-6403a42dff13d3593b541f75bb695584b5d6695675a4a1d1b2773081ede22e05?context=explore

It might be finally time to drop 2.6.

SuperQ:
I was trying to go directly to CircleCI's "next gen" images.

brian-brazil:
If we're dropping versions, then setup.py needs updating.

SuperQ:
I re-added Python 3.4 support.

brian-brazil:
:+1:

PR #602, opened 2020-11-20
Make OM code effectively 100% covered.

brian-brazil:
Also remove some dead code. The untaken branches claimed by coverage are
due to loops that always iterate at least once, or because state
machines only ever have known states.

@robskillington this is mostly testing the parser implementation, but this caught some missing general cases too.


robskillington (approved):
LGTM

PR #609, opened 2021-01-01
Fix Package Name typo in pip install instructions

ngrsdhbw:
Change package name in pip install instructions to match pypi with - instead of _

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The official Python 2 and 3 client for [Prometheus](http://prometheus.io).
 
 **One**: Install the client:
 ```
-pip install prometheus_client
+pip install prometheus-client
 ```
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
The current instructions seem to work for me, are you sure this is broken?

ngrsdhbw:
> The current instructions seem to work for me, are you sure this is broken?

This is probably due to the fact that pypi automatically converts the _ to - since both of these links work:
https://pypi.org/project/prometheus_client/
https://pypi.org/project/prometheus-client/

However when you are using a tool wihtout this logic like pycharm you do not find the package that you are looking for if you input the _
pycharm pypi underscore problem2

brian-brazil:
Thanks!

I don't see a problem here, but we should be consistent.

PR #610, opened 2021-01-05
Added basic authentication option in a start_http_server method

thebearinmind:
@brian-brazil Added basic authentication option in da start_http_server method. It was required in our company, so I though it might be useful for others as well.


brian-brazil:
Thanks for your PR, this is not something I'm looking to add. First start_http_server is meant to be a very simple server to get you up and running quickly, without any extra trappings. It's expected that in most production use cases you'd be using your existing http server. Secondly this is adding a dependency, and this library purposefully does not have any to avoid clashes and related fun.

michael1011101:
> Thanks for your PR, this is not something I'm looking to add. First start_http_server is meant to be a very simple server to get you up and running quickly, without any extra trappings. It's expected that in most production use cases you'd be using your existing http server. Secondly this is adding a dependency, and this library purposefully does not have any to avoid clashes and related fun.

Hi @brian-brazil,
Obviously, start_http_server is a simple and quick way to set up a server. Then, do you have plan or timeline to support SSL/TLS? Also, I am working on exporter and need to support https/ssl request. What can I do to support it? any best practice please share with me.

brian-brazil:
No, that's not planned either but you can do it yourself.

PR #619, opened 2021-02-03
add lock() and unlock() methods to registry class

lweith:
I write code that polls a large number of devices via threads and then feed the complete set of results into the registry as a big chunk and want all the data consistent at each scrape and that requires locking.  To that end, I have been using a wrapper class to implement locking for a while and thought others might find it helpful if the internal locking were presented directly.


lweith:
I actually wanted to block collection without blocking updates.  This change doesn't do that.

PR #621, opened 2021-02-08
Handing None Type request method

cVoltic:
Added a change where Nulltype prevent string concatenation

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -90,6 +90,8 @@ def generate_latest(registry=REGISTRY):
 
     def sample_line(line):
         if line.labels:
+            if 'method' in line.labels and line.labels['method'] is None:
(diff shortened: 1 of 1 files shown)


csmarchbanks (commented):
Hello, thank you for your contribution!

What is the problem you are trying to solve here? The exposition format should not be handling specific labels differently than any others, so looking for a `method` label is not the correct behavior, and passing in the correct `method` is the responsibility of the caller.

cVoltic:
Hi Chris, thanks you for the feed back. Essentially, the problem I am experiencing is running Prometheus client scrapper on locust tool. I'm running into a string concat error particularly when Prometheus reach

'method':None

cVoltic:
Hi Chris, I think I'll close this now since I applied what you mentioned and it worked. Sorry for the confusion!

Issue #131, opened 2017-01-25
Add HTTPRedirectHandler for pushing to gateway


FUSAKLA:
If used in some kind of virtualization layer with own DNS it`s necessary to handle redirects.

Usage of python2 `urllib2.HTTPRedirectHandler` and python3 `urllib.request.HTTPRedirectHandler`
should solve this issue IMHO.

Or is there downside of redirecting which I'm missing?

brian-brazil:
I don't see a problem there, I'd suggest sending a PR after #126 is in.

espears1:
@brian-brazil is it possible to re-open this one? Also, do you know any reason why direct use of `HTTPRedirectHandler` wouldn't work? I am creating a handler from this class, yet I still see some HTTP 301 not getting handled.

csmarchbanks:
Hi @espears1, I am happy to reopen this as it seems like redirects are not working.

I believe that `HTTPRedirectHandler` is actually one of the default classes used already, see the [docs](https://docs.python.org/3/library/urllib.request.html#urllib.request.build_opener) for more information on the default handlers used. I think the issue is that the `HTTPRedirectHandler` does not redirect PUT requests, and will change POST requests to GET which will not work for push gateway.

I think the best solution would be to add another handler (like basic_auth_handler) that handles PUT and POST redirects including sending data. The reason for a separate handler is that it is encouraged (technically a MUST in [RFC 2616](https://tools.ietf.org/html/rfc2616#section-10.3.1)) to have user confirm a redirect POST/PUT, and I would say a separate handler classifies as the user giving confirmation that a redirect is expected. Plus having the default handler continue to use python defaults would be nice. Happy to hear opposing views though.

espears1:
@csmarchbanks re: https://github.com/prometheus/client_python/issues/131#issuecomment-771775942

First of all thank you so much for looking into this and providing that extra context. That definitely helps with my question and I will do more research using that info.

I did want to clarify what you mean by saying 

> `HTTPRedirectHandler` is actually one of the default classes used already.

I don't believe that's true for the prometheus client. As in the link above, I think there is just `default_handler` which hard-codes the use of `HTTPHandler` and then a separate one for `basic_auth_handler`. My initial idea was just to copy/paste `default_handler` and adjust the base handler class to use `HTTPRedirectHandler`, like this:

```
def redirect_handler(url, method, timeout, headers, data):

    def handle():
        request = Request(url, data=data)
        request.get_method = lambda: method
[...]

PR #622, opened 2021-02-08
Implement redirect handler option.

espears1:
This fixes https://github.com/prometheus/client_python/issues/131 and a significant amount of content is discussed in that issue.

This PR adds a new redirect handler utility class that deliberately forwards an unmodified request to the `Location` header provided by a redirect, while preserving the method, headers and data of the original request. As discussed in comments of the PR, HTTP RFC 2616 requires explicit user approval for redirects in certain situations that may alter the meaning or effect of the request. Given this, the default `HTTPRedirectHandler` in `urllib` does not honor redirects for `PUT` and does not preserve all headers and data.

In the case of Prometheus, many users will have some type of in-house "private" deployment of a Prometheus pushgateway server, and e.g. in a deployment infrastructure like Kubernetes where the ultimate IP-based host address could change often, there may be a need for a purely cosmetic redirect that fronts this host address with an easier static URL. In that case, the user would like to "opt in" to trusted pass-through redirects for all method types (including PUT).

This PR provides `passthrough_redirect_handler` to support this specific redirect option. Use of this special redirect handler would constitute the user's explicit approval for RFC2616.

Additionally, this PR adds a new "redirect" HTTPD server to the unit tests for the pushgateway, in order to simulate a real redirect during testing and confirm it is handled as expected. I've locally validated all testing and flake8 instructions - and will work to resolve any requested changes or build issues.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -10,7 +10,7 @@ from wsgiref.simple_server import make_server, WSGIRequestHandler, WSGIServer
 
 from .openmetrics import exposition as openmetrics
 from .registry import REGISTRY
-from .utils import floatToGoString
+from .utils import floatToGoString, PrometheusRedirectHandler
 
 try:
     from urllib import quote_plus
@@ -141,7 +141,7 @@ def generate_latest(registry=REGISTRY):
             raise
 
         for suffix, lines in sorted(om_samples.items()):
-            output.append('# HELP {0}{1} {2}\n'.format(metric.name, suffix, 
+            output.append('# HELP {0}{1} {2}\n'.format(metric.name, suffix,
                                                        metric.documentation.replace('\\', r'\\').replace('\n', r'\n')))
             output.append('# TYPE {0}{1} gauge\n'.format(metric.name, suffix))
             output.extend(lines)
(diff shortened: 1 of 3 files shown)


tests/test_exposition.py
```diff
@@ -330,6 +354,16 @@ def my_auth_handler(url, method, timeout, headers, data):
...
 
+    def test_push_with_redirect_handler(self):
+        def my_redirect_handler(url, method, timeout, headers, data):
+            return passthrough_redirect_handler(url, method, timeout, headers, data)
+
+        push_to_gateway(self.address, "my_job_with_redirect", self.registry, handler=my_redirect_handler)
+        self.assertEqual(self.requests[0][0].command, 'PUT')
+        self.assertEqual(self.requests[0][0].path, '/metrics/job/my_job_with_redirect')
+        self.assertEqual(self.requests[0][0].headers.get('content-type'), CONTENT_TYPE_LATEST)
+        self.assertEqual(self.requests[0][1], b'# HELP g help\n# TYPE g gauge\ng 0.0\n')
```
csmarchbanks, line 365: Should there also be assertions for the second(redirected) request?

csmarchbanks (commented):
Thanks! Generally this is looking good, just a couple of questions to begin with.

prometheus_client/exposition.py
```diff
@@ -10,7 +10,7 @@
 
 from .openmetrics import exposition as openmetrics
 from .registry import REGISTRY
-from .utils import floatToGoString
+from .utils import floatToGoString, PrometheusRedirectHandler
```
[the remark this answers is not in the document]
espears1: That's a good suggestion. It wasn't immediately obvious where the class should go, so I thought of utils - but setting it up with a clearer indication that end-users probably never need to touch it is better. I'll make this change.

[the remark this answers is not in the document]
espears1: fixed in https://github.com/prometheus/client_python/pull/622/commits/932cc17d3e103e7396b20a595c4af04fe2409bf0

tests/test_exposition.py
espears1: fixed in https://github.com/prometheus/client_python/pull/622/commits/912cb183ed6263e272869fb0b571e36c55e3073e

Let me know if you feel this is enough coverage, or if there's a different approach to take which can cover the test better.

espears1: Adding the extra test revealed some further complexity about how to handle `Request` objects in Python 2.7. It looks quite hairy, no simple solution. I will pick it back up tomorrow.

espears1: update: I was just making a dumb mistake on the Python 2.7 version check. I fixed that and it seems to have resolved the error, so I think it is ready for a new round of review.

csmarchbanks: I think this test looks reasonable.

csmarchbanks (approved):
Thanks, this looks good to me!

PR #624, opened 2021-02-09
Multiprocess environment variable in upper case

marcaurele:
I would like to propose a change to move the `prometheus_multiproc_dir` env variable in upper case to follow the convention of such variable. It feels strange to have to declare it in lower case.

The code adds a deprecation warning to inform the user about the change and transform it to upper case if needed. It also take care of precedence in case both are declared.

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,7 +502,7 @@ There's several steps to getting this working:
 
 **1. Gunicorn deployment**:
 
-The `prometheus_multiproc_dir` environment variable must be set to a directory
+The `PROMETHEUS_MULTIPROC_DIR` environment variable must be set to a directory
 that the client library can use for metrics. This directory must be wiped
 between Gunicorn runs (before startup is recommended).
 
--- a/README.md
+++ b/README.md
@@ -502,7 +502,7 @@ There's several steps to getting this working:
 
 **1. Gunicorn deployment**:
 
-The `PROMETHEUS_MULTIPROC_DIR` environment variable must be set to a directory
+The `prometheus_multiproc_dir` environment variable must be set to a directory
 that the client library can use for metrics. This directory must be wiped
 between Gunicorn runs (before startup is recommended).
 
--- a/prometheus_client/values.py
+++ b/prometheus_client/values.py
@@ -54,10 +54,10 @@ def MultiProcessValue(process_identifier=os.getpid):
             self._params = typ, metric_name, name, labelnames, labelvalues, multiprocess_mode
             # This deprecation warning can go away in a few releases when removing the compatibility
             if 'prometheus_multiproc_dir' in os.environ and 'PROMETHEUS_MULTIPROC_DIR' in os.environ:
-                warnings.warn("You have declared both env variable in upper case and lower case, keep the upper case only PROMETHEUS_MULTIPROC_DIR'", DeprecationWarning)
+                warnings.warn("You have declared both env variable in upper case and lower case, keep the upper case only PROMETHEUS_MULTIPROC_DIR", DeprecationWarning)
             if 'prometheus_multiproc_dir' in os.environ and 'PROMETHEUS_MULTIPROC_DIR' not in os.environ:
                 os.environ['PROMETHEUS_MULTIPROC_DIR'] = os.environ['prometheus_multiproc_dir']
-                warnings.warn("You should declare the env variable in upper case PROMETHEUS_MULTIPROC_DIR'", DeprecationWarning)
+                warnings.warn("You should declare the env variable in upper case PROMETHEUS_MULTIPROC_DIR", DeprecationWarning)
             with lock:
                 self.__check_for_pid_change()
                 self.__reset()
--- a/prometheus_client/values.py
+++ b/prometheus_client/values.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import os
 from threading import Lock
+import warnings
 
 from .mmap_dict import mmap_key, MmapedDict
 
@@ -64,7 +65,7 @@ def MultiProcessValue(process_identifier=os.getpid):
                 file_prefix = typ
             if file_prefix not in files:
                 filename = os.path.join(
-                    os.environ['prometheus_multiproc_dir'],
+                    os.environ.get('PROMETHEUS_MULTIPROC_DIR'),
                     '{0}_{1}.db'.format(file_prefix, pid['value']))
 
                 files[file_prefix] = MmapedDict(filename)
@@ -108,7 +109,10 @@ def get_value_class():
     # This needs to be chosen before the first metric is constructed,
     # and as that may be in some arbitrary library the user/admin has
     # no control over we use an environment variable.
-    if 'prometheus_multiproc_dir' in os.environ:
+    if 'prometheus_multiproc_dir' in os.environ and 'PROMETHEUS_MULTIPROC_DIR' not in os.environ:
+        os.environ['PROMETHEUS_MULTIPROC_DIR'] = os.environ['prometheus_multiproc_dir']
+        warnings.warn("You should declare the env variable in upper case PROMETHEUS_MULTIPROC_DIR'", DeprecationWarning)
+    if 'PROMETHEUS_MULTIPROC_DIR' in os.environ:
         return MultiProcessValue()
     else:
         return MutexValue
--- a/prometheus_client/values.py
+++ b/prometheus_client/values.py
@@ -52,6 +52,12 @@ def MultiProcessValue(process_identifier=os.getpid):
 
         def __init__(self, typ, metric_name, name, labelnames, labelvalues, multiprocess_mode='', **kwargs):
             self._params = typ, metric_name, name, labelnames, labelvalues, multiprocess_mode
+            # This deprecation warning can go away in a few releases when removing the compatibility
+            if 'prometheus_multiproc_dir' in os.environ and 'PROMETHEUS_MULTIPROC_DIR' in os.environ:
+                warnings.warn("You have declared both env variable in upper case and lower case, keep the upper case only PROMETHEUS_MULTIPROC_DIR'", DeprecationWarning)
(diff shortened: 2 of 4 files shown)


marcaurele:
cc @csmarchbanks

README.md
```diff
@@ -502,7 +502,7 @@ There's several steps to getting this working:
 
 **1. Gunicorn deployment**:
 
-The `prometheus_multiproc_dir` environment variable must be set to a directory
+The `PROMETHEUS_MULTIPROC_DIR` environment variable must be set to a directory
```
csmarchbanks, line 505: Let's keep the README value lowercase for at least a release or two so people will not be confused if using an older version.

csmarchbanks (commented):
Thank you for your contribution!

I think accepting an upper case environment variable certainly makes sense, I am not sure we need to deprecate the lower case version though as it is still valid. I don't have a windows machine to test on right now, but is there any concern about case insensitivity?

README.md
marcaurele: Would be good to start a changelog.md to keep tracks of changes between versions.

prometheus_client/values.py
```diff
@@ -108,7 +109,10 @@ def get_value_class():
     # This needs to be chosen before the first metric is constructed,
     # and as that may be in some arbitrary library the user/admin has
     # no control over we use an environment variable.
-    if 'prometheus_multiproc_dir' in os.environ:
+    if 'prometheus_multiproc_dir' in os.environ and 'PROMETHEUS_MULTIPROC_DIR' not in os.environ:
+        os.environ['PROMETHEUS_MULTIPROC_DIR'] = os.environ['prometheus_multiproc_dir']
+        warnings.warn("You should declare the env variable in upper case PROMETHEUS_MULTIPROC_DIR'", DeprecationWarning)
```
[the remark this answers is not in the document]
marcaurele: I would like to encourage people to use the upper case to follow standard practice in Unix env variables. If we don't put that in the readme, we don't say anything in the code to promote the use of the uppercase I don't see how people will move to that variable name in upper case.

prometheus_client/values.py
```diff
@@ -64,7 +65,7 @@ def __reset(self):
                 file_prefix = typ
             if file_prefix not in files:
                 filename = os.path.join(
-                    os.environ['prometheus_multiproc_dir'],
+                    os.environ.get('PROMETHEUS_MULTIPROC_DIR'),
```
csmarchbanks, line 68: For users that set the environment variable at runtime with something like `os.environ.setdefault('prometheus_multiproc_dir', '/tmp/multiproc')` will only checking the upper case version here cause their code to stop working?

Looking through github I see that people do sometimes specifically set the ValueClass to MultiProcessValue rather than use get_value_class.

prometheus_client/values.py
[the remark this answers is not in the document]
csmarchbanks: That seems reasonable, let's keep the deprecation notice. I do plan to merge your README change after I do a release with this change as well.

csmarchbanks (commented):
Thanks for the reasoning on the deprecation notice. I did a bit of testing and found one concern.

prometheus_client/values.py
marcaurele: Good you looked at that. I'll push a fix to cover that use case too.

marcaurele: I updated the PR but now I'm struggling to write a correct set of tests. Can you help me on what?

csmarchbanks: Thanks for updating the PR, I think it looks reasonable but will do one more round of testing after getting the tests working.

It looks like the issue in the tests is that the `__init__` function for an `MmapedValue` is not actually called until a metric is created. That means in each of your tests you will need to create a registry, collector, and perhaps a counter like the other tests below.

Here is a patch that seems to work for me, but double check that it looks sane to you too. Plus you might be able to get rid of some of the duplication.

```diff
diff --git a/tests/test_multiprocess.py b/tests/test_multiprocess.py
index a2a1dbf..845cc3d 100644
--- a/tests/test_multiprocess.py
+++ b/tests/test_multiprocess.py
@@ -32,12 +32,17 @@ class TestMultiProcessDeprecation(unittest.TestCase):
     def tearDown(self):
         del os.environ['prometheus_multiproc_dir']
         del os.environ['PROMETHEUS_MULTIPROC_DIR']
+        values.ValueClass = MutexValue
         shutil.rmtree(self.tempdir)

     def test_deprecation_warning(self):
         os.environ['prometheus_multiproc_dir'] = self.tempdir
         with warnings.catch_warnings(record=True) as w:
-            get_value_class()
+            values.ValueClass = get_value_class()
+            registry = CollectorRegistry()
+            collector = MultiProcessCollector(registry, self.tempdir)
+            Counter('c', 'help', registry=None)
+
             assert os.environ['PROMETHEUS_MULTIPROC_DIR'] == self.tempdir
             assert len(w) == 1
             assert issubclass(w[-1].category, DeprecationWarning)
@@ -47,8 +52,12 @@ class TestMultiProcessDeprecation(unittest.TestCase):
         os.environ['prometheus_multiproc_dir'] = 'should not be picked'
         os.environ['PROMETHEUS_MULTIPROC_DIR'] = self.tempdir
         with warnings.catch_warnings(record=True) as w:
-            get_value_class()
-            assert len(w) == 0
+            values.ValueClass = get_value_class()
+            registry = CollectorRegistry()
+            collector = MultiProcessCollector(registry, self.tempdir)
+            Counter('c', 'help', registry=None)
+
+            assert len(w) == 1


 class TestMultiProcess(unittest.TestCase):
```

marcaurele: Thanks @csmarchbanks , tests are passing now.

prometheus_client/values.py
```diff
@@ -51,6 +52,12 @@ class MmapedValue(object):
 
         def __init__(self, typ, metric_name, name, labelnames, labelvalues, multiprocess_mode='', **kwargs):
             self._params = typ, metric_name, name, labelnames, labelvalues, multiprocess_mode
+            # This deprecation warning can go away in a few releases when removing the compatibility
+            if 'prometheus_multiproc_dir' in os.environ and 'PROMETHEUS_MULTIPROC_DIR' in os.environ:
+                warnings.warn("You have declared both env variable in upper case and lower case, keep the upper case only PROMETHEUS_MULTIPROC_DIR'", DeprecationWarning)
```
csmarchbanks, line 57: You have an extra single tick after PROMETHEUS_MULTIPROC_DIR here.

prometheus_client/values.py
```diff
@@ -51,6 +52,12 @@ class MmapedValue(object):
 
         def __init__(self, typ, metric_name, name, labelnames, labelvalues, multiprocess_mode='', **kwargs):
             self._params = typ, metric_name, name, labelnames, labelvalues, multiprocess_mode
+            # This deprecation warning can go away in a few releases when removing the compatibility
+            if 'prometheus_multiproc_dir' in os.environ and 'PROMETHEUS_MULTIPROC_DIR' in os.environ:
```
csmarchbanks, line 56: If I instantiate multiple metrics, I end up getting both warnings because this clause is triggered on the second metric after the capital environment variable is set. Is there a good way to not get two warnings in that case? Here is a quick example:
```python
from prometheus_client import generate_latest, CollectorRegistry, multiprocess, Gauge

Gauge('metric_1', 'First metric', registry=registry)
Gauge('metric_2', 'Second metric', registry=registry)
```
Since they are just warnings, not a big deal, but it is a bit confusing.

One option would be to always warn if the lowercase is set, then only replace if the capitalized version is not set? Users with both set would only get the `You should declare the env variable in upper case PROMETHEUS_MULTIPROC_DIR` in that case.

prometheus_client/values.py
```diff
@@ -51,6 +52,12 @@ class MmapedValue(object):
...
+                warnings.warn("You have declared both env variable in upper case and lower case, keep the upper case only PROMETHEUS_MULTIPROC_DIR'", DeprecationWarning)
+            if 'prometheus_multiproc_dir' in os.environ and 'PROMETHEUS_MULTIPROC_DIR' not in os.environ:
+                os.environ['PROMETHEUS_MULTIPROC_DIR'] = os.environ['prometheus_multiproc_dir']
+                warnings.warn("You should declare the env variable in upper case PROMETHEUS_MULTIPROC_DIR'", DeprecationWarning)
```
csmarchbanks, line 60: Also an extra single tick in this line.

csmarchbanks (commented):
Nice work, I noticed a couple extra characters in the warnings, and came across one edge case in testing. Curious on your thoughts there, if it would be too complex I think the current behavior is fine.

prometheus_client/values.py
marcaurele: Good catch!

prometheus_client/values.py
marcaurele: Yes, correct. I'd rather display the warning for the one which only defined the value in small letter. I'll remove the case for both.

csmarchbanks: :+1: Sounds good, let me know once that is complete and I will test again.

marcaurele: Have a look. I also change the message on the deprecation warning as it did not sound so much as one. I think the wording is much better now. Don't hesitate if you have any other proposal to do.

csmarchbanks (commented):
The new messaging looks good to me. I don't know how I missed it in my previous tests but there are a few places in `prometheus_client/multiprocess.py` where `prometheus_multiproc_dir` is still used. Specifically, when creating a MultiProcessCollector, and during `mark_process_dead`. Those will need to be updated too. Sorry about missing that earlier and needing extra revisions!

It looks like the tests are passing only because self.tempdir is passed in as a path rather than the default of None. It might be worth having a test case for the None scenario as well.

marcaurele:
@csmarchbanks you can do your tests again. Let me know if there's anything else that needs to be changed.

csmarchbanks (approved):
Awesome, I ran through my tests again and this is looking good. Thanks for all the work!

PR #628, opened 2021-02-14
readme: inform on the upper case variable

marcaurele:
Following PR #624, so it's not forgotten to merge that change to promote the upper case variable in a few release.

cc @csmarchbanks


csmarchbanks (approved):
Figure it has been long enough that v0.10.1 looks like a good release.

PR #631, opened 2021-02-23
Add separate handling of the /favicon.ico path in the default wsgi implementation

ngrsdhbw:
By default many browsers will attempt to load a /favicon.ico resource from the server. Currently the default wsgi implementation serves metrics on any path and thus will incorrectly return metrics when being asked for a favicon. This PR fixes this behavior by serving an empty response.


csmarchbanks (approved):
Thanks, this seems reasonable to me!

PR #644, opened 2021-04-07
mark_process_dead respects old env var

### LINKED ISSUE #643: Issue with prometheus_multiproc_dir deprecation
Hi !

I noticed an issue with prometheus_multiproc_dir deprecation in the latest 0.10.0 version.
In `mark_process_dead()` the environment variable transition code is not present, which means that if the uppercase version is not set, the function does not work.
https://github.com/prometheus/client_python/pull/624/files#diff-95ab27209a655ee3a508d436368ba4e1d155784c5885e18be09b61c2a5ec03efL155

Since we can call `mark_process_dead()` without going through the `__init__()` methods that set the new environment variable, it would required something like this : https://github.com/prometheus/client_python/pull/624/files#diff-f1ab5b765643c2beafab5bcd9874a839f7ebf080300e1ee0dbdb821755f8e904R56-R58

csmarchbanks:
Some users use mark_process_dead without going through any of the
`__init__()` logic which sets the new environment variable.

I chose not to put in the warnings in another place as a user will go
through the deprecation path somewhere already.

Fixes #643

### DIFF
--- a/prometheus_client/multiprocess.py
+++ b/prometheus_client/multiprocess.py
@@ -157,7 +157,7 @@ class MultiProcessCollector(object):
 def mark_process_dead(pid, path=None):
     """Do bookkeeping for when one process dies in a multi-process setup."""
     if path is None:
-        path = os.environ.get('PROMETHEUS_MULTIPROC_DIR')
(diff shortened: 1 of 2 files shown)


sd2k (approved):
LGTM!

Issue #649, opened 2021-05-07
write_to_textfile: fails to overwrite file on windows


johanvdw:
On linux, write_to_textfile can overwrite an existing file.
However, this fails if we try this on windows:
```
"C:\...\dashboard\metrics.py", line 169, in write_metrics
write_to_textfile(outfile, registry)
File "C:\...\env\lib\site-packages\prometheus_client\exposition.py", line 275, in write_to_textfile
os.rename(tmppath, path)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: '..\metrics\cable_alert.prom.6092.3608' -> '..\metrics\cable_alert.prom'
```

bluppfisk:
should be good to close now with #650 merged. Thanks Chris!

PR #650, opened 2021-05-10
allow textfile to be overwritten on windows

bluppfisk:
This is a prospective fix for issue #649 

For POSIX installations, nothing changes. For Windows installations, if python>=3.3 is detected, it will use `os.replace`, else it will do a double operation of `os.remove` followed by `os.rename`, either of which may fail and result in (temporarily or permanently) broken metrics.

This can be refined to catch such errors.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -271,8 +271,20 @@ def write_to_textfile(path, registry):
     tmppath = '%s.%s.%s' % (path, os.getpid(), threading.current_thread().ident)
     with open(tmppath, 'wb') as f:
         f.write(generate_latest(registry))
-    # rename(2) is atomic.
-    os.rename(tmppath, path)
+    
+    # rename(2) is atomic but fails on Windows if the destination file exists
+    if os.name == 'posix':
+        os.rename(tmppath, path)
+    else:
+        if sys.version_info <= (3,3):
+            # Unable to guarantee atomic rename on Windows and Python<3.3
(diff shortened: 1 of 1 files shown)


bluppfisk:
Python 3.4 test fails because the typing package was added in v3.5. Unsure how to fix

prometheus_client/exposition.py
```diff
@@ -271,8 +271,20 @@ def write_to_textfile(path, registry):
     tmppath = '%s.%s.%s' % (path, os.getpid(), threading.current_thread().ident)
     with open(tmppath, 'wb') as f:
         f.write(generate_latest(registry))
-    # rename(2) is atomic.
-    os.rename(tmppath, path)
+    
+    # rename(2) is atomic but fails on Windows if the destination file exists
+    if os.name == 'posix':
```
csmarchbanks, line 276: I wonder if it is worth keeping the `posix` behavior as the `else` case, and only do the remove/rename or replace logic if os.name is `nt`. That way it is clear that the special behavior is for windows?

csmarchbanks (commented):
Thanks, one comment, but I don't feel too strongly about it.

The test failure appears to be due to a bad release of attrs: https://www.attrs.org/en/stable/changelog.html, but that release has been yanked so i am not sure why it is still being chosen in the tests. I will dig in a bit further when I have a moment.

csmarchbanks (approved):
:+1: Thanks!

PR #657, opened 2021-05-23
wsgi server: allow binding to ipv6 address

### LINKED ISSUE #567: start_http_server incapable to listen on IPv6 sockets
This is a very similar issue to #132 and #69.

#69 was closed as a bug in cPython. That bug was resolved in cPython 3.8, however `client_python` has since migrated to WSGI.

This is how it all behaves with the latest Python (3.8.5):
```
# python3.8
Python 3.8.5 (default, Jul 23 2020, 07:58:41)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import prometheus_client
>>> prometheus_client.start_http_server(port=9097, addr='::')
  File "/usr/local/lib/python3.8/site-packages/prometheus_client/exposition.py", line 78, in start_wsgi_server
    httpd = make_server(addr, port, app, ThreadingWSGIServer, handler_class=_SilentHandler)
  File "/usr/local/lib/python3.8/wsgiref/simple_server.py", line 154, in make_server
    server = server_class((host, port), handler_class)
  File "/usr/local/lib/python3.8/socketserver.py", line 452, in __init__
    self.server_bind()
  File "/usr/local/lib/python3.8/wsgiref/simple_server.py", line 50, in server_bind
    HTTPServer.server_bind(self)
  File "/usr/local/lib/python3.8/http/server.py", line 138, in server_bind
    socketserver.TCPServer.server_bind(self)
  File "/usr/local/lib/python3.8/socketserver.py", line 466, in server_bind
    self.socket.bind(self.server_address)
socket.gaierror: [Errno -9] Address family for hostname not supported
```

The issue lies with the `ThreadingWSGIServer` class, which defaults to `AF_INET` address family.
A workaround similar to https://github.com/bottlepy/bottle/commit/af547f495d673a910b1c70d9ff36c2cddafeb22d can be implemented in `exposition.py`:
```
# diff -u /usr/local/lib/python3.8/site-packages/prometheus_client/exposition.py.broken /usr/local/lib/python3.8/site-packages/prometheus_client/exposition.py
--- /usr/local/lib/python3.8/site-packages/prometheus_client/exposition.py.broken	2020-07-23 11:28:12.198472090 +0000
+++ /usr/local/lib/python3.8/site-packages/prometheus_client/exposition.py	2020-07-23 11:29:07.506269752 +0000
@@ -75,7 +75,12 @@
 def start_wsgi_server(port, addr='', registry=REGISTRY):
     """Starts a WSGI server for prometheus metrics as a daemon thread."""
     app = make_wsgi_app(registry)
-    httpd = make_server(addr, port, app, ThreadingWSGIServer, handler_class=_SilentHandler)
+    server_cls = ThreadingWSGIServer
+    if ':' in addr: # Fix wsgiref for IPv6 addresses.
+        if getattr(server_cls, 'address_family') == socket.AF_INET:
+            class server_cls(server_cls):
+                address_family = socket.AF_INET6
+    httpd = make_server(addr, port, app, server_cls, handler_class=_SilentHandler)
     t = threading.Thread(target=httpd.serve_forever)
     t.daemon = True
     t.start()
```

And voila, now it works as expected:
```
# python3.8
Python 3.8.5 (default, Jul 23 2020, 07:58:41)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import prometheus_client
>>> prometheus_client.start_http_server(port=9097, addr='::')
>>> prometheus_client.start_http_server(port=9098)
>>>
```
```
# netstat -ltnp
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
(...)
tcp        0      0 0.0.0.0:9098            0.0.0.0:*               LISTEN      20262/python3.8
(...)
tcp6       0      0 :::9097                 :::*                    LISTEN      20262/python3.8
(...)
```

This is a non-invasive fix which doesn't break any existing code while at the same time provides those of us who wish our exporters listened on IPv6 sockets.

Is it possible to adjust `exposition.py` this way, please? Thank you.

martinetd:
Reproduce the fix in python3.8 http.server by doing a dns lookup and
using the correct address family

Fixes #567

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -138,10 +138,24 @@ class ThreadingWSGIServer(ThreadingMixIn, WSGIServer):
     daemon_threads = True
 
 
-def start_wsgi_server(port, addr='', registry=REGISTRY):
+def _get_best_family(address, port):
+    """Automatically select address family depending on address"""
+    # HTTPServer defaults to AF_INET, which will not start properly if
+    # binding an ipv6 address is requested.
+    # This function is based on what upstream python did for http.server
+    # in https://github.com/python/cpython/pull/11767
+    infos = socket.getaddrinfo(address, port)
+    family, _, _, _, sockaddr = next(iter(infos))
+    return family, sockaddr[0]
+
+
+def start_wsgi_server(port, addr='0.0.0.0', registry=REGISTRY):
     """Starts a WSGI server for prometheus metrics as a daemon thread."""
+    class TmpServer(ThreadingWSGIServer):
+        """Copy of ThreadingWSGIServer to update address_family locally"""
+    TmpServer.address_family, addr = _get_best_family(addr, port)
     app = make_wsgi_app(registry)
-    httpd = make_server(addr, port, app, ThreadingWSGIServer, handler_class=_SilentHandler)
+    httpd = make_server(addr, port, app, TmpServer, handler_class=_SilentHandler)
     t = threading.Thread(target=httpd.serve_forever)
     t.daemon = True
     t.start()


martinetd:
Not sure what policy here is with squashes, I'd rather keep the python 2 compat commit separate but ask if you want me to squash it down.

For pyhton3.4 it's no longer installable on my distro and I cannot see circeci error without logging in, granting way more permissions that I'm comfortable with, so if someone would be kind enough to give me the error I'll try to fix it too... I tried to make it close to what's described on python doc but to no avail :/

csmarchbanks:
Thanks, I will check this out when I get a moment, but will need to catch up on some of the context in discussed in the related issue.

Don't worry about the 3.4 failure, it is a known issue right now due to a bad version of a 3rd party library used in tests being pushed. That version was then yanked from pypi, but the tests are still pulling it for some reason. Specifically the issue is [attrs 21.1.0](https://pypi.org/project/attrs/21.1.0/) does not work on 3.4.

martinetd:
Thanks, no rush.

I recap'd what I had understood of the issue in the last post of the issue if that's good enough for you, but there are at least 3 other similar issues on the project so information is spread a bit everywhere with various links to python bugs that ultimately didn't really help or workarounds that aren't appropriate for merging or looked too kludgy (forcing AF_INET6 or detecting ipv6 by looking for colon character)
This PR does what python3.8 http.server does when invoked from command-line so hopefully it's good enough :)

prometheus_client/exposition.py
```diff
@@ -143,9 +143,16 @@ class ThreadingWSGIServer(ThreadingMixIn, WSGIServer):
...
+def _get_best_family(address, port):
+    infos = socket.getaddrinfo(address, port)
+    family, _, _, _, sockaddr = next(iter(infos))
+    return family, sockaddr[0]
+
+
 def start_wsgi_server(port, addr='', registry=REGISTRY):
     """Starts a WSGI server for prometheus metrics as a daemon thread."""
     app = make_wsgi_app(registry)
+    ThreadingWSGIServer.address_family, addr = _get_best_family(addr, port)
```
csmarchbanks, line 155: Rather than update the address_family on the class, could we make a new class variable in this function and change it on that?

csmarchbanks (commented):
Thanks for the work! Generally I am ok with this approach, however I ran into a couple of issues in testing (linux only so far):
1. It does not work if I do not specify an address. It tries to look up empty string as a DNS record.
2. If I specify 'localhost' as my address, it may bind to IPv6, which means users going to 127.0.0.1 would no longer reach the service. I wonder if we should default to IPv4 if available?

prometheus_client/exposition.py
martinetd: I just did what http/server.py does as a good monkey but they certainly don't care about what kind of side effects they have for `python -m http` invocations.
I'm however not familiar with how this part of python works -- I just tried assigning the class to a new variable and that doesn't copy it e.g. this doesn't work:
```
      LocalServer = ThreadingWSGIServer
      LocalServer.address_family, addr = _get_best_family(addr, port)
      print(addr, LocalServer.address_family, ThreadingWSGIServer.address_family)
      httpd = make_server(addr, port, app, LocalServer, handler_class=_SilentHandler)
```
will print AF_INET6 twice or AF_INET twice depending on what was set.

stackoverflow has some horrible things with type() but I'm not sure this is worth the effort/risk unless you know how to do that reliably

martinetd:
> It does not work if I do not specify an address. It tries to look up empty string as a DNS record.

Ah, good point we should make addr=0.0.0.0 by default to keep the behaviour (I think it's that currently? need to check)
Interestingly for http.server it defaults to None, which with socket.AI_PASSIVE resolves as both 0.0.0.0 and :: in that order keeping only 0.0.0.0... But passing it for python2 (which doesn't allow kwargs) would mean something like `socket.getaddrinfo(address, port, 0, socket.SOCK_STREAM, 0, socket.AI_PASSIVE)`... Well, I guess argument meaning shouldn't change; but just changing the default addr value seems simpler.

> If I specify 'localhost' as my address, it may bind to IPv6, which means users going to 127.0.0.1 would no longer reach the service. I wonder if we should default to IPv4 if available?

Hm, that will definitely bind to ::1 on most systems yes; this is more annoying.
In general I'd really rather default to ipv6, that's a behaviour configurable at OS level and something everyone should be used to by now, having it the other way around would be surprising for new users... vs. breakage for old users (it's not just localhost, if they were using a name with both local A and AAAAs it would previously default to the A) so keeping old behaviour...
Ultimately it's up to you but I guess minimizing breakage is more important? I'll wait for your decision on this.

prometheus_client/exposition.py
csmarchbanks: :(, I will try taking a look when I get a chance. My fear with this solution is if someone sets up multiple servers the address_family would be shared between them and cause issues just as you saw here.

csmarchbanks:
> Well, I guess argument meaning shouldn't change; but just changing the default addr value seems simpler.

Could we check for empty string for addr and just default to IPv4 for that case without going through `_get_best_family`?

> Ultimately it's up to you but I guess minimizing breakage is more important? I'll wait for your decision on this.

Yeah, typically minimizing breakage is a goal I have, but if it makes a solution too complicated some breakage is ok for edge cases like this. The address_family issue in the other thread is more of a concern to me.

prometheus_client/exposition.py
martinetd: I've just checked though calling `start_http_server` with different address families in a row works fine, e.g.:
```
    start_http_server(8000, addr='::1')
    start_http_server(8001, addr='127.0.0.1')
```
will work and listen to both as expected

Since these are started off the main thread in most cases, we could just always restore the class address_family to AF_INET afer we're done creating the server -- would that be acceptable?
If subservers are created in another thread though this would be racy with whatever happens in other threads, but I don't think that's quite common.. Not much way of knowing though.

csmarchbanks: Having a library that effectively mutates global state just does not feel right to me. Restoring the address_family might be ok, but as you point out the thread safety that worries me. Certainly it is an edge case, but would be hard to debug if something does go wrong. I will think about this some more/try some things when I have time.

csmarchbanks (commented):
0.0.0.0 as a default sounds good to me. Thanks for all the work, I need a bit of time to think if I am ok with the potential to have some thread safety issues/if anything else could work. The address_family is the only concern I have remaining though.

martinetd:
Hello,

it's been a little while and I had totally forgotten about it until I upgraded the machine I use this on and my kludge got overwritten. Is there anything I can do to help move forward?

I appreciate that modifying the global `ThreadingWSGIServer` class is bad from a helper position, but unless we figure a way of "deep copying" the class I really don't see any other way about it. My suggestion of restoring the original value is as good as I can do.

csmarchbanks:
I have been thinking about this, and sadly just don't feel like there is a great solution right now :(. I don't want a library to be changing classes, so I guess the recommendation as it stands would be to use something else to stand up the server for IPv6. If I think of something else I will let you know though!

martinetd:
Hello, thanks for not forgetting about this! :)

My problem is that I'm not a direct user of this library, but just using exporters which do -- I can't very well go about telling the prometheur tor exporter, dht22 exporter and all other uses that they must use something else because my setup requires ipv6 and it's not useable with this client, so it's much easier to patch this (as I do on my systems).


I've just had another look and it looks like we can just instantiate a new subclass locally, and that won't affect the parent class, so this is probably a better workaround than saving the old family and restoring it. How about this new version of the patch?

I've tested running twice in a row with different families of address and making sure the previous call didn't affect the current instance version.

prometheus_client/exposition.py
```diff
@@ -138,10 +138,19 @@ class ThreadingWSGIServer(ThreadingMixIn, WSGIServer):
     daemon_threads = True
 
 
-def start_wsgi_server(port, addr='', registry=REGISTRY):
+def _get_best_family(address, port):
```
csmarchbanks, line 141: Would you add a comment for this function explaining why it is needed and that it is based on https://github.com/python/cpython/pull/11767 just so there is some easy to reference context in the future?

csmarchbanks (commented):
:+1: Good idea! I tested this a bit and am happy with it. One small comment just so that there is some context for future use.

prometheus_client/exposition.py
martinetd: Sure thing, just force pushed an update!

csmarchbanks (approved):

martinetd:
Thanks!

csmarchbanks:
Thank you, and good idea to copy the class!

PR #658, opened 2021-05-26
Document implications of negative observations

### LINKED ISSUE #638: Document implications of negative observations for Summaries and Histograms
See https://github.com/prometheus/client_golang/issues/850 for details.

I'll create documentation on prometheus/docs, so that we only have to link to it.

beorn7:
Fixes #638.

It's of course weird to talk abou negative values of an
"amount". However, for the Gauge, which quite regularly will receive
negative increments, the word "amount" is used, too, so I didn't want
to change it everywhere to "value" or something.

AFAIK the only place where negative values are rejected is indeed the
Counter.

Signed-off-by: beorn7 <beorn@grafana.com>

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -441,7 +441,15 @@ class Summary(MetricWrapperBase):
         self._created = time.time()
 
     def observe(self, amount):
-        """Observe the given amount."""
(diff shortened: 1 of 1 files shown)


csmarchbanks (approved):
Thanks!

PR #664, opened 2021-06-04
Allow counters to emit exemplars via OpenMetrics

csmarchbanks:
The parsing code already contains tests for parsing exemplars on counters, just the exposition is affected.

@brian-brazil as a member of OpenMetrics would you be willing to review this?

### DIFF
--- a/prometheus_client/openmetrics/exposition.py
+++ b/prometheus_client/openmetrics/exposition.py
@@ -8,6 +8,14 @@ CONTENT_TYPE_LATEST = str('application/openmetrics-text; version=0.0.1; charset=
 """Content type of the latest OpenMetrics text format"""
 
 
+def _is_valid_exemplar_metric(metric, sample):
+    if metric.type == 'counter' and sample.name.endswith('_total'):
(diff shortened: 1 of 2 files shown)


prometheus_client/openmetrics/exposition.py
```diff
@@ -28,8 +36,8 @@ def generate_latest(registry):
                 else:
                     labelstr = ''
                 if s.exemplar:
-                    if metric.type not in ('histogram', 'gaugehistogram') or not s.name.endswith('_bucket'):
-                        raise ValueError("Metric {0} has exemplars, but is not a histogram bucket".format(metric.name))
+                    if not _is_valid_exemplar_metric(metric, s):
+                        raise ValueError("Metric {0} has exemplars, but is not a counter or histogram bucket".format(metric.name))
```
brian-brazil, line 40: I'd say histogram bucket or counter

Otherwise it could be misread that counter buckets are a thing.

brian-brazil (commented):
:+1:

PR #666, opened 2021-06-10
Raise the correct exception if the metrics are not observable

### LINKED ISSUE #665: No error on missing labels for Gauge.set_function()
If I try to assign a function to a Gauge using `set_function()`, the Prometheus library does not error even if the labels are missing.

This code should error, but it does not:

```python
import prometheus_client

processed_input_messages = prometheus_client.Gauge(
    name="my_gauge",
    documentation="A generic Gauge.",
    labelnames=["a", "b"],
)

if __name__ == '__main__':
    # This should fail, because the "a" and "b" labels are not provided!
    processed_input_messages.set_function(lambda: 1)
```

In general, it would be important to check, for every metric type, if all the labels are set when calling `.set()`, `.inc()`, `.set_function()`, and the other similar metrics.

pietrodn:
Always check that a metric is observable before changing its value. Fixes #665.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -19,6 +19,21 @@ except ImportError:
     import unittest
 
 
+def assert_not_observable(fn, *args, **kwargs):
+    """
+    Assert that a function call falls with a ValueError exception containing
+    'missing label values'
+    """
+
+    try:
+        fn(*args, **kwargs)
+    except ValueError as e:
+        assert 'missing label values' in str(e)
+        return
+
+    assert False, "Did not raise a 'missing label values' exception"
+
+
 class TestCounter(unittest.TestCase):
     def setUp(self):
         self.registry = CollectorRegistry()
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -100,6 +117,21 @@ def test_gauge(self):
...
+
+        assert_not_observable(self.gauge_with_label.inc)
+
+    def test_dec_not_observable(self):
+        """.dec() must fail if the gauge is not observable."""
+
+        assert_not_observable(self.gauge_with_label.dec)
+
+    def test_set_not_observable(self):
+        """.dec() must fail if the gauge is not observable."""
```
daquinoaldo, line 131:
```suggestion
        """.set() must fail if the gauge is not observable."""
```

ocardia (approved):

tests/test_core.py
pietrodn: Fixed

daquinoaldo (approved):

csmarchbanks:
Hello, thank you for your PR and issue! I am taking a few days of time off, but plan to review it next week.

csmarchbanks (approved):
Thanks! Most of these would error already, but a message like `ValueError: counter metric is missing label values` is much nicer than `AttributeError: 'Counter' object has no attribute '_value'`.

PR #675, opened 2021-06-24
[#674] Made the registry restriction handle newly added metrics.

### LINKED ISSUE #674: Registering a metric after creating RestrictedRegistry with its name does not add the metric to RestrictedRegistry
The method `restricted_registry` works by filtering all metrics once at the time of its creation, as one can see from looking at the file `registry.py`.

If you register a metric after creating the restriction for it, it is not added to the restricted registry - while the docstring could be interpreted as claiming otherwise.

Solution 1: the class could be documented in a way that makes its behaviour explicit;
Solution 2: the class could be expanded to recollect metrics from its parent on every `collect()` call.

pavel-lexyr:
Fixes #674.

### DIFF
--- a/prometheus_client/metrics_core.py
+++ b/prometheus_client/metrics_core.py
@@ -58,7 +58,7 @@ class Metric(object):
             self.samples,
         )
 
-    def restricted_metric(self, names):
+    def _restricted_metric(self, names):
         """Build a snapshot of a metric with samples restricted to a given set of names."""
         samples = [s for s in self.samples if s[0] in names]
         if samples:
--- a/prometheus_client/metrics_core.py
+++ b/prometheus_client/metrics_core.py
@@ -58,6 +58,15 @@ class Metric(object):
             self.samples,
         )
 
+    def restricted_metric(self, names):
+        """Build a snapshot of a metric with samples restricted to a given set of names."""
+        samples = [s for s in self.samples if s[0] in names]
+        if samples:
+            m = Metric(self.name, self.documentation, self.type)
+            m.samples = samples
+            return m
+        return None
+
 
 class UnknownMetricFamily(Metric):
     """A single unknown metric and its samples.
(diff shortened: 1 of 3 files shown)


prometheus_client/metrics_core.py
```diff
@@ -58,6 +58,15 @@ def __repr__(self):
             self.samples,
         )
 
+    def restricted_metric(self, names):
```
csmarchbanks, line 61: Do you think there is a good use case for using this function externally? If not perhaps it should be `_restricted_metric` so we could easily change/remove it later.

csmarchbanks (commented):
Generally :+1: with one comment. Thanks!

prometheus_client/metrics_core.py
pavel-lexyr: Fixed. Good catch!

csmarchbanks (approved):
Thanks!

prometheus_client/registry.py
```diff
@@ -94,28 +94,7 @@ def restricted_registry(self, names):
...
-        collectors = set()
-        metrics = []
-        with self._lock:
-            if 'target_info' in names and self._target_info:
-                metrics.append(self._target_info_metric())
-                names.remove('target_info')
-            for name in names:
-                if name in self._names_to_collectors:
-                    collectors.add(self._names_to_collectors[name])
-        for collector in collectors:
```
brian-brazil, line 106: Removing this code has caused a serious performance regression. The whole point of this feature is not to collect from collectors that aren't needed, so that it's very cheap to access just one or two collectors.

csmarchbanks: OK, thank you for the context.

@pavel-lexyr perhaps let's approach this by improving the documentation, would you be willing to open a PR to change this back but with more complete documentation?

pavel-lexyr: Addressed your concerns in #680. Let me know if you find the implementation agreeable.

Issue #662, opened 2021-06-02
Error "Status must be a string" happened with Python2.7


Travis-Sun:
Traceback (most recent call last):
  File "/usr/lib/python2.7/wsgiref/handlers.py", line 85, in run
    self.result = application(self.environ, self.start_response)
  File "build/bdist.linux-x86_64/egg/prometheus_client/exposition.py", line 125, in prometheus_app
    start_response(status, [header])
  File "/usr/lib/python2.7/wsgiref/handlers.py", line 173, in start_response
    assert type(status) is StringType,"Status must be a string"
AssertionError: Status must be a string

csmarchbanks:
Thank you for the report, do you have a small example that reproduces this issue?

Geethree:
Yeah I just hit this in py2.7 as well. This appears to be a py2/py3 string issue.
In py2 [status](https://github.com/prometheus/client_python/blob/master/prometheus_client/exposition.py#L125) should be encoded and in py3 it should be a string.

I worked around this for another thing by just casting `status` to `str(status)`

Simple reproduced issue here: https://github.com/Geethree/prometheus-py27-string-bug

I'll open a pull request

PR #679, opened 2021-07-01
Python2.7 - AssertionError: Status must be a string

Geethree:
# TL;DR
As noted #662. When running in python2.7 when requesting an endpoint, wsgi returns:
```shell
prometheus_1  | Traceback (most recent call last):
prometheus_1  |   File "/usr/local/lib/python2.7/wsgiref/handlers.py", line 85, in run
prometheus_1  |     self.result = application(self.environ, self.start_response)
prometheus_1  |   File "/usr/local/lib/python2.7/site-packages/prometheus_client/exposition.py", line 125, in prometheus_app
prometheus_1  |     start_response(status, [header])
prometheus_1  |   File "/usr/local/lib/python2.7/wsgiref/handlers.py", line 173, in start_response
prometheus_1  |     assert type(status) is StringType,"Status must be a string"
prometheus_1  | AssertionError: Status must be a string
```
This is reproduced here: https://github.com/Geethree/prometheus-py27-string-bug

I'm not really sure of:
- If this is really the best way to get around this issue, happy to chat about it!
- The best way to construct a test to catch the AssertionError in the Daemon thread, happy to modify if someone can toss me a few tips I'd be happy to write some tests =)


Geethree:
@csmarchbanks @Travis-Sun =)

prometheus_client/exposition.py
```diff
@@ -122,7 +122,7 @@ def prometheus_app(environ, start_response):
             # Bake output
             status, header, output = _bake_output(registry, accept_header, params)
         # Return output
-        start_response(status, [header])
+        start_response(str(status), [header])
```
csmarchbanks, line 125: I wonder if lines 118 and 119 should just be cast as `str` similar to what the end of `_bake_output` does? Do you only see this issue for requests going to `/favicon.ico` or other locations as well?

csmarchbanks (commented):
Thank you for the fix! One comment, I am wondering if headers should also be cast as str, which is not happening on 119.

Would you be willing to add a test for this?

PR #680, opened 2021-07-01
[#674, #675] Mitigate performance regression by filtering restricted registry collections

pavel-lexyr:
#675 introduces a significant performance regression in restricted registry collecting. This attempts to reduce the regression by re-introducing filter code removed in the previous version.

### DIFF
--- a/prometheus_client/registry.py
+++ b/prometheus_client/registry.py
@@ -135,18 +135,16 @@ class RestrictedRegistry(object):
         self._registry = registry
 
     def collect(self):
-        names = copy.copy(self._name_set)
         collectors = set()
-        yield_target_info = False
+        target_info_metric = None
         with self._registry._lock:
-            if 'target_info' in names and self._registry._target_info:
-                yield_target_info = True
-                names.remove('target_info')
-            for name in names:
-                if name in self._registry._names_to_collectors:
+            if 'target_info' in self._name_set and self._registry._target_info:
+                target_info_metric = self._registry._target_info_metric()
+            for name in self._name_set:
+                if name != 'target_info' and name in self._registry._names_to_collectors:
                     collectors.add(self._registry._names_to_collectors[name])
-        if yield_target_info:
-            yield self._registry._target_info_metric()
+        if target_info_metric:
+            yield target_info_metric
         for collector in collectors:
             for metric in collector.collect():
                 m = metric._restricted_metric(self._name_set)
--- a/prometheus_client/registry.py
+++ b/prometheus_client/registry.py
@@ -135,10 +135,20 @@ class RestrictedRegistry(object):
         self._registry = registry
 
     def collect(self):
-        for metric in self._registry.collect():
-            m = metric._restricted_metric(self._name_set)
-            if m:
-                yield m
+        names = copy.copy(self._name_set)
+        collectors = set()
+        with self._registry._lock:
+            if 'target_info' in names and self._registry._target_info:
+                yield self._registry._target_info_metric()
+                names.remove('target_info')
+            for name in names:
+                if name in self._registry._names_to_collectors:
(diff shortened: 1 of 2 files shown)


pavel-lexyr:
py2.7 passes locally. What do the logs say on the CI?

csmarchbanks:
```
==================================== ERRORS ====================================
_____________________ ERROR collecting tests/test_core.py ______________________
ImportError while importing test module '/home/circleci/project/tests/test_core.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_core.py:5: in <module>
    from unittest.mock import MagicMock
E   ImportError: No module named mock
```

You can use a skipIf and import only for the test such as done in https://github.com/prometheus/client_python/blob/master/tests/openmetrics/test_parser.py#L492.

pavel-lexyr:
Aha. That worked - thank you!

csmarchbanks (approved):
This approach looks good to me, and I appreciate the test you added. I will give some time for @brian-brazil to review as well.

prometheus_client/registry.py
```diff
@@ -135,10 +135,20 @@ def __init__(self, names, registry):
         self._registry = registry
 
     def collect(self):
-        for metric in self._registry.collect():
-            m = metric._restricted_metric(self._name_set)
-            if m:
-                yield m
+        names = copy.copy(self._name_set)
```
brian-brazil, line 138: Why is this copy here? This set shouldn't be changing after the object is instantiated.

prometheus_client/registry.py
```diff
@@ -135,10 +135,20 @@ def __init__(self, names, registry):
...
+        collectors = set()
+        with self._registry._lock:
+            if 'target_info' in names and self._registry._target_info:
+                yield self._registry._target_info_metric()
```
brian-brazil, line 142: I don't think it is safe to yield under the lock.

prometheus_client/registry.py
pavel-lexyr: Trying to do it in the same way collections used to be filtered with target_info. There used to be a set variable that was modified if target_info was found - in order not to change the class field in the same way, we copy it to `names`.

prometheus_client/registry.py
pavel-lexyr: Fixed. Good catch!

prometheus_client/registry.py
```diff
@@ -135,10 +135,23 @@ def __init__(self, names, registry):
...
-            m = metric._restricted_metric(self._name_set)
-            if m:
-                yield m
+        names = copy.copy(self._name_set)
+        collectors = set()
+        yield_target_info = False
+        with self._registry._lock:
+            if 'target_info' in names and self._registry._target_info:
+                yield_target_info = True
+                names.remove('target_info')
```
brian-brazil, line 144: You don't need this, the registry should be preventing this from happening in the first place.

prometheus_client/registry.py
```diff
@@ -135,10 +135,23 @@ def __init__(self, names, registry):
         self._registry = registry
 
     def collect(self):
-        for metric in self._registry.collect():
-            m = metric._restricted_metric(self._name_set)
-            if m:
-                yield m
+        names = copy.copy(self._name_set)
+        collectors = set()
+        yield_target_info = False
```
brian-brazil, line 140: I'd have this as the thing to yield, rather than a bool for cleanliness. Also avoids a race.

pavel-lexyr: Done.

prometheus_client/registry.py
pavel-lexyr: Not sure I understood it right - but done.

brian-brazil:
:+1:

csmarchbanks (approved):
Thanks for the additional review and fixes!

PR #687, opened 2021-08-17
Fix instance_ip_grouping_key not working on macOS

### LINKED ISSUE #629: `instance_ip_grouping_key` does not work on macOS
macOS 11.1
Python 3.8.5
prometheus-client 0.9.0

To reproduce:
```python
from prometheus_client import instance_ip_grouping_key

instance_ip_grouping_key()
```

What happens?
```
metrics.py:10: in <module>
    INSTANCE = instance_ip_grouping_key()
lib/python3.8/site-packages/prometheus_client/exposition.py:379: in instance_ip_grouping_key
    s.connect(('localhost', 0))
E   OSError: [Errno 49] Can't assign requested address
```

AceFire6:
Fixes #629

The solution is adapted from this StackOverflow answer: https://stackoverflow.com/a/28950776

@csmarchbanks I decided I'd just make the PR so long and if the solution isn't acceptable it can just be closed 👍

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -481,7 +481,16 @@ def _escape_grouping_key(k, v):
 def instance_ip_grouping_key():
     """Grouping key with instance set to the IP Address of this host."""
     with closing(socket.socket(socket.AF_INET, socket.SOCK_DGRAM)) as s:
-        s.connect(('10.255.255.255', 1))
+        if sys.platform == 'darwin':
+            # This check is done this way only on MacOS devices
+            # it is done this way because the localhost method does
+            # not work.
+            # This method was adapted from this StackOverflow answer:
+            # https://stackoverflow.com/a/28950776
+            s.connect(('10.255.255.255', 1))
+        else:
+            s.connect(('localhost', 0))
+
         return {'instance': s.getsockname()[0]}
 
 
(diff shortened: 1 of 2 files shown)


csmarchbanks:
Thank you for the PR, I just got back from a couple weeks of vacation, and will take a look at this soon. Just need to make sure I fully understand the workaround.

csmarchbanks (commented):
I apologize for the delay. I spent some time testing this, and my main concern is that it looks like the result could be different  with this change. E.g. on my local machine `instance_ip_grouping_key()` changes from `{'instance': '127.0.0.1'}` to `{'instance': '192.168.1.107'}`. For many cases the new result probably makes more sense, but this could be a breaking change if someone is configuring their servers with a specific IP address.

AceFire6:
Ah okay. That isn't good!

I wasn't really sure how to test it and we didn't end up using the patch upstream as we had intended to when I made this PR.

I guess this doesn't fit the bill then. I'm not really sure where to go from here so I guess I'll leave what happens next up to you.

Thanks for the review 👍

csmarchbanks:
One option would be to check if the platform is Darwin and use your new version in that case, but keep the old behavior for all other platforms? If there is a comment for why the two are handled slightly differently I would be happy moving forward with that.

AceFire6:
@csmarchbanks Cool, I can change the PR to do that 👍

AceFire6:
@csmarchbanks I've made the change to choose which method is used based on the platform 👍

Issue #568, opened 2020-07-24
High cpu when using Gunicorn with multiprocess


onzo-mateuszzakarczemny:
I run flask app on Gunicorn 20.0.4 and Kubernetes. My python is v3.7.5. I have a problem with increasing CPU over a time.
![image](https://user-images.githubusercontent.com/29543308/88379047-7e903f00-cda2-11ea-8fbf-0e5b22b082a8.png)
After investigating stack traces using py-spy I noticed that the issue is caused by prometheus metrics lib.
It doesn't clean up metric files from old workers. Therefore over a time merge process take more and more CPU. 

After deleting following files the CPU usages dropped significantly:
```
ll -lh
total 7.0M
drwxr-xr-x 2 root root 4.0K Jul 24 06:13 ./
drwxrwxrwt 1 root root 4.0K Jul 19 14:14 ../
-rw-r--r-- 1 root root 1.0M Jul 24 08:14 counter_106.db
-rw-r--r-- 1 root root 1.0M Jul 23 18:41 counter_112.db
-rw-r--r-- 1 root root 1.0M Jul 24 04:07 counter_118.db
-rw-r--r-- 1 root root 1.0M Jul 24 04:54 counter_136.db
-rw-r--r-- 1 root root 1.0M Jul 24 08:40 counter_142.db
-rw-r--r-- 1 root root 1.0M Jul 20 16:44 counter_16.db
-rw-r--r-- 1 root root 1.0M Jul 20 11:24 counter_17.db
-rw-r--r-- 1 root root 1.0M Jul 21 01:40 counter_18.db
-rw-r--r-- 1 root root 1.0M Jul 21 20:14 counter_40.db
-rw-r--r-- 1 root root 1.0M Jul 21 17:17 counter_52.db
-rw-r--r-- 1 root root 1.0M Jul 21 21:29 counter_58.db
-rw-r--r-- 1 root root 1.0M Jul 23 07:19 counter_70.db
-rw-r--r-- 1 root root 1.0M Jul 22 19:49 counter_82.db
-rw-r--r-- 1 root root 1.0M Jul 22 18:59 counter_88.db
-rw-r--r-- 1 root root 1.0M Jul 24 08:43 histogram_106.db
-rw-r--r-- 1 root root 1.0M Jul 24 04:15 histogram_112.db
-rw-r--r-- 1 root root 1.0M Jul 24 05:02 histogram_118.db
[...]

PR #694, opened 2021-09-13
Create multiprocess files with process uuid

ethervoid:
### **This is a draft open to discussion**

In order to be able to clean-up the process files once it's stopped or destroyed, we need to identify those files with a UUID. If not, those files are going to be processed and collected and will keep growing until we restart the server or master process to perform the cleanup of the metrics directory

This change uses the psutil library that gives a hash based on PID+Process creation time in order to differentiate two processes with the same PID but those are different.

I have some questions about the feasibility of this change:

- Is the approach ok? Mainly is respectful in case of exception and even though some files could be left behind is not the same situation as we have now. Also, there is an improvement that I can add and is marking those files I couldn't delete as "type_pid_uuid_defunct" so we can do something to directly remove those files
- Is it ok to start using an external library? I've seen you're trying to avoid it
- Tests are failing because the test process is not running and the unittest.mock library is only from 3.3 it's ok to move all the multiprocess tests to required that version to run?

This change aims to fix https://github.com/prometheus/client_python/issues/568

// @csmarchbanks

Signed-off-by: Mario de Frutos <mario@defrutos.org>

### DIFF
--- a/prometheus_client/multiprocess.py
+++ b/prometheus_client/multiprocess.py
@@ -6,10 +6,12 @@ import json
 import os
 import warnings
 
+import psutil
+
 from .metrics_core import Metric
 from .mmap_dict import MmapedDict
 from .samples import Sample
-from .utils import floatToGoString
+from .utils import floatToGoString, get_process_data
 
 try:  # Python3
     FileNotFoundError
@@ -158,7 +160,18 @@ def mark_process_dead(pid, path=None):
     """Do bookkeeping for when one process dies in a multi-process setup."""
     if path is None:
         path = os.environ.get('PROMETHEUS_MULTIPROC_DIR', os.environ.get('prometheus_multiproc_dir'))
-    for f in glob.glob(os.path.join(path, 'gauge_livesum_{0}.db'.format(pid))):
-        os.remove(f)
-    for f in glob.glob(os.path.join(path, 'gauge_liveall_{0}.db'.format(pid))):
-        os.remove(f)
+    try:
+        process = psutil.Process(pid)
+        types = ["counter", "histogram", "summary", "gauge_min", "gauge_max", "gauge_livesum", "gauge_liveall"]
+        for typ in types:
+            for f in glob.glob(os.path.join(path, '{0}_{1}_{2}.db'.format(typ, pid, process.__hash__()))):
+                os.remove(f)
+    except (psutil.NoSuchProcess, psutil.AccessDenied):
+        # We couldn't found the dead process, no further action required because with the uuid of the process there is
+        # no much we can do
+        pass
+    except OSError:
+        # Error deleting some of the files, for example in windows if they're in use they'll raise this error
+        # Maybe we can rename them as _defunk if needed to do a future cleanup
+        pass
+
--- a/prometheus_client/mmap_dict.py
+++ b/prometheus_client/mmap_dict.py
@@ -134,6 +134,13 @@ class MmapedDict(object):
             self._m.close()
             self._m = None
             self._f.close()
+            try:
+                # Given that we're using uuid for the process we can safely
+                # remove the file because is not going to be shared by other process
+                os.remove(self._f.name)
+            except OSError:
+                # In windows if the file is in use raises an error
+                pass
             self._f = None
 
 
--- a/prometheus_client/utils.py
+++ b/prometheus_client/utils.py
@@ -1,4 +1,7 @@
 import math
+import sys
+
+import psutil
 
 INF = float("inf")
 MINUS_INF = float("-inf")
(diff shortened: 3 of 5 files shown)


csmarchbanks (commented):
Thank you for the PR. I think the idea behind this (identifying a process based on PID & start time) could work, however, as you mention, one of the goals of this project is to not require external dependencies. I wonder if you could generate a start time (even if approximate) without the psutil library?

prometheus_client/multiprocess.py
```diff
@@ -158,7 +160,18 @@ def mark_process_dead(pid, path=None):
...
         path = os.environ.get('PROMETHEUS_MULTIPROC_DIR', os.environ.get('prometheus_multiproc_dir'))
-    for f in glob.glob(os.path.join(path, 'gauge_livesum_{0}.db'.format(pid))):
-        os.remove(f)
-    for f in glob.glob(os.path.join(path, 'gauge_liveall_{0}.db'.format(pid))):
-        os.remove(f)
+    try:
+        process = psutil.Process(pid)
+        types = ["counter", "histogram", "summary", "gauge_min", "gauge_max", "gauge_livesum", "gauge_liveall"]
+        for typ in types:
+            for f in glob.glob(os.path.join(path, '{0}_{1}_{2}.db'.format(typ, pid, process.__hash__()))):
```
roidelapluie, line 167: is this safe? Wouldn't this reset some counters partially, therefore producing bad data?

roidelapluie:
```suggestion
            for f in glob.glob(os.path.join(path, '{0}_{1}_{2}.db'.format(typ, pid, hash(process)))):
```

Isn't hash() the pythonic way to get a hash?

roidelapluie:
I'd be very cautious with this pull request. This raises some questions, notably the need for all the try/except.

prometheus_client/multiprocess.py
ethervoid: This should only affect files for a destroyed/stopped worker. I can see the point of not being able to scrape the data because we're removing the files so we can use the other approach which is to mark them as defunct and do the cleanup after a number of minutes/hours.

I'm old school haha I'll change to hash instead of __hash__

ethervoid:
@roidelapluie yep, that's why it's a draft and not a proper PR.I can see the point but we need to be sure the process data is still there because we're relying on the process creation time to differentiate two processes with the same PID.

ethervoid:
@csmarchbanks yeah we can do it but we need, for performance reasons, to do a native part of the code and has to be multi-environment but that could lead to a increased risk for this PR because it has to deal with the code for the library and also the native C code to get the creation_time for the process

prometheus_client/multiprocess.py
csmarchbanks: The problem is that if you just remove an old file that contained say 100 increments of a counter then when you go to read all of the files during the next scrape your new value will be 100 lower which `rate` and other counter functions in prometheus interpret as a reset.

Anything with a counter component (histograms, though not gauge histograms, sum and count fields on a summary, etc...), will need to be merged and have those old increments kept around somewhere. Ideally in one file instead of the current behavior of keeping more and more files on disk as PIDs churn.

ethervoid: I see, thank you for the explanation, but shouldn't be that scenario happening when the server is restarted and all the files from the metrics folder are wiped? I thought Prometheus server, once the scrape is done, takes care of it and doesn't need the data in the target

csmarchbanks: The problem with the current code happens when a process is stopped and cleaned up, but the server is not wiped. You then end up in a state where there is a partial reset (counter value is lowered just by the amount of requests handled by the now stopped and cleaned up process) which can cause large issues with `rate` calculations.

How a partial reset is handled is that if a counter goes from 100 to 80 on cleanup (partial reset) it is assumed by Prometheus that the counter went from 100 to 0 to 80, causing a big spike during that period of time, when actually no requests may have come through.

ethervoid: I see, I knew about the reset stuff but I thought Prometheus doesn't take the disappeared metric as a 0 but just as disappeared and didn't take into account. Not very much into the internals of Prometheus yet.

So, what would be the suggestion to clean up those files safely? We're removing the files if there is a fork when the mmap object is closed but when the process is dead we can only mark them as defunct and then do the clean up later or something like that? That wouldn't fix the problem because in environments with a high process recycling they're going to hit the same issue.

csmarchbanks: One idea is to merge the values into a single file, for example https://github.com/prometheus/client_python/pull/518 (which is unsafe and has not been updated in awhile). It would be necessary to make sure there are not race conditions or other issues with any new approach.

ethervoid: I see. Then this goes beyond my current knowledge of prometheus. I'll close this pR then. Thank you for your time

Issue #491, opened 2019-12-03
Type hint stubs


paravoid:
It'd be awesome if client_python was type hinted, as that would allow downstream consumers to typecheck their API calls (one has to explicitly `# type: ignore` the import right now).

My understanding is that the project is actively trying to support (very) old versions of Python. In this case, there are two options: a) type hints in comments b) type stubs (`.pyi`). The latter is preferrable; this can be shipped by either [typeshed](https://github.com/python/typeshed), or, ideally, [in the software itself](https://mypy.readthedocs.io/en/latest/installed_packages.html). For more about the pros and cons of each solution, I've found Bernat Gabor's [blog post](https://www.bernat.tech/the-state-of-type-hints-in-python/) to be informative.

Adding the hints across the entire code base is easy but still, not a trivial amount of work. Tools like [MonkeyType](https://github.com/Instagram/MonkeyType/issues) can help bootstrap that process. Given the good code test coverage, I tried running the test suite with MonkeyType (`monkeytype run -m pytest`) and it generated the majority of the hints. There were, however, some corner cases that it wasn't able to figure out or otherwise failed to generate, and I'm not familiar enough with the codebase to help with that.

Thanks & thanks for all of the work you've been putting on this module!

brian-brazil:
This seems reasonable to add to the public bits of the API via comments.

> My understanding is that the project is actively trying to support (very) old versions of Python.

More that we're not going to drop them until they cause too much hassle, so far it has only been a little annoying.

akx:
#718 dropped support for Python 2 altogether, so real type hints could now be added.

paravoid:
I believe type hints now exist, so resolving this. Thanks!

PR #705, opened 2021-10-09
Annotate few functions and methods

takeda:
This is related to https://github.com/prometheus/client_python/issues/491, it doesn't resolve it, but it could be a step into that direction.

This theoretically should not break python 2.7

### DIFF
--- a/prometheus_client/context_managers.py
+++ b/prometheus_client/context_managers.py
@@ -1,21 +1,30 @@
 from timeit import default_timer
+from types import TracebackType
+from typing import (
+    Any, Callable, Literal, Optional, Type, TYPE_CHECKING, TypeVar,
+)
 
 from .decorator import decorate
 
+if TYPE_CHECKING:
+    from . import Counter
+    F = TypeVar("F", bound=Callable[..., Any])
+
 
 class ExceptionCounter:
-    def __init__(self, counter, exception):
+    def __init__(self, counter: "Counter", exception: Type[BaseException]) -> None:
         self._counter = counter
         self._exception = exception
 
-    def __enter__(self):
+    def __enter__(self) -> None:
         pass
 
-    def __exit__(self, typ, value, traceback):
+    def __exit__(self, typ: Optional[Type[BaseException]], value: Optional[BaseException], traceback: Optional[TracebackType]) -> Literal[False]:
         if isinstance(value, self._exception):
             self._counter.inc()
+        return False
 
-    def __call__(self, f):
+    def __call__(self, f: "F") -> "F":
         def wrapped(func, *args, **kwargs):
             with self:
                 return func(*args, **kwargs)
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -1,6 +1,9 @@
 from threading import Lock
 import time
 import types
+from typing import (
+    Any, Callable, Dict, Optional, Sequence, Tuple, Type, TypeVar,
+)
 
 from . import values  # retain this import style for testability
 from .context_managers import ExceptionCounter, InprogressTracker, Timer
@@ -8,9 +11,12 @@ from .metrics_core import (
     Metric, METRIC_LABEL_NAME_RE, METRIC_NAME_RE,
     RESERVED_METRIC_LABEL_NAME_RE,
 )
-from .registry import REGISTRY
+from .registry import CollectorRegistry, REGISTRY
 from .samples import Exemplar
 from .utils import floatToGoString, INF
+
+T = TypeVar('T', bound='MetricWrapperBase')
+F = TypeVar("F", bound=Callable[..., Any])
 
 
 def _build_full_name(metric_type, name, namespace, subsystem, unit):
@@ -56,8 +62,8 @@ def _validate_exemplar(exemplar):
 
 
 class MetricWrapperBase:
-    _type = None
-    _reserved_labelnames = ()
+    _type: Optional[str] = None
+    _reserved_labelnames: Sequence[str] = ()
 
     def _is_observable(self):
         # Whether this metric is observable, i.e.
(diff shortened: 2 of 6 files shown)


takeda:
@csmarchbanks is there a chance to get this merged? Should I update this patch to resolve current conflicts?

csmarchbanks:
Hello, sorry, I missed this one! Thank you for the PR. I am hoping to review https://github.com/prometheus/client_python/pull/718 this week, and merge it soon. At that point I believe we can support full typing instead of just the annotations. Would it make sense to wait until after #718 is merged?

takeda:
Yeah, that's fine.

takeda:
@csmarchbanks I rebased my code and adapted it to use annotations as 3.6+ is now required. I currently added types that my code is utilizing. This can be used to show how to add types to other parts of the project.

BTW: since types are not used when the code is run and adding typing involves time, the types could be added incrementally over time.

prometheus_client/context_managers.py
```diff
@@ -1,21 +1,30 @@
...
 from .decorator import decorate
 
+if TYPE_CHECKING:
+    from . import Counter
+    F = TypeVar("F", bound=Callable[..., Any])
+
 
 class ExceptionCounter:
-    def __init__(self, counter, exception):
+    def __init__(self, counter: "Counter", exception: Type[BaseException]) -> None:
```
csmarchbanks, line 15: :thinking: Is there any downside to importing `Counter` all of the time and keeping this type without quotes similar to other cases? To use this function the user will have to import Counter anyway.

prometheus_client/metrics.py
```diff
@@ -1,17 +1,23 @@
...
 from .metrics_core import (
     Metric, METRIC_LABEL_NAME_RE, METRIC_NAME_RE,
     RESERVED_METRIC_LABEL_NAME_RE,
 )
-from .registry import REGISTRY
+from .registry import CollectorRegistry, REGISTRY
 from .samples import Exemplar
 from .utils import floatToGoString, INF
 
+T = TypeVar('T', bound='MetricWrapperBase')
```
csmarchbanks, line 18: Rather than using `T` could we just use `"MetricWrapperBase"` to make everything clearer later on when reading the functions?

prometheus_client/metrics.py
```diff
@@ -213,6 +219,7 @@ def _multi_samples(self):
                 yield (suffix, dict(series_labels + list(sample_labels.items())), value, timestamp, exemplar)
 
     def _child_samples(self):  # pragma: no cover
+        # type: () -> Sequence[Tuple[str, Dict[str, str], float]]
```
csmarchbanks, line 222: This no longer needs to be a comment.

prometheus_client/samples.py
```diff
@@ -36,8 +37,15 @@ def __gt__(self, other):
 # Timestamp can be a float containing a unixtime in seconds,
 # a Timestamp object, or None.
 # Exemplar can be an Exemplar object, or None.
-Sample = namedtuple('Sample', ['name', 'labels', 'value', 'timestamp', 'exemplar'])
-Sample.__new__.__defaults__ = (None, None)
-
-Exemplar = namedtuple('Exemplar', ['labels', 'value', 'timestamp'])
-Exemplar.__new__.__defaults__ = (None,)
+class Exemplar(NamedTuple):
+    labels: Mapping[str, str]
```
csmarchbanks, line 41: We should probably use `Dict` here to be consistent with other label sets we use in for example `inc` on a counter. Alternatively we could have a custom type for label sets that is `Dict[str, str]`?

csmarchbanks (commented):
Awesome, thanks for the update. Generally this is looking pretty good, I left a handful of questions/comments.

prometheus_client/context_managers.py
csmarchbanks: Ohh wait, nevermind. Circular import, resolving this conversation.

prometheus_client/metrics.py
takeda: That wouldn't work the same way.

This type is used as a generic, it's basically bound to the class and ensures that type returned is of the same type as the mothod's class. Probably I might not even need the bound parameter here.

prometheus_client/metrics.py
takeda: Ooops, will fix it.

prometheus_client/samples.py
takeda: I can change it to `Dict` I'm not as familiar with the code so I put `Mapping` just in case there's a different type that act like a dict.

takeda: Do you think that type will be used frequently? If so, maybe it would be warranted to declare it specifically. For now I just changed it to `Dict[]`

prometheus_client/metrics.py
csmarchbanks: Ahh, yeah that makes sense. Thanks!

prometheus_client/samples.py
csmarchbanks: I think just a Dict for now makes sense, it would be easiest for users rather than navigating to another type definition.

prometheus_client/decorator.py
```diff
@@ -39,6 +39,9 @@
 import operator
 import re
 import sys
+from typing import Any, Callable, TypeVar
```
csmarchbanks, line 42: This file is actually a copy of http://pypi.python.org/pypi/decorator, you can see the file header for more details. That does mean we should not edit it so as to make future upgrades easier.

csmarchbanks (commented):
One small comment that I missed the first time around, otherwise :+1:. Thanks!

prometheus_client/decorator.py
takeda: I removed it, but out of curiosity. Why is this needed and why contextmanager is used?

For example:

```python
class ExceptionCounter:
    def __init__(self, counter: "Counter", exception: Type[BaseException]) -> None:
        self._counter = counter
        self._exception = exception

    def __enter__(self) -> None:
        pass

    def __exit__(self, typ: Optional[Type[BaseException]], value: Optional[BaseException], traceback: Optional[TracebackType]) -> Literal[False]:
        if isinstance(value, self._exception):
            self._counter.inc()
        return False

    def __call__(self, f: "F") -> "F":
        def wrapped(func, *args, **kwargs):
            with self:
                return func(*args, **kwargs)

        return decorate(f, wrapped)
```

could be:

```python
@contextmanager
def exception_counter(counter: "Counter", exception: Type[BaseException]) -> None:
  yield
  if isinstance(value, self._exception):
    self._counter.inc()
```

csmarchbanks: You can find the history of the vendored in file in the PR that added it, as well as the linked issue: https://github.com/prometheus/client_python/pull/91.

I think you probably could restructure the various context managers like you suggest now that we do not support older python versions. I believe that before 3.2 just using `@contextmanger` would not allow you to use it as a decorator as well so the class was nicer.

csmarchbanks (approved):
Thanks!

Issue #535, opened 2020-04-10
Async context manager


denisSurkov:
Usecase:

```
c = Counter("errors_count")

@c.count_exceptions()
def foo():  # counter = 1
      raise Exception()

@c.count_exceptions()
async def foo_async(): # counter still 1
      await sleep(1)
      raise Exception()
```

Quick search in project did nothing to me. 

It would be cool if counter (and other objects) would handle this right out of the box, especially if we're talking about metrics for web world.

brian-brazil:
How would that work, keeping in mind that we support all the way back to 2.6?

denisSurkov:
Well, I didn't think about it. Sorry.

LittleLampLight:
@brian-brazil, what do you think about Python version check inside decorator to exclude support for non-async syntax. Seems like it's not gonna affect backwards support. And will come really handy

brian-brazil:
I'm not sure how that'd work given that it's a syntax thing, and thus 2.6 would choke on it.

LittleLampLight:
Yup, actually i was thinking about depending on version import of async compatible syntax. But it looks too complex and such case may be implemented easier with context manager as well. 

```python3
c = Counter("errors_count")

async def foo_async():
    with c.count_exceptions():
        await sleep(1)
        raise Exception()
```

actually if someone really need the decorator then he can implement it by himself with a few lines of code

MSeal:
The silent failure is surprising here.

> How would that work, keeping in mind that we support all the way back to 2.6?

I believe it's possible to do version checks inside the __call__, then do a dynamic import for async to do the async check against the function passed in to avoid making code calls that aren't present in 2.6.

e.g. https://stackoverflow.com/questions/51344906/in-python-how-to-use-decorator-compatible-with-normal-function-and-coroutine-fun

> actually if someone really need the decorator then he can implement it by himself with a few lines of code

Sure, but it's unexpected and causes metrics to be easily missed in a Python 3 world where 2.x is past EOL.

nikicat:
Why don't just drop support for not-maintained-for-7-years Python 2.6? I vote for native asyncio support instead of ancient py2.x.

csmarchbanks:
I pulled some data for downloads over the last 6 months by python version:

Python Version | Downloads | Percent of Downloads
-- | -- | --
3.7 | 74784972 | 69.80%
3.6 | 15260503 | 14.24%
3.8 | 6414421 | 5.99%
[...]

PR #706, opened 2021-10-13
Add support for Python 3.10

hugovk:
Python 3.10 was released last week:

* https://discuss.python.org/t/python-3-10-0-is-now-available/10955

We can also remove some redundant Python 2.6 code.

### DIFF
--- a/tests/test_asgi.py
+++ b/tests/test_asgi.py
@@ -1,11 +1,9 @@
 from __future__ import absolute_import, unicode_literals
 
-from unittest import TestCase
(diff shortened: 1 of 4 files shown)


csmarchbanks:
Yep, I have been thinking about EOL for 2 as well. Last time I had pulled numbers was in February, which had closer to 5% usage: https://github.com/prometheus/client_python/issues/535#issuecomment-784386056. Usage dropping below 2% certainly makes deprecation seem reasonable (as well as more and more projects and AWS discontinuing support).

Issue #717, opened 2021-11-01
End Support for Python 2.7


csmarchbanks:
Python 2.7 usage has dropped significantly over the last 6-9 months, and over the last 3 months makes up less than 2% of downloads. At this point I think it is worth dropping support and allowing new features as well as some cleanup.

Current Python 2.7 users would be able to continue to use v0.12.x and earlier, and patches could be added to the v0.12.x release series if absolutely necessary.

If you have concerns (or support), please voice them here.

### 2021-08-01 - 2021-10-31

| category | percent | downloads  |
| -------- | ------: | ---------: |
| 3        |  96.57% | 77,679,427 |
| 2        |   1.98% |  1,596,557 |
| null     |   1.45% |  1,165,008 |
| Total    |         | 80,440,992 |

### 2021-10-01 - 2021-10-31

| category | percent | downloads  |
| -------- | ------: | ---------: |
| 3        |  97.39% | 30,050,858 |
| null     |   1.30% |    402,541 |
| 2        |   1.30% |    401,660 |
| Total    |         | 30,855,059 |

Compiled from 
```
pypistats python_major prometheus_client
```

roidelapluie:
Will we be able to mark is as not supporting 2.7, and will 2.7 users still be able to keep using the last version?

csmarchbanks:
Yes, for example users of Python 2.6 are currently still able to use https://pypi.org/project/prometheus-client/0.9.0/, but https://pypi.org/project/prometheus-client/0.10.0/ requires Python>=2.7.

roidelapluie:
I do not have specific concerns then, dropping 2.7 sounds good to me.

PR #718, opened 2021-11-02
Drop support for EOL Python 2.7 and 3.4-3.5

hugovk:
Fixes https://github.com/prometheus/client_python/issues/717.

Follow on from https://github.com/prometheus/client_python/pull/706#issuecomment-954858387.

Python 2.7 and 3.4 - 3.5 are EOL:

| cycle | latest |  release   |    eol     |
|:------|:-------|:----------:|:----------:|
| 3.6   | 3.6.15 | 2016-12-23 | 2021-12-23 |
| 3.5   | 3.5.10 | 2015-09-30 | 2020-09-13 |
| 3.4   | 3.4.10 | 2014-03-16 | 2019-03-18 |
| 2.7   | 2.7.18 | 2010-07-03 | 2020-01-01 |

https://endoflife.date/python

Here's the pip installs for prometheus_client from PyPI for October 2021 showing under 2% usage:

| category | percent |  downloads |
|:---------|--------:|-----------:|
| 3.7      |  72.89% | 22,490,512 |
| 3.8      |  14.20% |  4,380,172 |
| 3.6      |   6.85% |  2,113,567 |
| 3.9      |   2.93% |    904,146 |
| null     |   1.30% |    402,541 |
| 2.7      |   1.30% |    401,660 |
| 3.5      |   0.27% |     82,532 |
| 3.10     |   0.26% |     79,007 |
| 3.4      |   0.00% |        736 |
| 3.11     |   0.00% |        184 |
| 3.3      |   0.00% |          2 |
| Total    |         | 30,855,059 |

Source: `pip install -U pypistats && pypistats python_minor prometheus_client --last-month`

In fact, 2.7 + 3.4 + 3.5 usage combined is under 2% = 1.57%.

Also upgraded Python syntax with https://github.com/asottile/pyupgrade `--py36-plus`, and removed universal wheels, only needed for suporting both Python 2 and 3.


csmarchbanks:
Thanks! This will close https://github.com/prometheus/client_python/issues/717 (just making a reference for the future).

I posted that to the -dev mailing list yesterday, and am waiting to hear if there are significant concerns from the community before moving forward with review/merge on this. If I don't hear anything within a week at the most we can certainly move forward.

csmarchbanks:
I haven't heard any concerns on the other issue, so I plan to review this sometime this week!

README.md
```diff
@@ -1,6 +1,6 @@
 # Prometheus Python Client
 
-The official Python 2 and 3 client for [Prometheus](http://prometheus.io).
+The official Python 3 client for [Prometheus](https://prometheus.io).
```
csmarchbanks, line 3: Do you think we could just call this `The official Python client` rather than specifying version?

prometheus_client/decorator.py
```diff
@@ -283,15 +251,7 @@ def __call__(self, func):
...
-if n_args == 2 and not init.varargs:  # (self, genobj) Python 2.7
-    def __init__(self, g, *a, **k):
-        return _GeneratorContextManager.__init__(self, g(*a, **k))
-
-
-    ContextManager.__init__ = __init__
-elif n_args == 2 and init.varargs:  # (self, gen, *a, **k) Python 3.4
-    pass
-elif n_args == 4:  # (self, gen, args, kwds) Python 3.5
+if n_args == 4:  # (self, gen, args, kwds)
```
csmarchbanks, line 254: Is this if necessary anymore, or would it always be true in Python 3.5+?

csmarchbanks (commented):
This looks awesome, thanks for all your work! I left a couple comments/questions, but they are pretty minor. I might try to get a second set of eyes to make sure I am not missing any python 2/3 conversion differences.

README.md
hugovk: Good idea, will update.

prometheus_client/decorator.py
hugovk: Good question, because it was originally an `elif` and not `else`, I kept it as `if` to be on the cautious side and maintain the original functionality.

This is calculated from a "private/internal" `contextlib._GeneratorContextManagerBase` but at least it's still true in Python 3.10 and and and 3.11.0a2+.

csmarchbanks: Hmm, this file is just copied from https://github.com/micheles/decorator, so we should probably not modify it manually, but upgrade instead. In more recent versions this logic is gone completely, which makes sense as we aren't actually changing `__init__`. For this PR we could just leave the file alone, then upgrade in a separate PR?

hugovk: Good idea, I've rebased it out.

An idea for the followup: you could put the file in a `_vendor` directory to make it clear it's an external file (for example, pip and setuptools do this).

csmarchbanks: Thanks for rebasing it out! I like the vendor idea, it is called out in the NOTICE file right now, but that would make it even more clear.

csmarchbanks (approved):
:+1: from me, double checking with a second reviewer, then will merge. Thanks again!

PR #722, opened 2021-11-09
Remove trailing slash from pushgateway URLs

romuald:
When using the push gateway, URLs may be submitted as host or URL.

When using an URL scheme, in case user submits URL with a trailing slash
(eg: `http://localhost:9091/`), the pushgateway server will reply with a
301 (or 307 depending on the version) redirect that is not handled by
the default handler. Leading to a traceback

Simply remove any trailing slash from the base URL


csmarchbanks (approved):
Thanks!

PR #726, opened 2021-11-16
Catch non-integer bucket/count values.

brian-brazil:
@roidelapluie

### DIFF
--- a/prometheus_client/openmetrics/parser.py
+++ b/prometheus_client/openmetrics/parser.py
@@ -562,10 +562,16 @@ def text_fd_to_metric_families(fd):
 
             if typ == 'stateset' and name not in sample.labels:
                 raise ValueError("Stateset missing label: " + line)
-            if (typ in ['histogram', 'gaugehistogram'] and name + '_bucket' == sample.name
+            if (name + '_bucket' == sample.name
                     and (sample.labels.get('le', "NaN") == "NaN"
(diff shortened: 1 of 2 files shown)


prometheus_client/openmetrics/parser.py
```diff
@@ -562,10 +562,16 @@ def build_metric(name, documentation, typ, unit, samples):
 
             if typ == 'stateset' and name not in sample.labels:
                 raise ValueError("Stateset missing label: " + line)
-            if (typ in ['histogram', 'gaugehistogram'] and name + '_bucket' == sample.name
+            if (name + '_bucket' == sample.name
```
csmarchbanks, line 565: I am curious here around removing the type for this check? Looking at the OM spec it is only a SHOULD that `_bucket`/`_count` series are not used by other types so I would have expected the type checks to remain for all of these.

brian-brazil: It is redundant, as it is a MUST that they're not used by others as a suffix so any deviations will be caught elsewhere. That bit of the spec is around naming metrics, not use of suffixes.

csmarchbanks: Ahh right, they would be potential clashes, that makes sense. Thanks.

csmarchbanks (approved):

roidelapluie (approved):
Thanks!

PR #734, opened 2021-12-06
Add a quantity parameter to Histogram.observe

### LINKED ISSUE #733: Parameterize the number of observations to make in Histogram.observe
`Histogram.observe` finds a bucket and then increments its value by 1.  Sometimes I want to observe a value greater than 1.  I could do this with a for loop:

```
for i in range(N):
  metric.observe(x)
```

if I don't mind my metrics making everything slow but I do.  Instead, it would be lovely to be able to write:

```
metric.observe(x, quantity=N)
```

exarkun:
Fixes #733

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -581,7 +581,7 @@ class Histogram(MetricWrapperBase):
                 self._labelvalues + (floatToGoString(b),))
             )
 
-    def observe(self, amount, exemplar=None):
+    def observe(self, amount, exemplar=None, quantity=1):
         """Observe the given amount.
 
         The amount is usually positive or zero. Negative values are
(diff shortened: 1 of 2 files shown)


csmarchbanks (commented):
Hello, thank you for your issue and PR! I think I am ok with the change, but before merging I am curious what your use case is? At first it seems a bit strange to me to have a large amount of observations all with the same value.

exarkun:
> Hello, thank you for your issue and PR! I think I am ok with the change, but before merging I am curious what your use case is? At first it seems a bit strange to me to have a large amount of observations all with the same value.

Thanks.

I came up with this while trying to implement a histogram for part of an accounting system.  I wanted the histogram to place observations into buckets based on price and scale the quantity of the observation by the cost to deliver the service.

I admit that I'm not entirely confident this is going to produce a useful histogram... but it seems like a reasonable thing to be able to ask `client_python` to do (without a very expensive for loop).

sdfordham:
Similar to @csmarchbanks I cannot think of a use case for this and adding new parameters to base metrics observe methods should have a much higher need threshold than demonstrated here imo. Your particular use case would be better served with a custom collector.

csmarchbanks:
Interesting, I don't think scaling the quantity based on price would create a valid histogram, but it is an interesting use case. Two histograms one for price and one for cost might be more correct?

For now I would prefer not to expand the parameters, but if a use case/performance issue does show up I would certainly reconsider.

exarkun:
> Interesting, I don't think scaling the quantity based on price would create a valid histogram, but it is an interesting use case. Two histograms one for price and one for cost might be more correct?
> 
> For now I would prefer not to expand the parameters, but if a use case/performance issue does show up I would certainly reconsider.

Okay.  I don't have any additional arguments to make in favor of adding this.

PR #740, opened 2021-12-20
Allow adding arbitrary labels to metrics when generating output

willnewton:
It's useful to be able to apply a label across all metrics, for
example when using a multi-processing framework to apply a
label for worker ID to all metrics (including from builtin
collector types). This change alows passing a dictionary of
labels and values to generate_latest which will then be applied
to all generated metrics.

Signed-off-by: Will Newton <will.newton@gmail.com>

Let me know if this approach looks viable to you, happy to rework it if you have suggestions for how to do it in a better way.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -135,15 +135,16 @@ def start_wsgi_server(port, addr='', registry=REGISTRY):
 start_http_server = start_wsgi_server
 
 
-def generate_latest(registry=REGISTRY):
+def generate_latest(registry=REGISTRY, extra_labels={}):
     """Returns the metrics from the registry in latest text format as a string."""
 
     def sample_line(line):
-        if line.labels:
+        labels = {**line.labels, **extra_labels}
+        if labels:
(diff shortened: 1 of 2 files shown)


prometheus_client/exposition.py
```diff
@@ -135,15 +135,16 @@ def start_wsgi_server(port, addr='', registry=REGISTRY):
 start_http_server = start_wsgi_server
 
 
-def generate_latest(registry=REGISTRY):
+def generate_latest(registry=REGISTRY, extra_labels={}):
```
csmarchbanks, line 138: If this is to be accepted we would also want this in `openmetrics/exposition.py`.

csmarchbanks (commented):
Hello, thank you for the PR.

I am open to a feature that allows extra labels to be added to all metrics, but I am not sure this is the correct way to do it. What would you think of providing a new type of registry that wraps a CollectorRegistry that will add new labels to each metric?

willnewton:
I opened a new PR #741 that adds a new collector type. It's a little more complex but conceptually cleaner that way I think.

PR #747, opened 2022-01-13
Change to imports to fix go-to-declaration in editors

angusholder:
@csmarchbanks Just a small tweak to improve how prometheus_client behaves with IDEs/editors:

In PyCharm, when ctrl-clicking functions/classes of this library to go to their declaration, it takes me to `prometheus_client/__init__.py` instead of where the function/class was actually defined. This seems to happen because `__init__.py` uses assignments instead of imports to re-export members, so I've changed them all to imports.

### DIFF
--- a/prometheus_client/__init__.py
+++ b/prometheus_client/__init__.py
@@ -7,37 +7,37 @@ from . import (
 
 __all__ = ['Counter', 'Gauge', 'Summary', 'Histogram', 'Info', 'Enum']
 
-CollectorRegistry = registry.CollectorRegistry
-REGISTRY = registry.REGISTRY
-Metric = metrics_core.Metric
-Counter = metrics.Counter
-Gauge = metrics.Gauge
-Summary = metrics.Summary
-Histogram = metrics.Histogram
-Info = metrics.Info
-Enum = metrics.Enum
+from .registry import CollectorRegistry
+from .registry import REGISTRY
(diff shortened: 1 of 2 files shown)


prometheus_client/__init__.py
```diff
@@ -4,40 +4,21 @@
     exposition, gc_collector, metrics, metrics_core, platform_collector,
     process_collector, registry,
```
csmarchbanks, line 5: Are all of these imports still necessary?

csmarchbanks (commented):
Thanks! Just one small comment.

prometheus_client/__init__.py
angusholder: There's a chance it causes breakage for some people - currently you can access those packages without explicitly importing them, eg:

```python
import prometheus_client
prometheus_client.metrics.Gauge
```

but if we remove those imports from `__init__.py`, their code would break and need changes to their imports:

```python
import prometheus_client.metrics
prometheus_client.metrics.Gauge
```

Not that there's much reason for people to use those packages instead of the top level, but I thought it wasn't worth risking breakage

csmarchbanks: That's true. Hopefully it is unlikely too many people do that, but I agree there is no reason to break it.

csmarchbanks (approved):
Thanks!

PR #754, opened 2022-01-26
Declare `registry` on `MetricWrapperBase` as `Optional`

reivilibre:
Dear maintainer @csmarchbanks,

The release of 0.13.0 yesterday appears to have caused us some type checking errors with Mypy (presumably this is a recent effort to annotate this library with types — thanks!).

I think this particular annotation is incomplete though — our code relies on passing `None` here and the code beneath says the following:
```python
            if registry:
                registry.register(self)
```
In other words, I think this was intended to be optional.

Thanks!

Signed-off-by: Olivier Wilkinson (reivilibre) <oliverw@matrix.org>

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -677,7 +677,7 @@ class Enum(MetricWrapperBase):
                  namespace: str = '',
                  subsystem: str = '',
                  unit: str = '',
-                 registry: CollectorRegistry = REGISTRY,
+                 registry: Optional[CollectorRegistry] = REGISTRY,
                  _labelvalues: Optional[Sequence[str]] = None,
                  states: Optional[Sequence[str]] = None,
                  ):
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -107,7 +107,7 @@ class MetricWrapperBase:
                  namespace: str = '',
                  subsystem: str = '',
                  unit: str = '',
-                 registry: CollectorRegistry = REGISTRY,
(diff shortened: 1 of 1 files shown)


csmarchbanks (commented):
Thanks and good catch! When I look at your linked PR you are also using a custom proxy registry instead of a CollectorRegistry. Perhaps for now we should just leave out type checking on the registry until an abc or something else is created?

Would you also be able to fix the same issue in Enum in this file as well?

csmarchbanks:
FYI I am also going to allow `Any` as labelvalue types as we cast them to str anyway: https://github.com/prometheus/client_python/pull/755.

reivilibre:
> Perhaps for now we should just leave out type checking on the registry until a Protocol or abc is created?

I think it's fair enough to leave it as is (with the type annotations) — if people like us are so keen on doing dubious things then a `cast` is OK to work around the problem in the meantime (it's no worse than it being `Any` — but for people who *are* respecting the annotations, it provides a little bit more safety net).

> FYI I am also going to allow Any as labelvalue types as we cast them to str anyway

Thanks! I noticed they get cast to `str` internally but it's one of those things where you shouldn't second-guess the type annotations too much (though I suppose at this point, people are relying on the casts being there). It's good to have it stated explicitly, though — thanks for that (and for letting me know).

I've updated the PR to do the same for `Enum` :).

csmarchbanks (approved):

csmarchbanks:
That sounds good, thanks!

PR #756, opened 2022-01-26
Use Iterable[Sample] for all sample collection

csmarchbanks:
This makes sure the types are all consistent and use the appropriate
NamedTuple.


fkaleo (approved):

sd2k (commented):
I'm not super familiar with the codebase, but in general I would prefer the functions returning concrete types (`Tuple[Sample]`) to be annotated as such, unless there's a specific reason for not doing (e.g. they might change in future). Not a massive issue though, and this should be forwards compatible.

PR #757, opened 2022-01-27
Add missing functions/classes to __all__

Pliner:
Hi,

`mypy==0.931.0` raises the following errors for `prometheus-client==0.13.0`:

image

Probably the reason of it are two things:
1. #747, https://github.com/prometheus/client_python/pull/747/files#diff-719659a170701715fc787a12f9ceb16910e3d9a5fbe2c5725f91fac76b29f0f1L10, https://github.com/prometheus/client_python/pull/747/files#diff-719659a170701715fc787a12f9ceb16910e3d9a5fbe2c5725f91fac76b29f0f1L11
2. As I know mypy relies on `__all__` declaration and I have no idea why it worked before.

I hope it is fine to include these two things to `__all__`, because of their wider usage compared to other classes/constants mentioned in #747.

Thanks.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -17,6 +17,21 @@ from .openmetrics import exposition as openmetrics
 from .registry import REGISTRY
 from .utils import floatToGoString
 
+__all__ = (
+    'CONTENT_TYPE_LATEST',
+    'delete_from_gateway',
+    'generate_latest',
+    'instance_ip_grouping_key',
+    'make_asgi_app',
+    'make_wsgi_app',
+    'MetricsHandler',
+    'push_to_gateway',
+    'pushadd_to_gateway',
+    'start_http_server',
+    'start_wsgi_server',
+    'write_to_textfile',
+)
+
 CONTENT_TYPE_LATEST = 'text/plain; version=0.0.4; charset=utf-8'
 """Content type of the latest text format"""
 PYTHON376_OR_NEWER = sys.version_info > (3, 7, 5)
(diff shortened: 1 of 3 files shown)


Pliner:
@csmarchbanks Could you look on it please?

csmarchbanks (commented):
Thanks!

If this is an issue for you with those two functions, it will probably be an issue for others with some of the other functions, would you be willing to just add them all?

csmarchbanks:
Also, it would be nice if this would fail in our linter, perhaps as part of this PR you could add `--no-implicit-reexport` to the mypy step to avoid this happening in the future?

Pliner:
@csmarchbanks Done. 

Thanks for the fast feedback!

prometheus_client/exposition.py
```diff
@@ -17,6 +17,21 @@
 from .registry import REGISTRY
 from .utils import floatToGoString
 
+__all__ = (
```
Pliner, line 20: I'm a bit unsure about this, but had to explicitly export in this way because of `make_asgi_app`.

Pliner:
```
prometheus_client/__init__.py:7: error: Module "prometheus_client.exposition" does not explicitly export attribute "make_asgi_app"; implicit reexport disabled
```

csmarchbanks: This seems fine to me, it makes it very clear that those are the things we are exporting.

csmarchbanks (approved):
Awesome, thanks!

PR #759, opened 2022-01-28
Add typing to Histogram/Summary/Gauge

Pliner:
Hi,

It is a small addition to Histogram/Summary/Gauge typings. 

PS I suppose at the current stage it is fine to set disallow_incomplete_defs to True.

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -413,7 +413,7 @@ class Gauge(MetricWrapperBase):
 
         self._raise_if_not_observable()
 
-        def samples(self) -> Iterable[Sample]:
+        def samples(_: Gauge) -> Iterable[Sample]:
             return (Sample('', {}, float(f()), None, None),)
 
         self._child_samples = types.MethodType(samples, self)
@@ -530,15 +530,15 @@ class Histogram(MetricWrapperBase):
     DEFAULT_BUCKETS = (.005, .01, .025, .05, .075, .1, .25, .5, .75, 1.0, 2.5, 5.0, 7.5, 10.0, INF)
 
     def __init__(self,
-                 name,
-                 documentation,
-                 labelnames=(),
-                 namespace='',
-                 subsystem='',
-                 unit='',
-                 registry=REGISTRY,
-                 _labelvalues=None,
-                 buckets=DEFAULT_BUCKETS,
+                 name: str,
+                 documentation: str,
+                 labelnames: Iterable[str] = (),
+                 namespace: str = '',
+                 subsystem: str = '',
+                 unit: str = '',
+                 registry: Optional[CollectorRegistry] = REGISTRY,
+                 _labelvalues: Optional[Sequence[str]] = None,
+                 buckets: Sequence[float] = DEFAULT_BUCKETS,
                  ):
         self._prepare_buckets(buckets)
         super().__init__(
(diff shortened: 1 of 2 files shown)


prometheus_client/metrics.py
```diff
@@ -413,7 +413,7 @@ def set_function(self, f):
 
         self._raise_if_not_observable()
 
-        def samples(self) -> Iterable[Sample]:
+        def samples(_: Gauge) -> Iterable[Sample]:
```
Pliner, line 416: It is the only one thing which is needed to enable `disallow_incomplete_defs`.

prometheus_client/metrics.py
```diff
@@ -413,10 +413,10 @@ def set_function(self, f):
...
             return (Sample('', {}, float(f()), None, None),)
 
-        self._child_samples = types.MethodType(samples, self)
+        self._child_samples = types.MethodType(samples, self)  # type: ignore
```
Pliner, line 419: `mypy` is not fond of such an overwrite of a method

Pliner:
@csmarchbanks Could you look on it please? Many thanks.

prometheus_client/metrics.py
```diff
@@ -579,7 +579,7 @@ def _metric_init(self):
                 self._labelvalues + (floatToGoString(b),))
             )
 
-    def observe(self, amount, exemplar=None):
+    def observe(self, amount: float = 1, exemplar: Optional[Dict[str, str]] = None) -> None:
```
csmarchbanks, line 582: I believe the amount should not have a default value here.

csmarchbanks (commented):
Thanks, this is great!

I will take a closer look next week, but generally this is looking good to me.

prometheus_client/metrics.py
Pliner: Ooops, sorry

Pliner: Fixed

prometheus_client/metrics.py
```diff
@@ -553,7 +553,7 @@ def __init__(self,
         )
         self._kwargs['buckets'] = buckets
 
-    def _prepare_buckets(self, buckets):
+    def _prepare_buckets(self, buckets: Sequence[float]) -> None:
```
csmarchbanks, line 556: We specifically cast each element in buckets to a float on the next line. Should this be something like a union of int/float/str instead?

csmarchbanks (commented):
One small comment, otherwise I think this looks good!

prometheus_client/metrics.py
Pliner: I thought about the same :)

Pliner: I've push a fix.

prometheus_client/metrics.py
```diff
@@ -553,8 +553,8 @@ def __init__(self,
         )
         self._kwargs['buckets'] = buckets
 
-    def _prepare_buckets(self, buckets):
-        buckets = [float(b) for b in buckets]
+    def _prepare_buckets(self, source_buckets: Sequence[Union[float, int, str]]) -> None:
```
Pliner, line 556: I was to rename this param, because after the fix `buckets = [float(b) for b in source_buckets]` has different type.

csmarchbanks: Makes sense, it is an internal function so should be fine.

csmarchbanks (approved):
Awesome, thanks!

PR #768, opened 2022-02-09
mark_process_dead: avoid error with bad env

romuald:
When the multiprocess environemnt is not properly set-up (missing
environment variable) and mark_process_dead is called, the method raises
an error on path.join() because the path is None which may be
problematic

This commit simply ignores the cleanup when the variable is not present

### Context

We are using the same library for multiprocess and non-multiprocess setups.

While the setup is obviously wrong on our part (adding `mark_process_dead` using `atexit`). I thought that check in prometheus_client would be nice addition if anyone would get in the same pitfall


csmarchbanks (commented):
Thanks!

My initial reaction is that the error is actually a good thing though as it tells a user that they should fix their environment. At a minimum it seems like some sort of warning should be output?

Issue #760, opened 2022-01-31
Histogram is not typed, and due to partial typing, this makes mypy --strict fail


tommyjcarpenter:
Since prometheus_client provides some types, adding it to `ignore_missing_imports` doesnt work.  The existence of this file: https://github.com/prometheus/client_python/blob/master/prometheus_client/py.typed 
means that `ignore_missing_imports` is ignored with `prometheus_client`. 

Counter  `inc` is typed: 
https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics.py#L272-L280

However, Histogram isnt typed, including `observe`:
https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics.py#L464-L476

This causes `mypy --strict` to fail when using Histogram. Typing all metrics will allow strict mypy users to stop adding `# type: ignore` anywhere this function is used.

csmarchbanks:
https://github.com/prometheus/client_python/pull/759 should fix this issue for histograms. The goal with typing was to add types over time as contributions allowed. It sounds like that does not work well for users using `mypy --strict` though which is disappointing.

I believe to fully allow running `--strict` we will need to add types for everything documented in `prometheus_client/__init__.py` which can be a goal for 0.14.0.

tommyjcarpenter:
Thanks for the reply! Mypy will only complain about what users are using from the package, in my case, Counters and Histograms. So in my case, the fix for histograms Is enough to make - - strict happy - it doesn't introspect on types beyond what the user imports.

But outside of my own use case, yes - - strict means "complete typing" and any users using an unryped function will have strict fail.

tkukushkin:
You can try to check type completeness by pyright [tool](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#verifying-type-completeness). It would be nice to add this check to the CI.

tkukushkin:
In my experience Pyright shows more type problems. And it does not enforce you to cover private API and other implementation details, you can cover public API at first and cover private API gradually later.

Just in case, I am not talking about pyright as type checker, but only as type completeness checker. You can check types by mypy with your config (strict or not) and check type completeness by pyright. It will give you the best coverage of public API.

Perhaps @erictraut can tell us more accurately what the difference is.

erictraut:
[...]

PR #771, opened 2022-02-09
Add type annotations to everything referenced in __init__.py

csmarchbanks:
Fixes: #760 

This should cover most of the public surface area used by others. There is more typing work to be done before we can enable stricter types, but this will be a good next step.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -6,6 +6,7 @@ import socket
 from socketserver import ThreadingMixIn
 import sys
 import threading
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple
 from urllib.error import HTTPError
 from urllib.parse import parse_qs, quote_plus, urlparse
 from urllib.request import (
@@ -14,7 +15,7 @@ from urllib.request import (
 from wsgiref.simple_server import make_server, WSGIRequestHandler, WSGIServer
 
 from .openmetrics import exposition as openmetrics
-from .registry import REGISTRY
+from .registry import CollectorRegistry, REGISTRY
 from .utils import floatToGoString
 
 __all__ = (
@@ -101,7 +102,7 @@ def _bake_output(registry, accept_header, params):
     return '200 OK', ('Content-Type', content_type), output
 
 
-def make_wsgi_app(registry=REGISTRY):
+def make_wsgi_app(registry: CollectorRegistry = REGISTRY) -> Callable:
     """Create a WSGI app which serves the metrics from a registry."""
 
     def prometheus_app(environ, start_response):
@@ -149,7 +150,7 @@ def _get_best_family(address, port):
     return family, sockaddr[0]
 
 
-def start_wsgi_server(port, addr='0.0.0.0', registry=REGISTRY):
+def start_wsgi_server(port: int, addr: str = '0.0.0.0', registry: CollectorRegistry = REGISTRY) -> None:
     """Starts a WSGI server for prometheus metrics as a daemon thread."""
     class TmpServer(ThreadingWSGIServer):
         """Copy of ThreadingWSGIServer to update address_family locally"""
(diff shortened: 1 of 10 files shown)


prometheus_client/exposition.py
```diff
@@ -271,27 +272,34 @@ def factory(cls, registry):
...
         os.replace(tmppath, path)
     else:
         os.rename(tmppath, path)
 
 
-def _make_handler(url, method, timeout, headers, data, base_handler):
+def _make_handler(
+    url: str,
+    method: str,
+    timeout: Optional[float],
```
tommyjcarpenter, line 294: Seems fine to me, but most cases where I see `Optional`, there is typically a default of `=None`; but in this case I guess the caller of this is explicitly required to pass in an explicit `None`..

prometheus_client/exposition.py
```diff
@@ -343,8 +357,13 @@ def handle():
 
 
 def push_to_gateway(
-        gateway, job, registry, grouping_key=None, timeout=30,
-        handler=default_handler):
+    gateway: str,
+    job: str,
+    registry: CollectorRegistry,
+    grouping_key: Optional[Dict[str, str]] = None,
```
tommyjcarpenter, line 363: if grouping key is reused below, if it has a constant structure, may want to use a `TypedDict`

prometheus_client/exposition.py
```diff
@@ -343,8 +357,13 @@ def handle():
...
+    timeout: Optional[float] = 30,
```
tommyjcarpenter, line 364: here timeout has a default.

prometheus_client/metrics_core.py
```diff
@@ -169,10 +201,17 @@ def __init__(self, name, documentation, count_value=None, sum_value=None, labels
         if labels is None:
             labels = []
         self._labelnames = tuple(labels)
```
tommyjcarpenter, line 203: maybe `self._labelnames = tuple(labels) if labels else ()`

prometheus_client/samples.py
```diff
@@ -4,30 +4,30 @@
...
 
-    def __init__(self, sec, nsec):
+    def __init__(self, sec: Union[int, float], nsec: Union[int, float]) -> None:
         if nsec < 0 or nsec >= 1e9:
             raise ValueError(f"Invalid value for nanoseconds in Timestamp: {nsec}")
         if sec < 0:
             nsec = -nsec
-        self.sec = int(sec)
-        self.nsec = int(nsec)
+        self.sec: int = int(sec)
```
tommyjcarpenter, line 12: did mypy complain here when theres an explicit int() cast but it didnt infer the type fro that?

tommyjcarpenter (approved):
thanks so much!

prometheus_client/exposition.py
csmarchbanks: I don't think it it has a constant structure, and `TypedDict` isn't available earlier than 3.8 anyway :(. I really want to use `Protocol` in a few places as well but cannot yet.

prometheus_client/exposition.py
csmarchbanks: Yeah, this one is an internal and requires that we explicitly pass timeout though even if the value is None (or a default value) from one of the other handlers. A bit unusual I agree.

prometheus_client/metrics_core.py
csmarchbanks: Decent idea as a future cleanup, going to leave it out of this PR to keep the focus on types clear :).

prometheus_client/samples.py
csmarchbanks: Ahh, good catch. I think this was from a test with pyright which somehow didn't infer it from the cast. We aren't using pyright yet so I removed them.

csmarchbanks (commented):
Thanks for the review!

prometheus_client/exposition.py
tkukushkin: It looks like you can use typing-extensions for `TypedDict` and `Protocol`.

prometheus_client/samples.py
tkukushkin: It is a good idea to describe the types of public fields in the body of the class. Other users can use pyright for type checking their projects and pyright doesn't trying to infer types in such cases (`__init__` can be much more complicated), in py.typed project all public members should be type annotated.

csmarchbanks: Interesting, thanks for the link! I updated this code back to having explicit int types and looked through the other classes assigning to public variables in this PR and it looks like they are covered. Hopefully that helps out pyright users while we work on getting more coverage.

Issue #672, opened 2021-06-24
Add ability to turn off _created metrics


csmarchbanks:
As per the [June 24th 2021 Prometheus Dev Summit](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#heading=h.omrxayfajpwh) there is consensus that all first party clients should have a parameter to turn off _created metrics.

baryluk:
Any update on this?

`_created` metrics are not very useful (I personally never used them), and add to resource consumption through the monitoring stack.

Thanks.

csmarchbanks:
Hello, no progress has been made but I would be happy to accept a contribution to provide a flag.

csmarchbanks:
Did you set `PROMETHEUS_DISABLE_CREATED_SERIES=True` as well? This library tries to match as many SHOULD clauses as possible in OpenMetrics so the created series are on by default.

nidemidovich:
Yes, I did. I even tried set the variable before import `prometheus_client` as mentioned in #710

nidemidovich:
Any updates on this? @csmarchbanks Sorry for mentioning

csmarchbanks:
I haven't had time yet, I am hoping to try to reproduce today, but it might have to wait until next week.

baryluk:
I was using this flag for some time, and sometimes it does not work, even if I set it in-process before importing `prometheus_client`, it does not work as expected. I could envvar set it on a process startup, but it just clutters startup scripts, or hand started stuff. I do not have anything that imports it transitively before my import.

So, I hacked and just set internal variable:


```python3
#!/usr/bin/env python3

import dataclasses
import os
import sys
import time
from typing import Optional

os.environ["PROMETHEUS_DISABLE_CREATED_SERIES"] = "True"

import prometheus_client

prometheus_client.metrics._use_created = False

...
```

csmarchbanks:
Yeah, we need to make that variable settable in code. See https://github.com/prometheus/client_python/issues/933.

nidemidovich:
I'll try @baryluk hack for now. But hope you figure out how to solve the issue in better way. M

ay be reopen the issue?

csmarchbanks:
[...]

PR #774, opened 2022-02-16
Disable exporting of _created metrics via env var

csmarchbanks:
Users can disable the automatic creation of _created metrics by
specifying the environment variable PROMETHEUS_DISABLE_CREATED_SERIES.
This is controlled by an environment variable as an end user may not
have control over which registry is being used/when metrics are created.

Alternatively, we could force this to be done on the registry level and filter the metrics out
but that would not help users with custom registries. Happy to hear feedback that another
approach should be taken though.

Closes: #672

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,13 @@ h = Histogram('request_latency_seconds', 'Description of histogram')
 h.observe(4.7, {'trace_id': 'abc123'})
 ```
 
+### Disabling `_created` metrics
+
+By default counters, histograms, and summaries export an additional series
+suffixed with `_created` and a value of the unix timestamp for when the metric
+was created. If this information is not helpful, it can be disabled by setting
+the environment variable `PROMETHEUS_DISABLE_CREATED_SERIES=True`.
+
 ### Process Collector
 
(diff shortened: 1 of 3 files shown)


SuperQ:
It would be useful to have a way that wasn't an environmental side-effect. Something like an explicit default registry method to turn it off. This makes it easier to deploy to fleets of services that share common library patterns.

csmarchbanks:
I am happy to accept something like that. Making life easier for users both with or without control over the source code is always nice.

dswarbrick:
Another slight rough edge to this: the environment variable needs to be set _before_ importing `prometheus_client`, since the `_get_use_created()` is only checked once when the module-level `_use_created` variable is set.

This means that when trying to manipulate this option from within Python via `os.environ`, you run into a flake8 lint nag:
```
import os

# Disable automatic addition of _created series. Must be set before importing prometheus_client.
os.environ["PROMETHEUS_DISABLE_CREATED_SERIES"] = 'true'

from prometheus_client import CollectorRegistry, Counter, Gauge, generate_latest
```
```
E402 module level import not at top of file
```

Issue #775, opened 2022-02-17
Add support for GZIP encoding


ivan-valkov:
As described in the [expositions format spec](https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format), Prometheus supports the optional `Content-Encoding: gzip` header.

Other client libraries already support encoding the scraped metrics, for example:
* [client_golang](https://github.com/prometheus/client_golang/blob/34ca1203775369991cf963e1b71fe1f230064dc5/prometheus/promhttp/http.go#L157)
* [client_java](https://github.com/prometheus/client_java/blob/master/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java#L108)

It will be useful to support this in the python client as well to reduce bandwidth used by applications using the client for exporting metrics.

PR #776, opened 2022-02-17
Add support for gzip content-encoding

ivan-valkov:
Resolves https://github.com/prometheus/client_python/issues/775

If an `Accept-Encoding:gzip` header is present, will compress the output and set the correct `Content-Encoding` header in the response. Added tests for both the wsgi and asgi implementations.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -238,7 +238,7 @@ def generate_latest(registry: CollectorRegistry = REGISTRY) -> bytes:
 def choose_formatter(accept_header: str) -> Tuple[Callable[[CollectorRegistry], bytes], str]:
     accept_header = accept_header or ''
     for accepted in accept_header.split(','):
-        if accepted.split(';')[0].strip().lower() == 'application/openmetrics-text':
+        if accepted.split(';')[0].strip() == 'application/openmetrics-text':
             return (openmetrics.generate_latest,
                     openmetrics.CONTENT_TYPE_LATEST)
     return generate_latest, CONTENT_TYPE_LATEST
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -265,7 +265,7 @@ class MetricsHandler(BaseHTTPRequestHandler):
         accept_encoding_header = self.headers.get('Accept-Encoding')
         params = parse_qs(urlparse(self.path).query)
         # Bake output
-        status, headers, output = _bake_output(registry, accept_header, accept_encoding_header, params)
+        status, headers, output = _bake_output(registry, accept_header, accept_encoding_header, params, False)
         # Return output
         self.send_response(int(status.split(' ')[0]))
         for header in headers:
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -39,6 +39,13 @@ CONTENT_TYPE_LATEST = 'text/plain; version=0.0.4; charset=utf-8'
 PYTHON376_OR_NEWER = sys.version_info > (3, 7, 5)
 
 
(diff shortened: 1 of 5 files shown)


prometheus_client/exposition.py
```diff
@@ -227,15 +235,23 @@ def sample_line(line):
     return ''.join(output).encode('utf-8')
 
 
-def choose_encoder(accept_header: str) -> Tuple[Callable[[CollectorRegistry], bytes], str]:
+def choose_formatter(accept_header: str) -> Tuple[Callable[[CollectorRegistry], bytes], str]:
     accept_header = accept_header or ''
     for accepted in accept_header.split(','):
-        if accepted.split(';')[0].strip() == 'application/openmetrics-text':
+        if accepted.split(';')[0].strip().lower() == 'application/openmetrics-text':
```
csmarchbanks, line 241: Are these `.lower()` calls necessary?

csmarchbanks (commented):
Thanks! Generally this looks reasonable though I have not fully tested it yet.

One thing to note, the golang client allows disabling compression as part of the options passed to the handler. I am wondering if we should do something similar here? The reason being, in some cases the cpu usage required to compress a large response can slow everything down too much.

prometheus_client/exposition.py
brian-brazil: As this is the reference parser, it probably shouldn't accept incorrect input like this.

ivan-valkov: for `accept-encoding`, the value is case insensitive according to the spec - https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.2.1
I (wrongly?) assumed this will be the case for the `accept` header as well, but checking it now, it is undefined in the spec - https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2
If you think we shouldn't add the `lower()` here for the `accept` headers, I am fine to remove this. However, I believe we should keep the `lower()` when checking the `accept-encoding` header values.

ivan-valkov:
> One thing to note, the golang client allows disabling compression as part of the options passed to the handler. I am wondering if we should do something similar here? The reason being, in some cases the cpu usage required to compress a large response can slow everything down too much.

Yeah, definitely, this makes sense. What would be the preferred semantic? GZIP compression enabled or disabled by default? I am conscious that if we enable it by default, some existing users of the client might see unexpected CPU spikes when using the new version of the client, so maybe we should require enabling GZIP compression explicitly?

csmarchbanks (commented):
I would say let's default to negotiating gzip as most use cases will not see much of a CPU spike, and advanced users can turn it off. This also aligns us with client_golang (enabled by default) and client_java (always enabled).

prometheus_client/exposition.py
```diff
@@ -38,6 +39,13 @@
 PYTHON376_OR_NEWER = sys.version_info > (3, 7, 5)
 
 
+def _get_disable_compression() -> bool:
```
csmarchbanks, line 42: Instead of an environment variable what would you think of passing an additional argument to `make_(w|a)sgi_app`? My thought is that most users should have control over creating the apps so an env var is not necessary. We try to only use env vars in places where users might not have access to the code, such as registering metrics.

ivan-valkov: Yeah, makes sense. I suppose we need to expose this option in `start_http_server` as well?

csmarchbanks: I think that would be nice, though not completely required as `start_http_server` doesn't need to support all configuration for advanced user.

csmarchbanks (approved):
This looks great, thanks a lot!

PR #781, opened 2022-03-09
Simplify type annotations for number types

zhanpon:
@csmarchbanks Thanks for maintaining the project! A small patch to improve the readability of type annotations:

int values are acceptable to float variables:
https://www.python.org/dev/peps/pep-0484/#the-numeric-tower

Signed-off-by: Yiyang Zhan <pon.zhan@gmail.com>


csmarchbanks (approved):
Thanks, I had originally put the Union in to make it clear that we can actually treat int and float differently, but I agree, this is much simpler.

Issue #793, opened 2022-04-06
Gauge - there is `livesum` and `all` but what about `sum`?


nkov:
I have a multi-process application where multiple processes are used to track a single value. For instance, lets say I have 3 processes: A, B, and C behind a load balancer.

`A` receives a request to increment the Gauge: `A = 1, B = 0, C = 0`
`B` receives a request to decrement the Gauge: `A = 1, B = -1, C = 0`

So using `livesum`, my current gauge value is 0 (as expected). However, what would happen if `B` suddenly needs to terminate (eg. due to auto-scaling) or restart (due to a crash, memory management, etc.)? How does the client handle this? Do I assume correctly that the Gauge will now show the incorrect value of `1`, because "live sum" only considers the living processes?

csmarchbanks:
Your analysis is correct, https://github.com/prometheus/client_python/pull/794 was just opened to add the missing modes, so hopefully there will be a sum option soon!

csmarchbanks:
Fixed in #794

nkov:
@csmarchbanks Awesome, what a quick turnaround. When could we expect this to be released to PyPI?

csmarchbanks:
I usually try to cut a release about once every three months but there isn't a specific cadence. I may try to review the multiprocess help text PR and cut a release after that is merged though.

PR #794, opened 2022-04-07
Add `sum`, `livemin`, and `livemax` multiprocess modes for `Gauge`s

JoshKarpel:
Hi!

I work on a multiprocess Flask application and we recently found ourselves wanting the equivalent of the `max` multiprocessing mode, but only for live processes. I came here to poke around the code and also noticed this recent issue asking for a `sum` mode https://github.com/prometheus/client_python/issues/793 , so it seems like there's some interest in more multiprocess modes.

If I understand correctly, the core change needed to support this is actually around the deletion of the files when a process dies - files for `live*` metrics should be deleted, others should not. So I added the new modes with the same collection logic as their non-live counterparts and expanded the check in `mark_process_dead()` to automatically include any metric with `live` as part of its multiprocess mode name (this might be excessively fancy, but it seemed useful to reduce potential future changes).

I also added tests to cover these new modes based on the existing tests and updated the `README.md` to describe the new modes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,9 +614,12 @@ Gauges have several modes they can run in, which can be selected with the `multi
 
 - 'all': Default. Return a timeseries per process alive or dead.
 - 'liveall': Return a timeseries per process that is still alive.
+- 'sum': Return a single timeseries that is the sum of the values of all processes, alive or dead.
 - 'livesum': Return a single timeseries that is the sum of the values of alive processes.
 - 'max': Return a single timeseries that is the maximum of the values of all processes, alive or dead.
+- 'livemax': Return a single timeseries that is the maximum of the values of alive processes.
 - 'min': Return a single timeseries that is the minimum of the values of all processes, alive or dead.
+- 'livemin': Return a single timeseries that is the minimum of the values of alive processes.
 
 ```python
 from prometheus_client import Gauge
--- a/README.md
+++ b/README.md
@@ -609,17 +609,17 @@ def child_exit(server, worker):
 
 **4. Metrics tuning (Gauge)**:
 
-When `Gauge` metrics are used, additional tuning needs to be performed.
+When `Gauge`s are used in multiprocess applications,
+you must decide how to handle the metrics reported by each process.
 Gauges have several modes they can run in, which can be selected with the `multiprocess_mode` parameter.
 
-- 'all': Default. Return a timeseries per process alive or dead.
-- 'liveall': Return a timeseries per process that is still alive.
-- 'sum': Return a single timeseries that is the sum of the values of all processes, alive or dead.
-- 'livesum': Return a single timeseries that is the sum of the values of alive processes.
-- 'max': Return a single timeseries that is the maximum of the values of all processes, alive or dead.
-- 'livemax': Return a single timeseries that is the maximum of the values of alive processes.
-- 'min': Return a single timeseries that is the minimum of the values of all processes, alive or dead.
-- 'livemin': Return a single timeseries that is the minimum of the values of alive processes.
+- 'all': Default. Return a timeseries per process (alive or dead), tagged by the process's `pid` (the tag is added internally).
+- 'min': Return a single timeseries that is the minimum of the values of all processes (alive or dead).
+- 'max': Return a single timeseries that is the maximum of the values of all processes (alive or dead).
+- 'sum': Return a single timeseries that is the sum of the values of all processes (alive or dead).
+
+Prepend 'live' to the beginning of the mode to return the same result but only considering living processes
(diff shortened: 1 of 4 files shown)


csmarchbanks:
Thanks! I should have some time next week to give this a review.

README.md
```diff
@@ -614,9 +614,12 @@ Gauges have several modes they can run in, which can be selected with the `multi
 
 - 'all': Default. Return a timeseries per process alive or dead.
 - 'liveall': Return a timeseries per process that is still alive.
+- 'sum': Return a single timeseries that is the sum of the values of all processes, alive or dead.
 - 'livesum': Return a single timeseries that is the sum of the values of alive processes.
 - 'max': Return a single timeseries that is the maximum of the values of all processes, alive or dead.
+- 'livemax': Return a single timeseries that is the maximum of the values of alive processes.
 - 'min': Return a single timeseries that is the minimum of the values of all processes, alive or dead.
+- 'livemin': Return a single timeseries that is the minimum of the values of alive processes.
```
nkov, line 622: since the support is comprehensive with this change, wonder if we should simplify this to be something like:
```md
- 'all': Default. Return a timeseries per process (alive or dead)
- 'sum': Return a single timeseries that is the some of the values of all processes (alive or dead).
- 'max': Return a single timeseries that is the maximum of the values of all processes (alive or dead).
- 'min': Return a single timeseries that is the minimum of the values of all processes (alive or dead).

Prepend 'live' to the beginning of the mode to return the same result but only considering living processes
- eg: 'liveall, 'livesum', 'livemax', 'livemin'
```

prometheus_client/multiprocess.py
```diff
@@ -63,7 +64,7 @@ def _parse_key(key):
             try:
                 file_values = MmapedDict.read_all_values_from_file(f)
             except FileNotFoundError:
-                if typ == 'gauge' and parts[1] in ('liveall', 'livesum'):
+                if typ == 'gauge' and 'live' in parts[1]:
```
nkov, line 67: maybe `parts[1].startswith('live')`?

prometheus_client/multiprocess.py
```diff
@@ -156,7 +157,6 @@ def mark_process_dead(pid, path=None):
     """Do bookkeeping for when one process dies in a multi-process setup."""
     if path is None:
         path = os.environ.get('PROMETHEUS_MULTIPROC_DIR', os.environ.get('prometheus_multiproc_dir'))
-    for f in glob.glob(os.path.join(path, f'gauge_livesum_{pid}.db')):
-        os.remove(f)
-    for f in glob.glob(os.path.join(path, f'gauge_liveall_{pid}.db')):
-        os.remove(f)
+    for mode in {m for m in Gauge._MULTIPROC_MODES if 'live' in m}:
```
nkov, line 160: if we agree on the previous one then probably would want to make the same change here: `m.startswith('live')`

csmarchbanks:
Hey, some personal stuff came up and it is looking unlikely I am going to be able to review this PR this week. Just wanted to let you know that I have not forgotten about it!

JoshKarpel:
> Some personal stuff came up and it is looking unlikely I am going to be able to review this PR this week. Just wanted to let you know that I have not forgotten about it!

No worries, and no rush! Hope everything turns out ok!

README.md
```diff
@@ -609,14 +609,17 @@ def child_exit(server, worker):
...
+When `Gauge`s are used in multiprocess applications,
+you must decide how to handle the metrics reported by each process.
 Gauges have several modes they can run in, which can be selected with the `multiprocess_mode` parameter.
 
-- 'all': Default. Return a timeseries per process alive or dead.
-- 'liveall': Return a timeseries per process that is still alive.
-- 'livesum': Return a single timeseries that is the sum of the values of alive processes.
-- 'max': Return a single timeseries that is the maximum of the values of all processes, alive or dead.
-- 'min': Return a single timeseries that is the minimum of the values of all processes, alive or dead.
+- 'all': Default. Return a timeseries per process (alive or dead), tagged by the process's `pid` (the tag is added internally).
```
csmarchbanks, line 616: For consistency in this readme and Prometheus could we use labelled and label in this line instead of tagged and tag?

csmarchbanks (commented):
Thanks! This looks great, one small comment otherwise :+1:.

README.md
JoshKarpel: Done!

csmarchbanks (approved):
Awesome, thanks again!

JoshKarpel:
> Awesome, thanks again!

My pleasure! Thanks for reviewing!

Issue #795, opened 2022-04-07
Breaking change in most recent release


richardARPANET:
```
  File "/usr/local/lib/python3.8/dist-packages/prometheus_flask_exporter/__init__.py", line 15, in <module>
    from prometheus_client.exposition import choose_encoder
```

added in `0.14.0` it seems. are you following https://semver.org/?

jmayse:
We are experiencing this issue as well importing `prometheus_client` from `prometheus_flask_exporter`. `prometheus_flask_exporter` has an unversioned import of `prometheus_client` in their requirements.txt and this latest release of `prometheus_client` seems to have broken `prometheus_flask_exporter`. However, this is likely a breaking change for `prometheus_flask_exporter`, not `prometheus_client`, no?

csmarchbanks:
Sorry about this, the breaking change was accidental, and was renamed in https://github.com/prometheus/client_python/pull/776. I am happy to change it back and created https://github.com/prometheus/client_python/pull/796.

Note that breaking changes are allowed in 0.x releases, but if it was a purposeful change a [CHANGE] entry would be added to the changelog.

PR #796, opened 2022-04-08
Rename choose_formatter back to choose_encoder

csmarchbanks:
Other libraries depend on this function and we did not mean to create a breaking change. Add a test to document that this is part of the API.

Fixes https://github.com/prometheus/client_python/issues/795

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -13,6 +13,7 @@ from urllib.parse import parse_qs, quote_plus, urlparse
 from urllib.request import (
     build_opener, HTTPHandler, HTTPRedirectHandler, Request,
 )
+import warnings
 from wsgiref.simple_server import make_server, WSGIRequestHandler, WSGIServer
 
 from .openmetrics import exposition as openmetrics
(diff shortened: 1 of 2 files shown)


Starkteetje:
Note that this renaming is again a breaking change. If you want to avoid it, you could define one of the names as an alias to the other to get backwards-compatibility with the prior versions and 0.14 (though that's obviously ugly)

Starkteetje:
Note that this renaming is again a breaking change. If you want to avoid it, you could define one of the names as an alias to the other to get backwards-compatibility with the prior versions and 0.14 (though that's obviously ugly)

csmarchbanks:
Yeah, that is the case and I would prefer not to maintain the second function name forever. I am hoping that most people who are pinning versions are either pinning both or will upgrade to the patch version fairly quickly. Since the release has only been out for three days a bugfix to revert to the previous behavior seems reasonable to me?

csmarchbanks:
To avoid the second breaking change in 0.14.x I added choose_formatter back in with a deprecation notice. It will be removed as part of 0.15, but that at least gives everyone some time to use either.

Issue #810, opened 2022-05-17
Multiprocessing with Gunicorn + FastAPI


IWillPull:
I've been trying to make multiprocessing work with FastAPI, but it just doesn't seem to happen.

This part is what I've been struggling the most:

```python
def app(environ, start_response):
    registry = CollectorRegistry()
    multiprocess.MultiProcessCollector(registry)
    data = generate_latest(registry)
    status = '200 OK'
    response_headers = [
        ('Content-type', CONTENT_TYPE_LATEST),
        ('Content-Length', str(len(data)))
    ]
    start_response(status, response_headers)
    return iter([data])
```
Had no idea where to put this, but after some experimentation it turns out, the simplest solution of creating a metrics app looks like this (including other parts from the documentation):

```python
app = FastAPI(debug=False)

# Preparing gunicorn multiprocessing HACKS
def make_metrics_app():
    registry = CollectorRegistry()
    multiprocess.MultiProcessCollector(registry)
    return prometheus_client.make_asgi_app(registry=registry)


metrics_app = make_metrics_app()
app.mount("/metrics", metrics_app))
```

Just posting this to help others save hours of researching and debugging.

csmarchbanks:
Thanks for the report! Would you be interested in adding some documentation to the README similar to how we have an example for flask, but under the multiprocess section?

IWillPull:
@csmarchbanks If you think this is a correct way of doing things, I will add the instructions to the README

ashirviskas:
(me from my alt acc)
Added my PR here: https://github.com/prometheus/client_python/pull/811

PR #812, opened 2022-05-18
Fastapi + Gunicorn example in README

ashirviskas:
Also provided additional multiprocessing instructions for FastAPI + Gunicorn setup with code examples as per this issue: #810

Signed-off-by: Matas Minelga <minematas@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,7 +434,7 @@ app = FastAPI(debug=False)
 def make_metrics_app():
     registry = CollectorRegistry()
     multiprocess.MultiProcessCollector(registry)
-    return prometheus_client.make_asgi_app(registry=registry)
+    return make_asgi_app(registry=registry)
 
 
 metrics_app = make_metrics_app()
--- a/README.md
+++ b/README.md
@@ -404,6 +404,56 @@ uwsgi --http 127.0.0.1:8000 --wsgi-file myapp.py --callable app
 
 Visit http://localhost:8000/metrics to see the metrics
 
+#### FastAPI + Gunicorn
+
+To use Prometheus with [FastAPI](https://fastapi.tiangolo.com/) and [Gunicorn](https://gunicorn.org/) we need to serve metrics through a Prometheus ASGI application.
+
+Save the snippet below in a `myapp.py` file
+
+```python
+from fastapi import FastAPI
+from prometheus_client import make_asgi_app
+
+# Create app
+app = FastAPI(debug=False)
+
+# Add prometheus asgi middleware to route /metrics requests
+metrics_app = make_asgi_app()
+app.mount("/metrics", metrics_app)
+```
+
+For Multiprocessing support, use this modified code snippet. Full multiprocessing intstructions are provided [here](https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn).
+
+```python
+from fastapi import FastAPI
+from prometheus_client import make_asgi_app
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -404,6 +404,56 @@ uwsgi --http 127.0.0.1:8000 --wsgi-file myapp.py --callable app
...
+from fastapi import FastAPI
+from prometheus_client import make_asgi_app
+
+app = FastAPI(debug=False)
+
+# Using multiprocess collector for registry
+def make_metrics_app():
+    registry = CollectorRegistry()
+    multiprocess.MultiProcessCollector(registry)
+    return prometheus_client.make_asgi_app(registry=registry)
```
hkiang01, line 437: @ashirviskas should be based on the [import](https://github.com/prometheus/client_python/pull/812/commits/60ced23a24abbd2694e405a128db37afa8bfd52d#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R429)
```python
return make_asgi_app(registry=registry)
```

ashirviskas: I might be late, but I just fixed it

README.md
```diff
@@ -404,6 +404,56 @@ uwsgi --http 127.0.0.1:8000 --wsgi-file myapp.py --callable app
...
+# Using multiprocess collector for registry
+def make_metrics_app():
+    registry = CollectorRegistry()
+    multiprocess.MultiProcessCollector(registry)
+    return make_asgi_app(registry=registry)
+
+
+metrics_app = make_metrics_app()
+app.mount("/metrics", metrics_app))
+
```
csmarchbanks, line 442: Nit: extra newline here.

csmarchbanks (commented):
Generally looks good to me with a small nit. In addition, the second commit needs to be signed off.

README.md
```diff
@@ -404,6 +404,56 @@ uwsgi --http 127.0.0.1:8000 --wsgi-file myapp.py --callable app
...
+# Using multiprocess collector for registry
+def make_metrics_app():
+    registry = CollectorRegistry()
+    multiprocess.MultiProcessCollector(registry)
+    return make_asgi_app(registry=registry)
+
+
+metrics_app = make_metrics_app()
+app.mount("/metrics", metrics_app))
```
JoshChristie, line 441: remove double `) at the end of line

gal-leib:
I know this isn't merged yet, but is it good practice?
Since in the documentation it says the Registry should be created in the state of a request, but here it will be created at the startup of the process no?

OliverFarren:
It's the 11 month anniversary of this PR, are we still deliberating or could this be merged in? 

As per the motive of the original #810, having this in the README.md would have saved significant time.

csmarchbanks:
I cannot merge this without the DCO being signed for all commits.

> I know this isn't merged yet, but is it good practice?
Since in the documentation it says the Registry should be created in the context of a request, but here it will be created at the startup of the process no?

As the readme mentions that is best practice to avoid metrics registering themselves with the multiprocess registry. Since the registry is only created inside of the create app function I think this approach is correct.

ashirviskas:
Whats the easiest way I could resign all the commits?

csmarchbanks (approved):
Looks like it ran when you force pushed (as it should). Thanks!

OliverFarren:
Wow! Thanks everyone for sorting today and for creating the additional PR in the first place.

PR #827, opened 2022-08-01
updated docs with code to remove default metrics

sss-ng:
I spent a while trying to figure out how to remove these default metrics. Im updating the docs with instructions to remove `process`, `gc`, and `platform` collector metrics. 

Please let me know if my language regarding prometheus is correct, and I will gladly update it

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,18 @@ metadata about the JVM in use is also included. This information is available as
 labels on the `python_info` metric. The value of the metric is 1, since it is the 
 labels that carry information.
 
+### Disabling Default Collector metrics
+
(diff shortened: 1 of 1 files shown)


csmarchbanks (commented):
Hello, thank you for the contribution. I am wondering if it is also worthwhile to mention that you can just create a new registry and use that registry instead of the default registry to avoid having the default collectors. Going with the new registry approach will also mean that any future default collectors will not need to be unregistered.

csmarchbanks:
Merging this, we can always add the additional suggested documentation later if we want.

PR #842, opened 2022-09-23
Improve cyclomatic complexity

hack-dream:
Hi! Using the example of my PR, I would like to show that the library needs to improve its cyclomatic complexity.
I tested with the utility [radon](https://pypi.org/project/radon/)

My changes
`
SummaryMetricFamily.__init__ - B (7)
to
SummaryMetricFamily.__init__ - B (6)

HistogramMetricFamily - B(7)
to
Deleted from >=B level complexity

HistogramMetricFamily.__init__ - B(7)
to
HistogramMetricFamily.__init__ - B(6)

CollectorRegistry._get_names - B(7)
to
CollectorRegistry._get_names - B(6)
`

`
629 blocks (classes, functions, methods) analyzed.
Average complexity: A (2.1701112877583464)

to

629 blocks (classes, functions, methods) analyzed.
Average complexity: A (2.1701112877583464)
`


prometheus_client/metrics_core.py
```diff
@@ -121,10 +121,12 @@ def __init__(self,
         if name.endswith('_total'):
             name = name[:-6]
         Metric.__init__(self, name, documentation, 'counter', unit)
-        if labels is not None and value is not None:
-            raise ValueError('Can only specify at most one of value and labels.')
+
         if labels is None:
             labels = []
+        elif value is not None:
```
csmarchbanks, line 127: I find the old logic to be clearer. It was clearly a check at the very beginning to make sure the input was valid, then we assigned a default value. Now we are doing both at the same time which IMO is harder to understand.

csmarchbanks (commented):
Thank you for your contribution, I am not sure if this is an improvement over the old code, generally lowering cyclomatic complexity isn't a goal in and of itself, the goal is the most readable code. I think some of these changes help with readability, e.g. the change to `registry.py`, but not all of them.

alnjieg (approved):

PR #853, opened 2022-11-14
Reduce the mmap minimum size

SuperQ:
Reduce the initial mmap file size from 1MiB to 64KiB. This should reduce the memory and disk overhead for larger numbers of multiprocess workers.

Signed-off-by: SuperQ <superq@gmail.com>


csmarchbanks (approved):
:+1:

PR #856, opened 2022-11-16
Add no-op asgi-lifespan handling

### LINKED ISSUE #855: ASGI lifespan crash
client_python version: testing on 0.15.0 but the offending code exists in the master branch as well.

I'm trying to use the prometheus ASGI handler with Hypercorns DispatcherMiddleware. It sends lifecycle events which are not properly handled:

```
def make_asgi_app(registry: CollectorRegistry = REGISTRY, disable_compression: bool = False) -> Callable:
    """Create a ASGI app which serves the metrics from a registry."""

    async def prometheus_app(scope, receive, send):
        assert scope.get("type") == "http"
        # Prepare parameters
```

`prometheus_app` doesn't HAVE to support lifecycles (DispatcherMiddleware will just ignore lifecycles if the handshake doesn't happen) but `assert scope.get("type") == "http"` will ensure any ASGI middleware sending lifecycle events will fail to start any prometheus services.

sheenobu:
Super simple hook that checks for lifespan and immediately responds type.complete, essentially making startup/shutdown lifespan events no-ops for the prometheus ASGI app.

Fixes #855

```
from hypercorn.middleware import DispatcherMiddleware
from prometheus_client import make_asgi_app
import asyncio

async def main():

    from hypercorn.asyncio import serve
    from hypercorn.config import Config

    app = DispatcherMiddleware({
        "/metrics": make_asgi_app(),
    })


    config = Config()
    config.bind = ["localhost:8080"]

    await serve(app, config)
```

Before:

```
    assert scope.get("type") == "http"
AssertionError
```

After:

```
[2022-11-15 16:05:20 -0800] [59048] [INFO] Running on http://127.0.0.1:8080 (CTRL + C to quit)
INFO:hypercorn.error:Running on http://127.0.0.1:8080 (CTRL + C to quit)
```

Signed-off-by: Sheena Artrip <sheena.artrip@gmail.com>

### DIFF
--- a/prometheus_client/asgi.py
+++ b/prometheus_client/asgi.py
@@ -9,6 +9,10 @@ def make_asgi_app(registry: CollectorRegistry = REGISTRY, disable_compression: b
     """Create a ASGI app which serves the metrics from a registry."""
 
     async def prometheus_app(scope, receive, send):
+        if scope.get("type") == "lifespan":    
+            payload = await receive()    
+            await send({'type': payload['type'] + ".complete"})    
+            return
         assert scope.get("type") == "http"
         # Prepare parameters
         params = parse_qs(scope.get('query_string', b''))


sheenobu:
> If an exception is raised when calling the application callable with a lifespan.startup message or a scope with type lifespan, the server must continue but not send any lifespan events.

this might be a better solution!

csmarchbanks (commented):
> If an exception is raised when calling the application callable with a lifespan.startup message or a scope with type lifespan, the server must continue but not send any lifespan events.

That seems nice, could we just change the assert to instead throw an exception? That would be nice for if other types show up in the future.

PR #858, opened 2022-11-21
add Timer decorator support partial function

dafu-wu:
We should support partial functions for Timer


csmarchbanks:
Thank you for the PR! `decorator.py` is actually copied from an old version of https://github.com/micheles/decorator, and we do not want to manually modify the contents. I would be ok updating to a newer version of decorator if it supports partial functions.

dafu-wu:
> 

ok, I tried to submit a pr in the decorator.

dafu-wu:
@https://github.com/micheles/decorator/pull/150

dafu-wu:
#https://github.com/micheles/decorator/pull/150

GhostLyrics:
Will this maybe as a side-effect solve issues with the decorator making an function untyped? (example attached)

```
mypy .
toolchain/prometheus_timetac_exporter.py:47:2: error: Untyped decorator makes function "collect_ttqueue_oldest_pending_job_timestamp" untyped  [misc]
    @METRIC_QUERY_RESULT_TIME.time()
     ^
Found 1 error in 1 file (checked 16 source files)
make: *** [lint] Error 1
```

Currently I've solved this via a `# type: ignore` directive.

Issue #861, opened 2022-11-30
application/openmetrics-text encoding cannot be selected in asgi app using the Accept header 


jezekra1:
The openmetrics format cannot be selected using the `Accept: application/openmetrics-text` header, because there is the extra header key in the beginning in `asgi.py:make_asgi_app` which I believe is incorrect. Compare with the `accept_encoding_header` just below.
https://github.com/prometheus/client_python/blob/master/prometheus_client/asgi.py#L15:
```python
         # Accept header key here:
        accept_header = "Accept: " + ",".join([
            value.decode("utf8") for (name, value) in scope.get('headers')
            if name.decode("utf8").lower() == 'accept'
        ])
        # No header key here:
        accept_encoding_header = ",".join([
            value.decode("utf8") for (name, value) in scope.get('headers')
            if name.decode("utf8").lower() == 'accept-encoding'
        ])

```

The condition in `exposition.py` is thus comparing the full header:

 `accepted.split(';')[0].strip()` results to `Accept: application/openmetrics-text`.

```python
def choose_encoder(accept_header: str) -> Tuple[Callable[[CollectorRegistry], bytes], str]:
    accept_header = accept_header or ''
    for accepted in accept_header.split(','):
        if accepted.split(';')[0].strip() == 'application/openmetrics-text':
            return (openmetrics.generate_latest,
                    openmetrics.CONTENT_TYPE_LATEST)
    return generate_latest, CONTENT_TYPE_LATEST
```

csmarchbanks:
Good find! Are you interested in creating a PR?

csmarchbanks:
Fixed in #864

PR #864, opened 2022-12-04
Fix wrong encoder used with `application/openmetrics-text` accept header

jezekra1:
Issue: https://github.com/prometheus/client_python/issues/861


csmarchbanks (approved):
Thanks!

PR #866, opened 2022-12-04
Support metric help text in multiprocess mode

evgenymarkov:
Hello ✋🏻 I ran into a problem - open telemetry collector shows errors because I collect metrics with the same name from two applications - one single-process, the other multi-process:

label:<name:"path_template" value:"/api/rest/v1/auth/me" > label:<name:"project" value:"auth" > counter:<value:890 >  has help "Total count of requests by method and path" but should have "Multiprocess metric"

Then I found old [issue](https://github.com/prometheus/client_python/issues/211) and I decided to fix it. @butlerx did almost all the work. I added only 1 test.


evgenymarkov:
See also: #866

evgenymarkov:
I'll close my PR because it's exactly the same as #804

Issue #871, opened 2022-12-19
Parser produces invalid metrics for no-space metrics


MatthewScholefield:
The prometheus metrics endpoint for an IoT tutorial emits metrics with no space after the metric like this:

```
# HELP pm02 Particulate Matter PM2.5 value
# TYPE pm02 gauge
pm02{id="airgrad0",mac="FF:FF:FF:FF:FF:FF"}11
# HELP rco2 CO2 value, in ppm
# TYPE rco2 gauge
rco2{id="airgrad0",mac="FF:FF:FF:FF:FF:FF"}532
# HELP atmp Temperature, in degrees Celsius
# TYPE atmp gauge
atmp{id="airgrad0",mac="FF:FF:FF:FF:FF:FF"}22.10
# HELP rhum Relative humidity, in percent
# TYPE rhum gauge
rhum{id="airgrad0",mac="FF:FF:FF:FF:FF:FF"}40
```

Parsing this didn't error but instead produced metrics with incorrect values:
```
[Metric(pm02, Particulate Matter PM2.5 value, gauge, , [Sample(name='pm02', labels={'id': 'airgrad0', 'mac': 'FF:FF:FF:FF:FF:FF'}, value=1.0, timestamp=None, exemplar=None)]),
 Metric(rco2, CO2 value, in ppm, gauge, , [Sample(name='rco2', labels={'id': 'airgrad0', 'mac': 'FF:FF:FF:FF:FF:FF'}, value=32.0, timestamp=None, exemplar=None)]),
 Metric(atmp, Temperature, in degrees Celsius, gauge, , [Sample(name='atmp', labels={'id': 'airgrad0', 'mac': 'FF:FF:FF:FF:FF:FF'}, value=2.1, timestamp=None, exemplar=None)]),
 Metric(rhum, Relative humidity, in percent, gauge, , [Sample(name='rhum', labels={'id': 'airgrad0', 'mac': 'FF:FF:FF:FF:FF:FF'}, value=0.0, timestamp=None, exemplar=None)])]
```
I think ideally perhaps it should raise a `ValueError` in this circumstance (or alternatively support this if it's considered a valid metrics format; I didn't find a spec on what the exact format is).

csmarchbanks:
Interesting, I definitely think this is an issue. The best spec I could find is https://prometheus.io/docs/instrumenting/exposition_formats/, which states
[...]

PR #872, opened 2022-12-21
Allow Prometheus parser to handle missing space

csmarchbanks:
If there is no delimiter separating the labels and value when parsing the Prometheus format the parser currently will chop off the first digit of the value. Instead, allow a missing delimiter in the Prometheus format to align with what the Go parser does.

In addition, add a test to specifically check that a delimiter is still required in OpenMetrics.

Fixes: #871


roidelapluie (approved):
LGTM

PR #873, opened 2022-12-22
Add tests for Python 3.11

csmarchbanks:
The base image has also been moved to the new cimg repositories which are direct replacements for circleci.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       python:
         type: string
     docker:
-      - image: circleci/python:<< parameters.python >>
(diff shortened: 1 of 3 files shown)


SuperQ (approved):
LGTM.

We could probably drop support for Python 3.6. It was EoL a year ago[0].

[0]: https://devguide.python.org/versions/

csmarchbanks:
Good point on 3.6, I don't know of any features that are blocked so I think I'll leave it alone for now. Once 3.7 is EOL I plan to drop both 3.6 and 3.7 to use some new typing features.

PR #891, opened 2023-02-16
Modify the way of starting service 

SilenceAdele:
* Modify the way of starting service


csmarchbanks (commented):
Hello, thank you for your PR!

I am not sure I like forcing a user to clone this repository to run the quickstart example, having it embedded in the README is nice for that. Is there a reason you like the separate file more?

SilenceAdele:
ok，I konw that what you mean.

PR #903, opened 2023-03-21
Extend type forwarding to other context managers

crutcher:
It would be really nice to be able to use the various decorators without losing the type information of the decorated method.


csmarchbanks (approved):
Thanks!

PR #905, opened 2023-03-24
Support `_total` suffix for counter sample names, according to the spec

alopezz:
I found out that the counter in the example given in this [section of the OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure) was not being parsed correctly by the parsing functions under `prometheus_client.parser`. Instead, it only works if the name of the sample matches the name of the metric family (if I get the terminology right).

This PR provides a simple fix for that. I'm aware that there's another parsing function under `prometheus_client.openmetrics.parser` which _does_ parse that correctly, but that one is also more strict in other ways, and depending on who's emitting the metrics it might not work.

@csmarchbanks

### DIFF
--- a/prometheus_client/parser.py
+++ b/prometheus_client/parser.py
@@ -158,7 +158,8 @@ def text_fd_to_metric_families(fd: TextIO) -> Iterable[Metric]:
             else:
                 new_samples = []
                 for s in samples:
-                    new_samples.append(Sample(s[0] + '_total', *s[1:]))
+                    sample_name = s.name if s.name.endswith('_total') else s.name + '_total'
+                    new_samples.append(Sample(sample_name, *s[1:]))
                     samples = new_samples
         metric = Metric(name, documentation, typ)
         metric.samples = samples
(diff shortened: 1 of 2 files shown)


csmarchbanks (commented):
This is an interesting issue, my reading of the [text exposition format](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information) is that the name in the type/help text must match the name of the metric for counters (summaries and histograms are the only types with suffixes). To me that means we should not be adding suffixes to counters when parsing, but the prometheus exposition format is also generally under-specified.

When I run your example on the existing code I get a metric family for an unknown metric with name `a_total`. This also seems to match what is done by the Go parser which is used by the Prometheus server, so I think the existing behavior should be kept, and if you are processing an OpenMetrics payload you should use the OpenMetrics parser. What are the additional constraints of that parser that you are having problems with?

alopezz:
I wasn't aware of what the actual target for `prometheus_client.parser` and how it differed from the `openmetrics` one; I was assuming that the former was simply a less strict version of the latter. I had missed that bit of documentation, which helps and I understand your points.

This fragment and its associated [commit](https://github.com/prometheus/client_python/commit/a4dd93bcc6a0422e10cfa585048d1813909c6786) message led me to think that this parser was also expected to follow OpenMetrics by dealing with the `_total` suffix specially:
https://github.com/prometheus/client_python/blob/c48e26e2283fee2c2ef628608e2fed6468f6a911/prometheus_client/parser.py#L153-L162

A "generic" parser that would handle both exposition formats (without requiring strict adherence to the OpenMetrics) spec would be helpful to avoid having to know the actual format precisely ahead of time. Also, in the specific case that triggered this, the same exposition that was using the `_total` suffix following the OpenMetrics was also missing a `# EOF` at the end.

PR #906, opened 2023-03-29
avoid installing __pycache__ directories

minrk:
the current prometheus wheels include `__pycache__` directories in their package data files.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,4 @@
-recursive-include tests *
-prune tests/__pycache__
+graft tests
+global-exclude *.py[cod]
+prune __pycache__
+prune */__pycache__


csmarchbanks (approved):
Thanks! When I am testing this out it appears to work for the source distribution, and I am still seeing the various compiled python files added to the wheel. It appears that MANIFEST.in only applies to the source distribution.

I am happy to merge this as it is a definite improvement but do you want to take a stab at that as well? If not I will work on adding some methods for cleaning the repo before building for the next release.

minrk:
How do you build your wheels? I couldn't find it to test with. I tested with the current standard `python3 -m build .` and it didn't include any pycache files in sdist or wheel, nor did `python3 setup.py bdist_wheel` (setuptools 65.6.3). So maybe it's less the manifest and more the release process that's doing it?

PR #914, opened 2023-05-29
Fix typo in README

cbguder:
@csmarchbanks


csmarchbanks (approved):
Thanks!

Looks like the CI failures are unrelated to your change. I will investigate those when I have time.

Issue #932, opened 2023-07-04
Exemplars not working


mumrah:
I am not able to get exemplars working in version 0.17.0. Here is a basic reproduction:

```python
from prometheus_client import start_http_server, Summary, Counter
import random
import time

# Create a metric to track time spent and requests made.
REQUEST_TIME = Summary('request_processing_seconds', 'Time spent processing request')
COUNTER = Counter("sample_count", "Sample counter")

# Decorate function with metric.
@REQUEST_TIME.time()
def process_request(t):
    """A dummy function that takes some time."""
    COUNTER.inc(1, exemplar={"foo": "bar"})
    time.sleep(t)

if __name__ == '__main__':
    # Start up the server to expose the metrics.
    start_http_server(8000)
    # Generate some requests.
    while True:
        process_request(random.random())
```

```
# HELP python_gc_objects_collected_total Objects collected during gc
# TYPE python_gc_objects_collected_total counter
python_gc_objects_collected_total{generation="0"} 246.0
python_gc_objects_collected_total{generation="1"} 114.0
python_gc_objects_collected_total{generation="2"} 0.0
# HELP python_gc_objects_uncollectable_total Uncollectable objects found during GC
# TYPE python_gc_objects_uncollectable_total counter
python_gc_objects_uncollectable_total{generation="0"} 0.0
python_gc_objects_uncollectable_total{generation="1"} 0.0
python_gc_objects_uncollectable_total{generation="2"} 0.0
# HELP python_gc_collections_total Number of times this generation was collected
# TYPE python_gc_collections_total counter
python_gc_collections_total{generation="0"} 39.0
python_gc_collections_total{generation="1"} 3.0
python_gc_collections_total{generation="2"} 0.0
# HELP python_info Python platform information
# TYPE python_info gauge
python_info{implementation="CPython",major="3",minor="9",patchlevel="16",version="3.9.16"} 1.0
# HELP request_processing_seconds Time spent processing request
# TYPE request_processing_seconds summary
request_processing_seconds_count 17.0
request_processing_seconds_sum 6.808643133
# HELP request_processing_seconds_created Time spent processing request
# TYPE request_processing_seconds_created gauge
request_processing_seconds_created 1.688505443345786e+09
# HELP sample_count_total Sample counter
# TYPE sample_count_total counter
sample_count_total 18.0
# HELP sample_count_created Sample counter
# TYPE sample_count_created gauge
sample_count_created 1.6885054433458078e+09
```

[...]

PR #934, opened 2023-07-06
Add documentation for how to export exemplars

csmarchbanks:
Expand our exemplar documentation to include information for how to use the appropriate exposition code and mention the feature flag required in Prometheus to ingest exemplars.

Suggested in #932

cc: @mumrah would you be willing to take a look at this and see if it would have helped you?

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,20 @@ h = Histogram('request_latency_seconds', 'Description of histogram')
 h.observe(4.7, {'trace_id': 'abc123'})
 ```
 
+Exemplars are only rendered in the OpenMetrics exposition format. If using the
+HTTP server or apps in this library, content negotiation can be used to specify
+OpenMetrics (which is done by default in Prometheus). Otherwise it will be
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -252,6 +252,18 @@ h = Histogram('request_latency_seconds', 'Description of histogram')
 h.observe(4.7, {'trace_id': 'abc123'})
 ```
 
+Exemplars are only rendered in the OpenMetrics exposition format. If using the
+HTTP server or apps in this library, content negotiation can be used to specify
+OpenMetrics (which is done by default in Prometheus). Otherwise it will be
+necessary to use `generate_latest` from
+`prometheus_client.openmetrics.exposition` to view exemplars.
+
+To view exemplars in Prometheus it is also necessary to enable the the
```
roidelapluie, line 261: Maybe we can link to the actual doc in addition

csmarchbanks: Good idea, updated.

roidelapluie (approved):

PR #935, opened 2023-07-11
Gauge typing improvement

rafsaf:
For python +3.8, use Literal for multiprocess_mode param in Gauge

### DIFF
--- a/prometheus_client/decorator.py
+++ b/prometheus_client/decorator.py
@@ -35,6 +35,7 @@ from __future__ import print_function
 
 import collections
 import inspect
+from inspect import getfullargspec
 import itertools
 import operator
 import re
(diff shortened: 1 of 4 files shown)


prometheus_client/metrics.py
```diff
@@ -17,6 +18,9 @@
 from .samples import Exemplar, Sample
 from .utils import floatToGoString, INF
 
+if sys.version_info >= (3, 8, 0):
```
csmarchbanks, line 21: Python 3.7 was just marked EOL a couple weeks ago, which reminded me to open https://github.com/prometheus/client_python/pull/936. What would you think of merging that first so we do not need this check?

rafsaf: Sure! I noticed python 3.6 and 3.7 around the repository and decided not to break it.

Let's then wait for the other PR and I will happily rebase

csmarchbanks: Support for 3.7 has been removed, so feel free to rebase this PR. Thank you!

rafsaf:
Rebased after https://github.com/prometheus/client_python/pull/936

Included removal of old sys.version_info code around project as a bit of refactoring

prometheus_client/decorator.py
csmarchbanks, line 1: This file is actually copied from https://github.com/micheles/decorator to avoid any dependencies and we do not want to modify it locally. At some point i need to update the version, but for now can we just keep it the same?

csmarchbanks (commented):
One small thing, otherwise :+1: and thank you!

rafsaf:
@csmarchbanks thanks, done: decorator.py left untouched

csmarchbanks (approved):
Great, thank you!

PR #936, opened 2023-07-12
Remove Support for Python < 3.8

csmarchbanks:
Python 3.7 is now end of lifed, which allows additional features from the core typings library.

I am going to leave this up for a couple days to see if there are significant objections.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,8 +75,6 @@ workflows:
           matrix:
             parameters:
                python:
-                 - "3.6"
-                 - "3.7"
                  - "3.8"
                  - "3.9"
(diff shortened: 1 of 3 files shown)


SuperQ:
Checking the major distros, I think the only down side is RHEL 7 defaults to Python 3.6 from what I can see.

csmarchbanks:
Good shout, I am comfortable bumping the version and requiring those running RHEL 7 to pin their version to 0.17.x. If needed I can do more bugfix releases to 0.17.x as well. Does that sound reasonable?

SuperQ (approved):
Sounds like a good plan.

Issue #912, opened 2023-05-17
Adding `_total` to `Metric` doesn't offset the `_total` if `_total` already exist


vivekpatani:
Here's the sample code
```python3
metric = Metric('num_requests_total', 'Total Number of Requests' , 'counter')
metric.add_sample('num_requests_total', value=stat['num_requests_total'], labels={'xyz': stat['name']}
```

This shows up on the prometheus endpoint as
```
# TYPE num_requests_total_total counter
num_requests_total{xyz="random_name"} 10.0
```

Now there are two `_total_total` at the end. Am I missing something here? I see the `counterMetric` function already has logic to remove the extra `_total` towards the end.

Thanks.

csmarchbanks:
This looks like a bug to me, the TYPE should also just be `num_requests_total` I believe.

mansoor17syed:
Hi,
I would like contribute here.
Please assign to me.

Thank You

vivekpatani:
any updates on this? Thanks.

mansoor17syed:
Hi
https://github.com/anantha96
Shall push the changes

anantha96:
Hi @csmarchbanks, Fix submitted in #951

csmarchbanks:
While reviewing #951 I realized that you are using an internal only method and we already have the appropriate logic in `Counter` and `CounterMetricFamily`. 

You will see in the docstring for `Metric` that:
```
This is intended only for internal use by the instrumentation client.
```

I am going to close this issue as you should be using either `Counter` or `CounterMetricFamily` instead of `Metric`.

PR #951, opened 2023-08-21
Remove _total when passed in counter metric name

anantha96:
Hi @csmarchbanks, requesting to please review the fix for #912


csmarchbanks:
See https://github.com/prometheus/client_python/issues/912#issuecomment-1714458302, but I don't think this is necessary as `Metric` is internal only and we handle removing an extra `_total` everywhere that instantiates a `Metric` already.

PR #960, opened 2023-09-20
Fix duplicated (+incorrect) metrics when using multiproc environment

RedKinda:
Currently, when using a multiproc environment, initializing metrics registers them as collectors. This means that in `registry.py:96` they report their metrics, even though they were already reported in the mmapped environment. In practice this meant dumping metrics would report some metrics twice with the same identifiers, once for the aggregated multiproc collector and once for the Histogram or other metrics used by the local process handling that specific `/metrics` request.

This PR prevents Metric-s being registered into the registry as collectors by checking for presence of the two relevant env vars.


csmarchbanks (commented):
Interesting, this is the reason why in examples we always create a new Collector registry that is not used for registering any metrics, e.g.
```
data = generate_latest(registry)
```
as part of the handler.

I am trying to think if there are any valid use cases for having a second registry with a different set of metrics, but as you say, I think they will all show up in the multiprocess output so probably not.

RedKinda:
I was following the steps shown in the fastapi section https://github.com/prometheus/client_python/#fastapi--gunicorn which uses `make_asgi_app`

Issue #847, opened 2022-10-20
"last" aggregation in Multiprocess_mode (Gauge)


singerjess:
Related to https://github.com/prometheus/client_python/issues/154.
I'm working with Fast API + Gunicorn, and I have several cases where I need to display a gauge metric, but only the "last" value received. Not the sum, or the maximum value (for example, the CPU usage of another device that reports that metric to my API).

Currently, the Multiprocess_mode doesn't have a way to provide the "last" value. I think that would be a valuable feature to add.

What I did locally, was add another metric with the timestamp of the last metric added by each node, and then with a custom collector select the process with the higher timestamp, and return the metric associated with that process id. 
This is a workaround, a cleaner way would be to add a new multiprocess mode to Gauges ("last"), and set the timestamp to the Gauge (currently I see it's null) when that mode is used. Want to check what you think of a potential solution like this before doing a PR.

I include here the code of the "accumulate_metrics" method from my custom collector. (There are a couple of technical details, since I can't add the "last" mode, I use the "liveall" mode and I specify the metric name starts with "last"):

```    
@staticmethod
    def _accumulate_metrics(metrics, accumulate):

        ### Custom code to get last value
        last_pid = -1
        last_timestamp = 0.0
        for index, metric in enumerate(metrics.values()):
            for s in metric.samples:
                name, labels, value, timestamp, exemplar = s
                current_pid = -1
                if name == "process_heartbeat_timestamp":
                    for l in labels:
                        if l[0] == 'pid':
                            current_pid = l[1]
                    if float(value) > float(last_timestamp):
                        last_timestamp = value
                        last_pid = int(current_pid)

        ### End custom code
        for index, metric in enumerate(metrics.values()):
            samples = defaultdict(float)
            buckets = defaultdict(lambda: defaultdict(float))
            samples_setdefault = samples.setdefault
            for s in metric.samples:
                name, labels, value, timestamp, exemplar = s

                if metric.type == 'gauge':
                    without_pid_key = (name, tuple(l for l in labels if l[0] != 'pid'))
                    ### Custom code to get last value
                    if metric._multiprocess_mode == 'liveall' and name.startswith("last"):
                        current_pid = -1
                        for l in labels:
                            if l[0] == 'pid':
                                current_pid = int(l[1])
                        current = samples_setdefault(without_pid_key, value)
                        if current_pid == last_pid:
                            samples[without_pid_key] = value
                    ### End custom code
                    elif metric._multiprocess_mode == 'min':
[...]

PR #967, opened 2023-10-17
Add mostrecent aggregation to Gauge

draftcode:
In the multiprocess mode, the process that exposes the metrics needs to aggregate the samples from other processes. Gauge metric allows users to choose the aggregation mode. This implements 'mostrecent' (and 'livemostrecent') mode where the last observed value is exposed.

In order to support this, the file format is expanded to store the timestamps in addition to the values. The stored timestamps are read by the reader process and it's used to find the latest value.

Closes https://github.com/prometheus/client_python/issues/847

Consideration on the atomicity:

Previously, mmap_dict.py had a comment saying "We assume that reading from an 8 byte aligned value is atomic". With this change, the value write becomes a 16 bytes 8-byte aligned write. The code author (draftcode) tried to find a basis on the original assumption, but couldn't find any. According to write(2), **if a file descriptor is shared**, the write becomes atomic. However, we do not share the file descriptors in the current architecture.

Considering that Ruby implementation also does the same and hadn't seen an issue with it, this write atomicity problem might be practically not an issue.

See also:

* https://github.com/prometheus/client_ruby/pull/172

  The approach and naming are taken from client_ruby.

* https://github.com/prometheus/client_golang/blob/v1.17.0/prometheus/metric.go#L149-L161

  client_golang has an API for setting timestamp already. It explains the use case for the timestamp beyond the client-local aggregation. In order to support the same use case in Python, further changes are needed.

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -346,7 +346,8 @@ class Gauge(MetricWrapperBase):
         d.set_function(lambda: len(my_dict))
     """
     _type = 'gauge'
-    _MULTIPROC_MODES = frozenset(('all', 'liveall', 'min', 'livemin', 'max', 'livemax', 'sum', 'livesum'))
+    _MULTIPROC_MODES = frozenset(('all', 'liveall', 'min', 'livemin', 'max', 'livemax', 'sum', 'livesum', 'mostrecent', 'livemostrecent'))
+    _MOST_RECENT_MODES = frozenset(('mostrecent', 'livemostrecent'))
 
     def __init__(self,
                  name: str,
@@ -357,7 +358,7 @@ class Gauge(MetricWrapperBase):
                  unit: str = '',
                  registry: Optional[CollectorRegistry] = REGISTRY,
                  _labelvalues: Optional[Sequence[str]] = None,
-                 multiprocess_mode: Literal['all', 'liveall', 'min', 'livemin', 'max', 'livemax', 'sum', 'livesum'] = 'all',
+                 multiprocess_mode: Literal['all', 'liveall', 'min', 'livemin', 'max', 'livemax', 'sum', 'livesum', 'mostrecent', 'livemostrecent'] = 'all',
                  ):
         self._multiprocess_mode = multiprocess_mode
         if multiprocess_mode not in self._MULTIPROC_MODES:
@@ -373,6 +374,7 @@ class Gauge(MetricWrapperBase):
             _labelvalues=_labelvalues,
         )
         self._kwargs['multiprocess_mode'] = self._multiprocess_mode
+        self._is_most_recent = self._multiprocess_mode in self._MOST_RECENT_MODES
 
     def _metric_init(self) -> None:
         self._value = values.ValueClass(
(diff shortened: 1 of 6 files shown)


prometheus_client/metrics.py
```diff
@@ -390,10 +390,15 @@ def dec(self, amount: float = 1) -> None:
         self._raise_if_not_observable()
         self._value.inc(-amount)
 
-    def set(self, value: float) -> None:
-        """Set gauge to the given value."""
+    def set(self, value: float, timestamp_sec: Optional[float] = None) -> None:
```
csmarchbanks, line 393: 🤔  I wonder if it would be best to start by not taking an additional timestamp_sec variable, and calculate the current time ourselves iff multiprocess_mode is `mostrecent` or `livemostrecent`?

To my knowledge most client libraries avoid letting users easily set timestamps on metrics as it becomes dangerous very quickly. This would also mean that users could use the new functionality without having to manually pass timestamps around.

prometheus_client/multiprocess.py
```diff
@@ -143,7 +150,7 @@ def _accumulate_metrics(metrics, accumulate):
                         samples[(metric.name + '_count', labels)] = acc
 
             # Convert to correct sample format.
-            metric.samples = [Sample(name_, dict(labels), value) for (name_, labels), value in samples.items()]
+            metric.samples = [Sample(name_, dict(labels), value, sample_timestamp_secs.get((name_, labels), None)) for (name_, labels), value in samples.items()]
```
csmarchbanks, line 153: I don't think we should be exposing the timestamp in the final rendered result. The fact that we store the timestamp should be an internal detail only for use with mostrecent.

prometheus_client/values.py
```diff
@@ -12,16 +13,23 @@ class MutexValue:
 
     def __init__(self, typ, metric_name, name, labelnames, labelvalues, help_text, **kwargs):
         self._value = 0.0
+        self._timestamp_sec = 0.0
```
csmarchbanks, line 16: I think I would slightly prefer just `timestamp` as the variable name across this PR. I don't think we have units in other variable names at least.

csmarchbanks (commented):
Thank you!

I left a few initial comments, but need to spend a bit more time testing/reviewing in addition to those.

prometheus_client/multiprocess.py
draftcode: OK. I can drop this from this PR. I have a usage on the timestamp part (actually, my usage is exactly the same as Ruby's PR, exposing GitHub API limits), but as you said in the other comment, this usage must be a rare case so need to be careful. Let me add that a FR separated from this PR.

prometheus_client/metrics.py
csmarchbanks: Hmm, I think I would prefer to use the same Gauge, mostly because `TimestampedGauge` sounds like it would export the timestamp as well, which I don't want to include as part of this PR. For inc/dec I think we can just throw if they are used with `mostrecent` or `livemostrecent`. I could be convinced though.

draftcode:
OK. Addressed the comments. Please take a look again.

prometheus_client/metrics.py
```diff
@@ -382,18 +383,25 @@ def _metric_init(self) -> None:
 
     def inc(self, amount: float = 1) -> None:
         """Increment gauge by the given amount."""
+        if self._multiprocess_mode in self._MOST_RECENT_MODES:
```
csmarchbanks, line 386: I wonder if instead of having to do multiple string comparisons in each call for inc/dec we should set this as `_is_most_recent` or something like that on the gauge at init time?

csmarchbanks (commented):
Generally this is looking good to me, just one small comment. Thanks!

prometheus_client/metrics.py
draftcode: Done.

prometheus_client/metrics.py
```diff
@@ -382,18 +384,25 @@ def _metric_init(self) -> None:
...
+        if self._is_most_recent:
+            raise RuntimeError("dec must not be used with the mostrecent mode")
         self._raise_if_not_observable()
         self._value.inc(-amount)
 
     def set(self, value: float) -> None:
         """Set gauge to the given value."""
         self._raise_if_not_observable()
-        self._value.set(float(value))
+        if self._multiprocess_mode in self._MOST_RECENT_MODES:
```
csmarchbanks, line 402: One more place to update :)

draftcode: Oops. Done.

csmarchbanks (commented):
The code looks good, thanks! I realized we should also update the "Metrics tuning (Gauge):" section in the [README](https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn) before merging, basically just adding a bullet point for 'recent'.

draftcode:
> The code looks good, thanks! I realized we should also update the "Metrics tuning (Gauge):" section in the [README](https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn) before merging, basically just adding a bullet point for 'recent'.

Done.

csmarchbanks (approved):
Thanks!

PR #968, opened 2023-10-19
add restricted registry documentation

remijouannet:
i found this undocumented feature in the code, it's pretty useful to debug an exporter that return thousand of metrics.
i don't know if having this feature undocumented was intentional or  not
(@csmarchbanks)

### DIFF
--- a/README.md
+++ b/README.md
@@ -740,7 +740,7 @@ for family in text_string_to_metric_families(u"my_gauge 1.0\n"):
 ## Restricted registry
 
 Registry support restriction to only return specific metrics.
-If you’re using the built-in HTTP server, you can use the GET parameter "name []", since it’s an array it can be used multiple times.
+If you’re using the built-in HTTP server, you can use the GET parameter "name[]", since it’s an array it can be used multiple times.
 If you’re directly using generate_latest, you can use the function restricted_registry().
 
 ```python
--- a/README.md
+++ b/README.md
@@ -743,22 +743,7 @@ Registry support querying a specific metric with the GET parameter "name[]".
 Since it's an array it can be use multiple times.
 
 ```python
-curl -v --get --data-urlencode "name[]=python_gc_objects_collected_total" --data-urlencode "name[]=python_info" http://127.0.0.1:9200/metrics
-* About to connect() to 127.0.0.1 port 9200 (#0)
-*   Trying 127.0.0.1...
-* Connected to 127.0.0.1 (127.0.0.1) port 9200 (#0)
-> GET /metrics?name[]=python_gc_objects_collected_total&name[]=python_info HTTP/1.1
-> User-Agent: curl/7.29.0
-> Host: 127.0.0.1:9200
-> Accept: */*
->
-* HTTP 1.0, assume close after body
-< HTTP/1.0 200 OK
-< Date: Thu, 19 Oct 2023 10:00:38 GMT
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -737,6 +737,39 @@ for family in text_string_to_metric_families(u"my_gauge 1.0\n"):
...
+> Host: 127.0.0.1:9200
+> Accept: */*
+>
+* HTTP 1.0, assume close after body
+< HTTP/1.0 200 OK
+< Date: Thu, 19 Oct 2023 10:00:38 GMT
+< Server: WSGIServer/0.2 CPython/3.9.3
+< Content-Type: text/plain; version=0.0.4; charset=utf-8
+< Content-Length: 454
+<
```
csmarchbanks, line 761: Having the verbose output is unnecessary in the readme.

README.md
```diff
@@ -737,6 +737,39 @@ for family in text_string_to_metric_families(u"my_gauge 1.0\n"):
     print("Name: {0} Labels: {1} Value: {2}".format(*sample))
 ```
 
+## Restricted registry
+
+Registry support querying a specific metric with the GET parameter "name[]".
```
csmarchbanks, line 742: This is only true if someone is using one of the built in handlers/apps, I wonder if it is worth calling that out? I.e. it won't magically work if someone is calling `generate_latest` from an endpoint themselves.

csmarchbanks (commented):
Thanks!

One note, you can also define a restricted registry from code, perhaps document how to do that as well as the `name[]` behavior?

README.md
remijouannet: done

README.md
remijouannet: done

README.md
```diff
@@ -737,6 +737,34 @@ for family in text_string_to_metric_families(u"my_gauge 1.0\n"):
     print("Name: {0} Labels: {1} Value: {2}".format(*sample))
 ```
 
+## Restricted registry
+
+Registry support restriction to only return specific metrics.
+If you’re using the built-in HTTP server, you can use the GET parameter "name []", since it’s an array it can be used multiple times.
```
csmarchbanks, line 743:
```suggestion
If you’re using the built-in HTTP server, you can use the GET parameter "name[]", since it’s an array it can be used multiple times.
```

README.md
```diff
@@ -737,6 +737,34 @@ for family in text_string_to_metric_families(u"my_gauge 1.0\n"):
     print("Name: {0} Labels: {1} Value: {2}".format(*sample))
 ```
 
+## Restricted registry
+
+Registry support restriction to only return specific metrics.
```
csmarchbanks, line 742:
```suggestion
Registries support restriction to only return specific metrics.
```

README.md
```diff
@@ -737,6 +737,34 @@ for family in text_string_to_metric_families(u"my_gauge 1.0\n"):
...
+If you’re using the built-in HTTP server, you can use the GET parameter "name []", since it’s an array it can be used multiple times.
+If you’re directly using generate_latest, you can use the function restricted_registry().
```
csmarchbanks, line 744:
```suggestion
If you’re directly using `generate_latest`, you can use the function `restricted_registry()`.
```

csmarchbanks (commented):
A few small suggestions, but generally 👍!

csmarchbanks (approved):

csmarchbanks:
Thanks!

PR #970, opened 2023-10-22
Mark collector parameter as optional in ProcessCollector and PlatformCollector

albertodonato:
The `registry` parameter can be passed as `None`, but this wasn't reflected in the typing.


csmarchbanks (approved):
Thanks! It looks like PlatformCollector has this issue as well, would you be interested in updating that too?

albertodonato:
Sure, done!

csmarchbanks (approved):
Thanks!

PR #973, opened 2023-10-26
Allow enabling/disabling _created metrics from code

### LINKED ISSUE #933: Feature request: allow disabling "_created" series from code
At the moment, "_created" series can be disabled by setting PROMETHEUS_DISABLE_CREATED_SERIES environment variable. If one wants to set it from python, he needs to do it before importing prometheus_client, because the value is calculated on import. 
Like this:
```
os.environ['PROMETHEUS_DISABLE_CREATED_SERIES'] = 'true'
import prometheus_client
```
That is not very pythonic, style checkers complain on imports not being at top of the file, etc.

Please add an ability to set this option from code, from function call, of just make `_use_created` variable in `metrics.py` public.

csmarchbanks:
Fixes #933

This takes the approach of globally enabling/disabling _created metrics still. Another option would be to enable/disable `_created` metrics on each registry, but that would be a larger change as we would have to keep persist the registry on each metric (or filter them out which seems more prone to errors/conflicts?). My thought is to start with the global option, but we could add the registry option for more fine use cases if required in the future?

@SuperQ WDYT?

### DIFF
--- a/prometheus_client/__init__.py
+++ b/prometheus_client/__init__.py
@@ -11,7 +11,10 @@ from .exposition import (
     write_to_textfile,
 )
 from .gc_collector import GC_COLLECTOR, GCCollector
-from .metrics import Counter, Enum, Gauge, Histogram, Info, Summary
+from .metrics import (
+    Counter, disable_created_metrics, enable_created_metrics, Enum, Gauge,
+    Histogram, Info, Summary,
+)
 from .metrics_core import Metric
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
This seems like a fine solution for now.

Hopefully the new created timestamp handling gets implemented in upstream Prometheus soon. Then there will be less of a need for this, as well as making the data useful.

csmarchbanks:
Agreed, I am looking forward to that feature in Prometheus. Thanks for the review!

PR #978, opened 2023-11-13
fix: error in determining timestamp  "less than"

weidongkl:
```
__lt__ is the negation of __gt__ by default.
so samples.Timestamp(1, 1) > samples.Timestamp(1, 1) is false
and samples.Timestamp(1, 1) < samples.Timestamp(1, 1) is true. but  samples.Timestamp(1, 1) < samples.Timestamp(1, 1) should be false too.
add __lt__ func to fix the bug

```


weidongkl:
/

PR #984, opened 2023-12-03
Add feature to set datetime object as value for gauge collector

AnthonyCalo:
Was working on project where the value of my gauge was to be a datetime. No built in features. This is simple update and has been tested in production to work well.

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -1,7 +1,7 @@
+from datetime import datetime
 import os
 from threading import Lock
 import time
-from datetime import datetime
 import types
 from typing import (
     Any, Callable, Dict, Iterable, List, Literal, Optional, Sequence, Tuple,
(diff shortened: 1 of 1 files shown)


csmarchbanks (commented):
Hmm, couldn't this be done with `.set(dt.timestamp())`? I think the reason we don't have this is that is not much harder to write and keeps our interface simple with only taking floats.

AnthonyCalo:
> Thank you for the contribution! Couldn't this be done with `.set(dt.timestamp())` though? I think the reason we don't have this is that is not much harder to write and keeps our interface simple with only taking floats.

When working with grafana for visuailzations they expect milliseconds rather than epoch seconds. So you have to multiply by 1000. in this case we drop decimals and make an integer. If using grafana and want to sort or visualize events the best way is to set the gauge value as the timestamp of the event. 

I understand a case against merging if it can be confusing. However it will be useful for some people.

Issue #916, opened 2023-05-31
.remove  and .clear not working with multiprocessing mode


MarcoRizk:
Trying to remove a given set of labels for a metric doesn't work.
The label value is still readable by the collector after it's presumed removal ?
any suggestions ?

hatamiarash7:
I have this problem too

csmarchbanks:
Yep, this is definitely an issue. At a minimum we should document that they do not work in the multiprocessing section of the readme, but ideally we could figure out a way to support remove/clear in multiprocess.

yctomwang:
@csmarchbanks Hi Chris, I also run into this issue. After looking around the code base I think its because of the disconnection between the remove of Normal Metrics and the multiprocess ones. I thinking add a warning message aswell as in the documentation/README would benefit a lot of others. Happy to put together a PR for this. Thanks again for maintiaining such an awesome libaray!

csmarchbanks:
:+1: That would be great, thank you!

yctomwang:
@csmarchbanks Hi Chris, I got the PR in, would be nice if you could review it when you have free time, thanks again.

PR #988, opened 2023-12-16
feat: Add an option to delete metrics from multiproc files by the provided key

alekseiapa:
**Changes Made:** This merge request introduces a new feature in the multiprocess mode that allows metrics deletion. The new functionality is encapsulated in a method called `delete_value` within the `MmapedDict` class.

**Motivation:** Until now, there was no option to remove individual metrics once they are written in multiprocess mode. It limited the flexibility of our system in dynamic scenarios where metrics might need to be removed post-write based on certain conditions. This created unnecessary clutter and could lead to memory inefficiencies over time.

**How it Works:** The `delete_value` method takes a key as an argument, which corresponds to the metric you want to delete. If the metric exists, it removes it from the dictionary and decreases the overall used memory size accordingly.

**Testing:** New unittest cases have been added in `TestMmapedDict`. These tests verify correct deletion of existing metrics, handle attempts of deleting non-existing keys, check the unaffected state of other metrics upon deletion of one, and write-new after delete operations.

**Impact:** The new deletion feature will allow more flexible and efficient handling of metrics in multiprocess mode. However, it should be used judiciously considering the operation involves memory manipulation.

The update doesn't affect the existing feature set and maintains backward compatibility.

### DIFF
--- a/prometheus_client/values.py
+++ b/prometheus_client/values.py
@@ -35,6 +35,10 @@ class MutexValue:
         with self._lock:
             return self._exemplar
 
+    # used in multiproc mode
+    def delete(self):
+        pass
+
 
 def MultiProcessValue(process_identifier=os.getpid):
     """Returns a MmapedValue class based on a process_identifier function.
@@ -121,6 +125,31 @@ def MultiProcessValue(process_identifier=os.getpid):
         def get_exemplar(self):
             # TODO: Implement exemplars for multiprocess mode.
             return None
+
+        def delete(self):
+            with lock:
+                # Create a new MmapedDict temporarily
+                temp_file = MmapedDict(self._file._fname + ".tmp")
+
+                # Copy all key-value pairs, excluding the one to delete
+                for k, v, ts in self._file.read_all_values():
+                    if k != self._key:
+                        temp_file.write_value(k, v, ts)
+
+                temp_file.close()
+                self._file.close()
+
+                # Atomically perform the act of deleting the old file
+                # and replacing it with the newly constructed one
+                os.rename(self._file._fname + ".tmp", self._file._fname)
+
+                # Reloading the dictionary for accessing fresher data
+                self._file = MmapedDict(self._file._fname)
+
+                # Update the corresponding entry in files dictionary
+                for prefix, file in files.items():
+                    if file._fname == self._file._fname:
+                        files[prefix] = self._file
 
     return MmapedValue
 
--- a/tests/test_multiprocess.py
+++ b/tests/test_multiprocess.py
@@ -381,6 +381,54 @@ class TestMultiProcess(unittest.TestCase):
             os.path.join(self.tempdir, 'gauge_livesum_9999999.db'),
         ]))
 
+    def test_delete_metrics(self):
+        c = Counter('c', 'help', registry=None)
+        g = Gauge('g', 'help', registry=None)
+
+        c.inc(1)
+        g.set(5)
+
+        # Test Counter deletion
+        self.assertEqual(1, self.registry.get_sample_value('c_total'))
+        c._value.delete()
+        self.assertIsNone(self.registry.get_sample_value('c_total'))
+
+        # Test Gauge deletion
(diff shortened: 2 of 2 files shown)


alekseiapa:
Hi @csmarchbanks ! May I ask you to take a look at the PR ? Thanks in advance.

csmarchbanks (commented):
🤔 I don't think this is safe, if a scrape came along on a different process while writing the left/right data I think it could end up being corrupted?

If that is true a different approach could be taken, for example we could use a specific NaN value to signal that a metric has been deleted, possibly `0x7ff0000000000002` since that is the stale NaN used in Prometheus. That wouldn't lower the disk usage but would be safer.

alekseiapa:
> 🤔 I don't think this is safe, if a scrape came along on a different process while writing the left/right data I think it could end up being corrupted?
> 
> If that is true a different approach could be taken, for example we could use a specific NaN value to signal that a metric has been deleted, possibly `0x7ff0000000000002` since that is the stale NaN used in Prometheus. That wouldn't lower the disk usage but would be safer.

@csmarchbanks  Thank you for looking into this. I do not think that would be corrupted because we lock the deletion process of a specific metric. In addition, I am making this request so that I do not need to monkey patch the `values` module with a customized method. I have been using this approach for a few months and have not encountered any problems yet.

csmarchbanks:
I believe the lock is only in the process updating, not in the process doing the read so unless we can guarantee an atomic write I have concerns about this. It would be fairly rare to have happen (metric deletion during a scrape).

Another approach would be to create a new file during deletion and then move it to the same name as the old file, like what is down for the textfile. That should be atomic and would clean up disk space in the long term (though double it in the short term).

alekseiapa:
@csmarchbanks  Hi. Thank you for your review. I went with the temporary file approach. May I ask you to take another look?

alekseiapa:
@csmarchbanks Hi. Could you please review the changes ?

prometheus_client/values.py
```diff
@@ -122,6 +126,26 @@ def get_exemplar(self):
...
+
+                temp_file.close()
+                self._file.close()
+
+                # Atomically perform the act of deleting the old file
+                # and replacing it with the newly constructed one
+                os.rename(self._file._fname + ".tmp", self._file._fname)
+
+                # Reloading the dictionary for accessing fresher data
+                self._file = MmapedDict(self._file._fname)
```
csmarchbanks, line 147: I think `files` also needs to be updated to point to the new file.

csmarchbanks (commented):
Hello! I left one comment, but haven't had the time to fully test this yet due to some leave. Hopefully next week.

prometheus_client/values.py
alekseiapa: Hi @csmarchbanks . I agree.

alekseiapa:
Hi, @csmarchbanks  . Thanks for the review. Could you please give it another look ? Thanks in advance.

alekseiapa:
Hi @csmarchbanks . The suggested change has been applied

anChaOs:
Hi, @csmarchbanks .

I think this pr is a possible solution to solve this issue(#916 ).
When it is necessary to call .remove() and .clear(), just call .delete() first.

Is that right?

csmarchbanks:
Yes, it would be a solution to those, or at least `.remove()`. If we can figure out the edge cases.

alekseiapa:
> Hi, @csmarchbanks .
> 
> I think this pr is a possible solution to solve this issue(#916 ). When it is necessary to call .remove() and .clear(), just call .delete() first.
> 
> Is that right?

Hi @anChaOs @csmarchbanks I use this approach in my env for the Gauge type of metrics only. So should I leave the functionality for this type only?

csmarchbanks:
Probably only Gauge to start with. I am also not sure if it should be implemented for any of the Gauge aggregations? Those still behave similarly where a removal of just one pid might be confusing when the output is aggregated.

I am also torn about just implementing `remove` instead of the new `delete` function. Keeping `remove` might be nice to avoid any new surface area?

Issue #982, opened 2023-11-24
Allow tls_auth_handler to support client side verification only


Roymprog:
The `tls_auth_handler` supports setting up a mTLS connection with the Prometheus push gateway, but I want only to verify the certificate of the server. In other words, I wish to skip setting `certfile` and `keyfile` in the `tls_auth_handler` method when the protocol id `ssl.PROTOCOL_TLS_CLIENT`.

The authentication on the server side I will handle differently.

PR #994, opened 2024-01-15
Add function to use TLS when pushing to gateway

Roymprog:
Closes https://github.com/prometheus/client_python/issues/982

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -462,7 +462,7 @@ def tls_auth_handler(
     ssl.CERT_REQUIRED and SSLContext.check_hostname by default. This can be
     disabled by setting insecure_skip_verify to True.
 
-    Both this handler and the TLS feature on pushgateay are experimental."""
+    Both this handler and the TLS feature on pushgateway are experimental."""
     context = ssl.SSLContext(protocol=protocol)
     if cafile is not None:
         context.load_verify_locations(cafile)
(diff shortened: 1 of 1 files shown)


csmarchbanks (commented):
Rather than create a second function could we add another argument for verify_mode to `tls_auth_handler`?

Roymprog:
We could, the reason I didn't do that was the name of the function. The name `tls_auth_handler` implies it does authentication. The function becomes confusing if you allow to not provide your own TLS certificate and only use it to verify the certificate of the server. 

I'd propose keeping one function called `tls_handler`, but that would be a breaking change.

csmarchbanks (commented):
Ahh right, that makes sense, and I think having an auth specific handler is nice.

What would this function provide over just using an `https` scheme in `push_to_gateway`? I guess the use case is for if you need to specify the cafile and verify mode?

Roymprog:
> What would this function provide over just using an https scheme in push_to_gateway? I guess the use case is for if you need to specify the cafile and verify mode?

That can be done. It's how I'm doing it right now. If you want to go that route, it would be better to make `_make_handler` a public function. Otherwise users would need to copy the boilerplate code from that function to create the handler. `HTTPShandler` cannot be passed to `push_to_gateway` directly, as it expects a callable (which `HTTPShandler`) is not.

csmarchbanks:
I might be missing something as well, but perhaps this is a bug? I believe `push_to_gateway` is supposed to just work for https, and you are saying that you need to pass a custom handler to make it work? I wonder if we just need to use the scheme of the url to choose HTTP vs HTTPS in `default_handler`?

Roymprog:
The `default_handler` works with HTTPS by default. The issue I have is that I need to use a custom SSL context because I need to load a custom trust location. I cannot do that through the `default_handler`. So, it's unnecessary to specify the scheme. 
What would be nice in that case is to have an optional SSL context argument for the `default_handler`.

csmarchbanks:
That makes sense, thank you.

Adding a custom context to `default_handler` as you suggest seems like it could be a good, and more general solution? That way users could customize the context as they need without us supporting every argument.

I am still thinking about it some and happy to hear other opinions, or advantages/disadvantages of that approach to what you have here. My main concern with this PR is that having two tls related handlers will be confusing.

PR #1005, opened 2024-02-06
Reset counter

### LINKED ISSUE #995: Provide a way to reset a Counter metric
Hi. My program performs some lengthy process of many iterations, and I use a Counter to track current iteration. However, it is normal for the process to finish at some point, and then the user can start it again, without restarting the app. 

> A counter is a cumulative metric that represents a single monotonically increasing counter whose value can only increase or be reset to zero on restart.
(https://prometheus.io/docs/concepts/metric_types/)

In my case it's not the whole app that restarts, but only the calculation procedure, but I think in spirit it's still a Counter's job, not a Gauge (a value that only increments and resets to zero), but there is currently no way in the library to set the Counter to zero.

Would it be possible to add "reset" method to a Counter class?

positron96:
Adds `.reset()` method to Counter metric.

Closes #995.

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -292,6 +292,12 @@ class Counter(MetricWrapperBase):
         # Count only one type of exception
         with c.count_exceptions(ValueError):
             pass
+            
+    You can also reset the counter to zero in case your logical "process" restarts
+    without restarting the actual python process.
+
+       c.reset()
+
     """
     _type = 'counter'
 
@@ -313,8 +319,7 @@ class Counter(MetricWrapperBase):
     def reset(self) -> None:
         """Reset the metric to zero."""
         self._value.set(0)
-        if _use_created:
-            self._created = time.time()
+        self._created = time.time()
 
     def count_exceptions(self, exception: Union[Type[BaseException], Tuple[Type[BaseException], ...]] = Exception) -> ExceptionCounter:
         """Count exceptions in a block of code or function.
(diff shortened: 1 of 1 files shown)


prometheus_client/metrics.py
```diff
@@ -310,6 +316,11 @@ def inc(self, amount: float = 1, exemplar: Optional[Dict[str, str]] = None) -> N
             _validate_exemplar(exemplar)
             self._value.set_exemplar(Exemplar(exemplar, amount, time.time()))
 
+    def reset(self) -> None:
+        """Reset the metric to zero."""
```
csmarchbanks, line 320: It would be nice to have some of the context around logical process restarts here as well. Perhaps
```suggestion
        """Reset the counter to zero. Use this when a logical process restarts without restarting the actual python process."""
```
?

csmarchbanks (commented):
Thanks! One small comment, and would you be willing to add a test as well please?

positron96:
> Thanks! One small comment, and would you be willing to add a test as well please

I'll see what I can about the test!

positron96:
@csmarchbanks do we need a test for "created" sample? I'm not entirely sure how to test that, as it's tied to system time

csmarchbanks:
Ideally yes, I would say just make sure that the value for `_created` is larger after the reset than it was originally, we don't need to assert the exact value. A brief sleep before the reset should cause that to happen consistently.

csmarchbanks (approved):
Thanks!

PR #1006, opened 2024-02-11
wsgi server: address family discovery is not quite right

jclulow:
The code introduced to improve binding to an IPv6 address is based on similar code in Python itself, but is missing some critical arguments to the **socket.getaddrinfo()** call: in particular, the socket type must be set to **SOCK_STREAM**, because we want a TCP connection; the **AI_PASSIVE** flag should also be passed because we intend to use the result for binding a listen socket rather than making an outbound connection.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -154,7 +154,7 @@ def _get_best_family(address, port):
     # binding an ipv6 address is requested.
     # This function is based on what upstream python did for http.server
     # in https://github.com/python/cpython/pull/11767
-    infos = socket.getaddrinfo(address, port)
+    infos = socket.getaddrinfo(address, port, type=socket.SOCK_STREAM, flags=socket.AI_PASSIVE)
     family, _, _, _, sockaddr = next(iter(infos))
     return family, sockaddr[0]
 


jclulow:
By way of example, this is the output with and without the flags on an Ubuntu 22.04 system:

```python
#!/usr/bin/env python3

import socket

print("WITH type set:")
print(socket.getaddrinfo("0.0.0.0", 9009,
    type=socket.SOCK_STREAM, flags=socket.AI_PASSIVE))

print("WITHOUT type set:")
print(socket.getaddrinfo("0.0.0.0", 9009))
```

```
WITH type set:
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('0.0.0.0', 9009))]
WITHOUT type set:
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('0.0.0.0', 9009)), (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_DGRAM: 2>, 17, '', ('0.0.0.0', 9009)), (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_RAW: 3>, 0, '', ('0.0.0.0', 9009))]
```

On some other platforms, **getaddrinfo()** _requires_ at least the socket type to function as is expected here; e.g.,

```
WITH type set:
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 2>, 6, '', ('0.0.0.0', 9009))]
WITHOUT type set:
  File "/tmp/sigh_test.py", line 10, in <module>
    print(socket.getaddrinfo("0.0.0.0", 9009))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/socket.py", line 962, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
socket.gaierror: [Errno 9] service name not available for the specified socket type
```

csmarchbanks (approved):
Thanks!

PR #1007, opened 2024-02-14
remove shebangs in py files don't need it

kloczek:


dulmandakh (approved):

khaled5321 (approved):

prometheus_client/__init__.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
```
csmarchbanks, line 1: This file does of a main in it. It doesn't do much but I think this belongs here still.

Issue #1018, opened 2024-03-30
Network scanners report issues because all HTTP methods are supported


andy-maier:
A Prometheus exporter implemented with the prometheus_client package will respond to all HTTP methods, even though it is supposed to succeed only on GET /metrics. This behavior may be reported by network scanners as an issue.

Actual and proposed behavior:

| HTTP method | resource | actual behavior | proposed behavior |
|:----------- |:----------- |:------------- |:------------ |
| OPTIONS | /* | 200 + metrics | 204 + Allow: OPTIONS,GET, no body |
| OPTIONS | any other | 200 + metrics | 405 + Allow: OPTIONS,GET, no body |
| HEAD | /metrics | 200 + metrics | 204, no body |
| GET | /metrics | 200 + metrics | 200 + metrics |
| HEAD | any other | 200 + metrics | 404, no body |
| GET | any other | 200 + metrics | 404, no body |
| POST | any | 200 + metrics | 405 + Allow: OPTIONS,GET, no body |
| PATCH | any | 200 + metrics | 405 + Allow: OPTIONS,GET, no body |
| DELETE | any | 200 + metrics | 405 + Allow: OPTIONS,GET, no body |
| CONNECT | any | 200 + metrics | 405 + Allow: OPTIONS,GET, no body |
| TRACE | any | 200 + metrics | 405 + Allow: OPTIONS,GET, no body |
| any other | any | 200 + metrics | 501, no body |

I am using `prometheus_client.start_http_server()` for this.

The proposed behavior is my interpretation of the rules in HTTP 1.1 as described in [RFC7231](https://www.rfc-editor.org/rfc/rfc7231).

Reproduction:
* Start Prometheus exporter that uses `prometheus_client.start_http_server()`
* Issue against the IP/port of the exporter:
  ```
  curl -k -i -X <method> http://<ip>:<port>/<resource>
  ```
  
Unfortunately, the Prometheus exporter best practices at https://prometheus.io/docs/instrumenting/writing_exporters/ do not touch on this subject.

andy-maier:
I have tried to address this issue in PR #1019.

csmarchbanks:
Fixedin #1019

PR #1019, opened 2024-03-30
Reject invalid HTTP methods and resources

andy-maier:
This change addresses issue #1018 (currently, any HTTP method is handled by returning success and metrics data, which causes network scanners to report issues).

Note, this needs careful review w.r.t.:
* Does Prometheus issue any HTTP requests that are now rejected.
* Does HEAD need to be supported.
* Does the `/metrics` resource for which metrics are returned, need to be configurable by the users of this package.
* Does the strict handling need to be enabled/disabled by the users of this package.
* Are unit tests needed.

Note, the pinning of asgiref==3.6.0 removes a test error in the py3.8 tox environment (same fix that already existed for the pypy3.8 tox environment). I don't know why the error on py3.8 comes up in the first place. I guess someone more experienced than me needs to look at that.

For details, see the commit message.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -118,12 +118,24 @@ def make_wsgi_app(registry: CollectorRegistry = REGISTRY, disable_compression: b
         accept_header = environ.get('HTTP_ACCEPT')
         accept_encoding_header = environ.get('HTTP_ACCEPT_ENCODING')
         params = parse_qs(environ.get('QUERY_STRING', ''))
-        if environ['PATH_INFO'] == '/favicon.ico':
+        method = environ['REQUEST_METHOD']
+
+        if method == 'OPTIONS':
+            status = '200 OK'
+            headers = [('Allow', 'OPTIONS,GET')]
+            output = b''
+        elif method != 'GET':
+            status = '405 Method Not Allowed'
+            headers = [('Allow', 'OPTIONS,GET')]
+            output = '# HTTP {}: {}; use OPTIONS or GET\n'.format(status, method).encode()
+        elif environ['PATH_INFO'] == '/favicon.ico':
             # Serve empty response for browsers
             status = '200 OK'
             headers = [('', '')]
             output = b''
         else:
+            # Note: For backwards compatibility, the URI path for GET is not
+            # constrained to the documented /metrics, but any path is allowed.
             # Bake output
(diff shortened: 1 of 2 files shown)


prometheus_client/exposition.py
```diff
@@ -118,11 +118,29 @@ def prometheus_app(environ, start_response):
...
+        elif environ['REQUEST_METHOD'] != 'GET':
+            status = '405 Method Not Allowed'
+            headers = [('Allow', 'OPTIONS,GET')]
+            output = '# HTTP {}: {}; use OPTIONS or GET\n'.format(status, environ['REQUEST_METHOD']).encode()
+        elif environ['PATH_INFO'] == '/favicon.ico':
             # Serve empty response for browsers
             status = '200 OK'
             headers = [('', '')]
             output = b''
+        elif environ['PATH_INFO'].strip('/') not in ('', 'metrics'):
```
csmarchbanks, line 140: This is the breaking change I am most worried about, some users might be using non-standard paths, or even just `/` to get their metrics as that all works right now and we don't require `/metrics`. What would you think of omitting this check?

csmarchbanks (commented):
I double checked and Prometheus only issues GET requests and there is no way to override that, so I think blocking non-GET methods is reasonable.

Would you be willing to move the test fix into a separate PR?

prometheus_client/exposition.py
kakkoyun: I agree with this. Even though `/metrics` is the convention, it's not enforced. e.g. for the client_golang, one can specify anything when they register a handler https://github.com/prometheus/client_golang/blob/e133e490296d2ff915bfc23cdec87c235ad36ef3/examples/simple/main.go#L43

andy-maier: I can move the test fix into a aeparate PR.

andy-maier: I was also unsure about the path checking and I can remove it.

andy-maier: I have now updated the PR to remove the path check for GET. For consistency, I have also removed the path check for OPTIONS.

@kakkoyun @csmarchbanks Could you please review again, and merge if you agree?

andy-maier:
@csmarchbanks I moved the test fix into its own PR #1053 .

csmarchbanks (approved):
LGTM, thanks for the updates!

PR #1029, opened 2024-05-10
Optimization: Use str join to replace if

manas-rust:
Use str join to replace if

```python
full_name = "_".join(filter(None, [namespace, subsystem, name]))
```

image


manas-rust:
linked： https://github.com/prometheus/client_python/pull/1034/files

PR #1031, opened 2024-05-11
Optimization: Provide code readability

manas-rust:


csmarchbanks (commented):
You have made 4 tiny PRs, I would consider them if you combine the PRs into one, but otherwise these look like auto-generated changes with little value.

manas-rust:
> You have made 4 tiny PRs, I would consider them if you combine the PRs into one, and especially if we add a lint step to the pipeline, but otherwise these look like auto-generated changes with little value.

linked： https://github.com/prometheus/client_python/pull/1034/files

PR #1032, opened 2024-05-13
chore: refactor label parser

sylwit:
Refactor label parser to make it more readable and and faster


prometheus_client/parser.py
```diff
@@ -37,57 +37,29 @@ def _replace_escaping(s: str) -> str:
     return ESCAPING_RE.sub(replace_escape_sequence, s)
 
 
-def _is_character_escaped(s: str, charpos: int) -> bool:
```
sylwit, line 40: this function is moved to openmetrics/parser.py

PR #1038, opened 2024-05-27
Fix timestamp comparison

### LINKED ISSUE #1037: Timestamp comparison is wrong
Currently, `samples.Timestamp` implementation for comparisons is:

```
    def __gt__(self, other: "Timestamp") -> bool:
        return self.sec > other.sec or self.nsec > other.nsec

    def __lt__(self, other: "Timestamp") -> bool:
        return self.sec < other.sec or self.nsec < other.nsec
```

what might not compare correctly. Example:

```
>>> a = Timestamp(2, 5)
>>> b = Timestamp(3, 4)
>>> a > b
True
```

In the example, the first part is `2<3`, what is false, forcing to perform the next part, what is `5>4`, returning `True`.

A good implementation could be:
```
    def __gt__(self, other: "Timestamp") -> bool:
        if self.sec == other.sec:
            return self.nsec > other.nsec
        return self.sec > other.sec

    def __lt__(self, other: "Timestamp") -> bool:
        if self.sec == other.sec:
            return self.nsec < other.nsec
        return self.sec < other.sec
```

Another option, perhaps even better, is to rely on python's tuples implementation:
```
    def __gt__(self, other: "Timestamp") -> bool:
        return (self.sec, self.nsec) > (other.sec, other.nsec)

    def __lt__(self, other: "Timestamp") -> bool:
        return (self.sec, self.nsec) < (other.sec, other.nsec)
```

magmax:
Fixes #1037, where comparison between `Timestamp` were wrong.


csmarchbanks (approved):
Thanks!

PR #1040, opened 2024-06-18
Add support for native histograms in OM parser

vesari:
This PR adds an OM parser for native histograms, as a first step towards the implementation of the following proposal https://github.com/prometheus/proposals/pull/32 . The next steps (as far as this repo is concerned) would be the OM exposition and then obviously the implementation of native histograms generation/writing logic.

### DIFF
--- a/prometheus_client/bridge/graphite.py
+++ b/prometheus_client/bridge/graphite.py
@@ -20,6 +20,13 @@ def _sanitize(s):
     return _INVALID_GRAPHITE_CHARS.sub('_', s)
 
 
+def safe_float_convert(value):
+    try:
+        return float(value)
+    except ValueError:
+        return None
+
+
 class _RegularPush(threading.Thread):
     def __init__(self, pusher, interval, prefix):
         super().__init__()
@@ -82,7 +89,9 @@ class GraphiteBridge:
                             for k, v in sorted(s.labels.items())])
                 else:
                     labelstr = ''
-                output.append(f'{prefixstr}{_sanitize(s.name)}{labelstr} {float(s.value)} {now}\n')
+                # using a safe float convert on s.value as a temporary workaround while figuring out what to do
+                # in case value is a native histogram structured value, if that's ever a possibility
+                output.append(f'{prefixstr}{_sanitize(s.name)}{labelstr} {safe_float_convert(s.value)} {now}\n')
 
         conn = socket.create_connection(self._address, self._timeout)
         conn.sendall(''.join(output).encode('ascii'))
@@ -92,3 +101,4 @@ class GraphiteBridge:
         t = _RegularPush(self, interval, prefix)
         t.daemon = True
         t.start()
+    
--- a/prometheus_client/bridge/graphite.py
+++ b/prometheus_client/bridge/graphite.py
@@ -20,13 +20,6 @@ def _sanitize(s):
     return _INVALID_GRAPHITE_CHARS.sub('_', s)
 
 
-def safe_float_convert(value):
-    try:
-        return float(value)
-    except ValueError:
-        return None
-
-
 class _RegularPush(threading.Thread):
     def __init__(self, pusher, interval, prefix):
         super().__init__()
@@ -89,9 +82,7 @@ class GraphiteBridge:
                             for k, v in sorted(s.labels.items())])
                 else:
                     labelstr = ''
-                # using a safe float convert on s.value as a temporary workaround while figuring out what to do
-                # in case value is a native histogram structured value, if that's ever a possibility
-                output.append(f'{prefixstr}{_sanitize(s.name)}{labelstr} {safe_float_convert(s.value)} {now}\n')
+                output.append(f'{prefixstr}{_sanitize(s.name)}{labelstr} {float(s.value)} {now}\n')
 
         conn = socket.create_connection(self._address, self._timeout)
         conn.sendall(''.join(output).encode('ascii'))
--- a/prometheus_client/bridge/graphite.py
+++ b/prometheus_client/bridge/graphite.py
@@ -92,4 +92,3 @@ class GraphiteBridge:
         t = _RegularPush(self, interval, prefix)
         t.daemon = True
         t.start()
-    
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -1,4 +1,3 @@
-from datetime import timedelta
 import os
 from threading import Lock
 import time
@@ -16,7 +15,7 @@ from .metrics_core import (
     RESERVED_METRIC_LABEL_NAME_RE,
 )
 from .registry import Collector, CollectorRegistry, REGISTRY
-from .samples import BucketSpan, Exemplar, NativeHistStructValue, Sample
+from .samples import Exemplar, Sample
 from .utils import floatToGoString, INF
 
 T = TypeVar('T', bound='MetricWrapperBase')
@@ -598,7 +597,7 @@ class Histogram(MetricWrapperBase):
                  buckets: Sequence[Union[float, str]] = DEFAULT_BUCKETS,
                  # native_hist_schema: Optional[int] = None, # create this dynamically?
                  # native_hist_bucket_fact: Optional[float] = None,
-                  # native_hist_zero_threshold: Optional[float] = None,
+                 # native_hist_zero_threshold: Optional[float] = None,
                  # native_hist_max_bucket_num: Optional[int] = None,
                  # native_hist_min_reset_dur: Optional[timedelta] = None,
                  # native_hist_max_zero_threshold: Optional[float] = None,
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -595,7 +595,7 @@ class Histogram(MetricWrapperBase):
                  registry: Optional[CollectorRegistry] = REGISTRY,
                  _labelvalues: Optional[Sequence[str]] = None,
                  buckets: Sequence[Union[float, str]] = DEFAULT_BUCKETS,
-                 # native_hist_schema: Optional[int] = None, # create this dynamically?
+                 # native_hist_schema: Optional[int] = None,
                  # native_hist_bucket_fact: Optional[float] = None,
                  # native_hist_zero_threshold: Optional[float] = None,
                  # native_hist_max_bucket_num: Optional[int] = None,
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -111,8 +111,8 @@ class MetricWrapperBase(Collector):
 
     def collect(self) -> Iterable[Metric]:
         metric = self._get_metric()
-        for suffix, labels, value, timestamp, exemplar in self._samples():
-            metric.add_sample(self._name + suffix, labels, value, timestamp, exemplar)
+        for suffix, labels, value, timestamp, exemplar, native_histogram_value in self._samples():
+            metric.add_sample(self._name + suffix, labels, value, timestamp, exemplar, native_histogram_value)
         return [metric]
 
     def __str__(self) -> str:
@@ -246,8 +246,8 @@ class MetricWrapperBase(Collector):
             metrics = self._metrics.copy()
         for labels, metric in metrics.items():
             series_labels = list(zip(self._labelnames, labels))
-            for suffix, sample_labels, value, timestamp, exemplar in metric._samples():
-                yield Sample(suffix, dict(series_labels + list(sample_labels.items())), value, timestamp, exemplar)
+            for suffix, sample_labels, value, timestamp, exemplar, native_histogram_value in metric._samples():
+                yield Sample(suffix, dict(series_labels + list(sample_labels.items())), value, timestamp, exemplar, native_histogram_value)
 
     def _child_samples(self) -> Iterable[Sample]:  # pragma: no cover
         raise NotImplementedError('_child_samples() must be implemented by %r' % self)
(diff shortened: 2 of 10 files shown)


prometheus_client/metrics.py
```diff
@@ -595,6 +595,14 @@ def __init__(self,
                  registry: Optional[CollectorRegistry] = REGISTRY,
                  _labelvalues: Optional[Sequence[str]] = None,
                  buckets: Sequence[Union[float, str]] = DEFAULT_BUCKETS,
+                 # native_hist_schema: Optional[int] = None, # create this dynamically?
```
csmarchbanks, line 598: I believe the internal code should create the schema, and may even change the schema value on occassion.

csmarchbanks (commented):
Just a quick read through, but I was thinking maybe we limit the changes in the initial PR to just the parser changes? We could get that in and start iterating on it, and it will then be useful for testing the exposition changes.

vesari:
> Just a quick read through, but I was thinking maybe we limit the changes in the initial PR to just the parser changes? We could get that in and start iterating on it, and it will then be useful for testing the exposition changes.

I totally agree!

prometheus_client/openmetrics/parser.py
```diff
@@ -364,6 +366,100 @@ def _parse_remaining_text(text):
     return val, ts, exemplar
 
 
+def _parse_nh_sample(text, suffixes):
+    labels_start = text.find("{")
+    # check if it's a native histogram with labels
+    re_nh_without_labels = re.compile(r'^[^{} ]+ {[^{}]+}$')
+    re_nh_with_labels = re.compile(r'[^{} ]+{[^{}]+} {[^{}]+}$')
```
csmarchbanks, line 373: I think we could probably do without a regex. What if we just read the metric name, then determine if there are labels by seeing if it is a space or a curly brace that ends the metric name? If it is a space then we know the value is next, if not we need to parse labels.

Once finished with the labels we can know if it is a native histogram or not by seeing if the value starts with a curly brace or not.

prometheus_client/openmetrics/parser.py
```diff
@@ -364,6 +366,100 @@ def _parse_remaining_text(text):
...
+
+    try:
+        pos_spans_text = spans['positive_spans']
+    except KeyError:
+        pos_spans = None
+    else:
+        elems = pos_spans_text.split(',')
+        arg1 = [int(x) for x in elems[0].split(':')]
+        arg2 = [int(x) for x in elems[1].split(':')]
+        pos_spans = (BucketSpan(arg1[0], arg1[1]), BucketSpan(arg2[0], arg2[1]))
```
csmarchbanks, line 426: I would avoid the else block and move these into the try block. That keeps all the happy path/text parsing code together.

prometheus_client/registry.py
```diff
@@ -128,7 +129,7 @@ def _target_info_metric(self):
         m.add_sample('target_info', self._target_info, 1)
         return m
 
-    def get_sample_value(self, name: str, labels: Optional[Dict[str, str]] = None) -> Optional[float]:
+    def get_sample_value(self, name: str, labels: Optional[Dict[str, str]] = None) -> Optional[Union[float, NativeHistStructValue]]:
```
csmarchbanks, line 132: Does this need to change yet? It would be nice to have no/minimal public interface changes since we won't be using these in the registry quite yet.

prometheus_client/samples.py
```diff
@@ -48,6 +65,6 @@ class Exemplar(NamedTuple):
 class Sample(NamedTuple):
     name: str
     labels: Dict[str, str]
-    value: float
+    value: Union[float, NativeHistStructValue]
```
csmarchbanks, line 68: Trying to decide if this should be a union or if we should have a separate field for `histogram_value` or something like that.

prometheus_client/openmetrics/parser.py
```diff
@@ -406,37 +502,38 @@ def do_checks():
...
-            value = s.value
-        elif suffix in ['_count', '_gcount']:
-            count = s.value
-        elif suffix in ['_sum']:
-            has_sum = True
-        elif suffix in ['_gsum']:
-            has_gsum = True
-            if s.value < 0:
-                has_negative_gsum = True
+        if len(suffix) != 0:
```
csmarchbanks, line 505: Rather than nesting everything further in what about a `continue` if `len(suffix) is 0`?

prometheus_client/openmetrics/parser.py
```diff
@@ -537,8 +636,18 @@ def build_metric(name, documentation, typ, unit, samples):
             else:
                 raise ValueError("Invalid line: " + line)
         else:
-            sample = _parse_sample(line)
-            if sample.name not in allowed_names:
+            if typ == 'histogram':
+                # set to true to account for native histograms naming exceptions/sanitizing differences
+                is_nh = True
+                sample = _parse_nh_sample(line, tuple(type_suffixes['histogram']))
```
csmarchbanks, line 642: Maybe for a future PR, but it feels a bit wrong to have to parse the entire sample as a native histogram, then bail and parse a normal sample. I wonder if instead we could have one function that parses everything consistently. I think it should be possible if we start on the left and work our way right parsing the name, then optional labels, then a value (either float or histogram - whitespace separates it from other fields), etc...

prometheus_client/metrics.py
```diff
@@ -595,6 +595,14 @@ def __init__(self,
...
                  _labelvalues: Optional[Sequence[str]] = None,
                  buckets: Sequence[Union[float, str]] = DEFAULT_BUCKETS,
+                 # native_hist_schema: Optional[int] = None,
+                 # native_hist_bucket_fact: Optional[float] = None,
+                 # native_hist_zero_threshold: Optional[float] = None,
+                 # native_hist_max_bucket_num: Optional[int] = None,
+                 # native_hist_min_reset_dur: Optional[timedelta] = None,
+                 # native_hist_max_zero_threshold: Optional[float] = None,
+                 # native_hist_max_exemplars: Optional[int] = None,
+                 # native_hist_exemplar_TTL: Optional[timedelta] = None,
```
csmarchbanks, line 605: Let's remove these commented out lines until they are actually used.

csmarchbanks (commented):
Nice! I left a handful of comments, they probably don't all need to be addressed in this PR, but could be refactorings in future PRs as well.

Mainly I would like to see as few changes to other files, and especially function signatures, as possible. Even if that means we do some things like skipping values that are histograms for now. That way hopefully no one depends on some code/type changes we might update soon.

prometheus_client/bridge/graphite.py
```diff
@@ -82,7 +89,9 @@ def push(self, prefix: str = '') -> None:
                             for k, v in sorted(s.labels.items())])
                 else:
                     labelstr = ''
-                output.append(f'{prefixstr}{_sanitize(s.name)}{labelstr} {float(s.value)} {now}\n')
+                # using a safe float convert on s.value as a temporary workaround while figuring out what to do
+                # in case value is a native histogram structured value, if that's ever a possibility
+                output.append(f'{prefixstr}{_sanitize(s.name)}{labelstr} {safe_float_convert(s.value)} {now}\n')
```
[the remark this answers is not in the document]
vesari: You're absolutely right. The thing is the `mypy` linter fails without this change. That's due to the native histogram value being a union. I noticed that only when I pushed the PR for the first time

prometheus_client/registry.py
vesari: Same as for the Graphite bridge file. `mypy` linting fails otherwise.

prometheus_client/samples.py
vesari: Given the union's "side-effects" in files like `registry` etc, maybe it should be a separate field? My initial thinking with the union was to possibly avoid function signature changes I guess, but maybe it wouldn't be that problematic?

csmarchbanks: Yeah, let's try having a separate Optional field for the native histogram and see what it looks like. It also makes it nice to just check if the value is none or not for if a value is a native histogram.

prometheus_client/samples.py
```diff
@@ -34,6 +34,23 @@ def __lt__(self, other: "Timestamp") -> bool:
         return self.nsec < other.nsec if self.sec == other.sec else self.sec < other.sec
 
 
+class BucketSpan(NamedTuple):
+    offset: int
+    length: int
+
+
+class NativeHistStructValue(NamedTuple):
```
csmarchbanks, line 42: Just a thought, we could probably just call this `NativeHistogram` rather than adding `StructValue`. For initial implementation we might want to prefix with `_` as well to these new types to make it clear they are internal only and may change. Or add a comment to that effect.

prometheus_client/samples.py
vesari: I separated them, and added a field for native histogram. I added it for last to minimize the risk of breaking something.

prometheus_client/openmetrics/parser.py
vesari: Can I ask why you don’t want to use regexps? The big benefit I see with regexps is that they are robust and catch badly formatted input, I find it really cumbersome to implement the same level of validation without them. I tried, but I couldn’t succeed.
 
For example, something like this fails:

```
first_curly = text.find("{") 
if text[first_curly - 1:first_curly]== " ": 
	name = text[:first_curly-1] 
	#(check no suffixes) 
	if name.endswith(suffixes): 
		raise ValueError("the sample name of a native histogram with labels should have no suffixes", name) 
	nh_value = _parse_nh_struct(text[first_curly:]) 
	return Sample(name, None, None, None, None, nh_value) 
if text[first_curly - 1:first_curly]!= " ": 
	first_back_curly = text.find("}") 
	label_bag = text[first_curly+1:first_back_curly] 
	labels = _parse_labels(label_bag) name = text[:first_curly] 
	#(check no suffixes) 
	if name.endswith(suffixes): 
		raise ValueError("the sample name of a native histogram should have no suffixes", name) 
	nh_value = _parse_nh_struct(text[first_back_curly+1:]) 
	return Sample(name, labels, None, None, None, nh_value) 
else: return 

```


>    What if we just read the metric name, then determine if there are labels by seeing if it is a space or a curly brace that ends the metric name? If it is a space then we know the value is next, if not we need to parse labels.

I think this algorithm will be far more complicated when also handling potentially misformatted input.

prometheus_client/bridge/graphite.py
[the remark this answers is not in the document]
vesari: Solved after separating fields

prometheus_client/registry.py
vesari: Solved after separating fields

prometheus_client/openmetrics/parser.py
vesari: I'd be grateful if we could leave this for another iteration, considering also the difficulties I've mentioned about parsing native histograms alone

prometheus_client/bridge/graphite.py
```diff
@@ -92,3 +92,4 @@ def start(self, interval: float = 60.0, prefix: str = '') -> None:
         t = _RegularPush(self, interval, prefix)
         t.daemon = True
         t.start()
+
```
csmarchbanks, line 95: Nit, could you just remove this line so we don't have a needless diff/history entry?

prometheus_client/metrics_core.py
```diff
@@ -236,6 +236,7 @@ def __init__(self,
                  sum_value: Optional[float] = None,
                  labels: Optional[Sequence[str]] = None,
                  unit: str = '',
+                 native_hist_bucket_factor: Optional[float] = None
```
csmarchbanks, line 239: 🤔 Not sure if we need this here at all (it will need to be in `metrics.py`). This function is for custom collectors, and I think if someone is creating a native histogram for that case they will end up needing to define all the spans themself. For now I would say just leave it out.

prometheus_client/openmetrics/parser.py
csmarchbanks: Honestly, my aversion to regex might just be a personal thing :). They can end up being performance bottlenecks, but in Python that is probably not a huge deal, or could be dealt with in the future. If it is getting too complicated let's stick with the regex for now. I would prefer a working solution right now and optimize later if we need to.

prometheus_client/openmetrics/parser.py
```diff
@@ -364,6 +364,101 @@ def _parse_remaining_text(text):
     return val, ts, exemplar
 
 
+def _parse_nh_sample(text, suffixes):
+    labels_start = text.find("{")
+    # check if it's a native histogram with labels
+    re_nh_without_labels = re.compile(r'^[^{} ]+ {[^{}]+}$')
+    re_nh_with_labels = re.compile(r'[^{} ]+{[^{}]+} {[^{}]+}$')
+    print('we are matching \'{}\''.format(text))
```
csmarchbanks, line 372: We should remove the debug printing before merging, there are a couple other lines in this function as well.

prometheus_client/openmetrics/parser.py
csmarchbanks: Yep, happy to have it in a future PR!

prometheus_client/samples.py
```diff
@@ -34,6 +34,23 @@ def __lt__(self, other: "Timestamp") -> bool:
         return self.nsec < other.nsec if self.sec == other.sec else self.sec < other.sec
 
 
+class BucketSpan(NamedTuple):
```
csmarchbanks, line 37: Can we put some docstrings on these types saying that they are experimental and for users to expect them to change at any time? I just don't want someone to import them expecting stability and for us to break them while we iterate.

csmarchbanks (commented):
Just a couple of nits/removals and then I think this iteration is ready to merge! Thanks a bunch!

prometheus_client/bridge/graphite.py
vesari: Done!

prometheus_client/metrics_core.py
vesari: Yes, I deleted it.

prometheus_client/openmetrics/parser.py
vesari: I could see just another one, I hope I removed them all now XD

prometheus_client/samples.py
vesari: Sure, you had told me already, sorry. I added a warning on both those newly added classes.

csmarchbanks (approved):
Thanks for all the work on this!

Issue #925, opened 2023-06-13
Debug request logging.


cirvladimir:
I'm trying to debug why Prometheus isn't pulling metrics from my client. Is there some way I can get this library to log each incoming http request, including the IP of the request?

Ideally it would be something like start_http_server(log_level="debug"), and then each request will be printed to stdout.

csmarchbanks:
That is not available today, I am not sure how I feel about the feature as a whole. I guess it would depend on the added complexity as we try to keep `start_http_server` very minimal. It should be pretty easy to create your own http server that has logging how you want it and use something like `make_wsgi_app` from this library to serve the results though?

Usually there is information in the Prometheus server for if you are not seeing metrics as well, either looking in `Status > Targets`, or enabling debug logging on the Prometheus server.

cirvladimir:
I resorted to debugging by just standing up a simple http server with python.

I guess there are workarounds, and the feature I requested may not be useful to anyone else, so maybe this issue can be closed.

PR #1050, opened 2024-07-26
Add debug flag&handler for request

alex-ak1:
Add parameter `debug: bool` for `start_wsgi_server` and `start_http_server`.

If debug=True wsgi server prints every request to console.

Related to #925


prometheus_client/exposition.py
```diff
@@ -140,6 +140,13 @@ def log_message(self, format, *args):
         """Log nothing."""
 
 
+class _PrintHandler(WSGIRequestHandler):
```
csmarchbanks, line 143: I think we could just use WSGIRequestHandler instead of creating a new handler. It will log by default.

prometheus_client/exposition.py
```diff
@@ -210,6 +217,7 @@ def start_wsgi_server(
         client_capath: Optional[str] = None,
         protocol: int = ssl.PROTOCOL_TLS_SERVER,
         client_auth_required: bool = False,
+        debug: bool = False
```
csmarchbanks, line 220: Rather than just a debug flag, I wonder if we should let users specify the handler_class they are interested in using. That would be more flexible than having flags for everything.

csmarchbanks (commented):
Since it has now been a couple people who have asked for this I am ok with the feature. Just a couple comments, some docstring would be nice as well so someone knows what it does.

PR #1051, opened 2024-07-29
Removed CBC ciphers to address CVE-2013-0169 (LUCKY13)

andy-maier:
Details:

This change removes the following CBC ciphers from the default set of ciphers in order to address CVE-2013-0169 (LUCKY13):

  - ECDHE-ECDSA-AES256-SHA384
  - ECDHE-RSA-AES256-SHA384
  - ECDHE-ECDSA-AES128-SHA256
  - ECDHE-RSA-AES128-SHA256
  
This is done by listing them in the code, i.e. without any way to configure that by the user.

The LUCKY13 vulnerability was discovered using the testssl.sh tool. See also https://github.com/drwetter/testssl.sh/issues/2537

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -170,6 +170,22 @@ def _get_ssl_ctx(
     """Load context supports SSL."""
     ssl_cxt = ssl.SSLContext(protocol=protocol)
 
+    # The following chipers will be removed if the default cipher set contains
+    # them. The reason for each cipher is stated in the comment.
+    remove_cipher_names = [
+        "ECDHE-ECDSA-AES256-SHA384",  # is a CBC cipher (CVE-2013-0169)
+        "ECDHE-RSA-AES256-SHA384",    # is a CBC cipher (CVE-2013-0169)
+        "ECDHE-ECDSA-AES128-SHA256",  # is a CBC cipher (CVE-2013-0169)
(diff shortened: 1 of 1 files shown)


andy-maier:
If there is anything I can do in the PR for the failing CircleCI tests test-3.8 and test-3.9, let me know.

csmarchbanks (commented):
I don't love this library needing to worry about ciphers and keeping up to date with them. I think I would be ok with providing functionality to allow a user to specify the list of ciphers they want to support, but even that is getting into the territory where they should consider just standing up their own HTTP server instead of the lightweight one we provide.

andy-maier:
@csmarchbanks I know. I don't like it either, but the big advantage of the provided HTTP server is that it is built-in, simple to use, and it works great, even for our exporters that can have large amounts of metrics.

And I think you don't need to worry about checking for such things yourself, as users will bring it up, and may even fix it, like in this PR.

McTonderski (approved):

andy-maier:
The problematic ciphers have been removed in the Python ssl module in Python 3.10 (PR https://github.com/python/cpython/pull/25778).

We have vendored a locally patched version of prometheus-client that contains the fix in this PR here because we still support Python 3.8 in our fix stream version, but will use the fixed ssl module on Python>=3.10.

I am closing this PR.

PR #1053, opened 2024-07-30
Fixed asgiref version for Python 3.8

andy-maier:
Details:

* Needed to pin asgiref to ==3.6.0 also for py3.8 to circumvent the same error as for pypy3.8.

Note: This was previously in PR #1019 and was moved into its own PR.

Since the other PRs currently all succeed on Python 3.8, I'm not sure this change is needed anymore.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,8 +7,8 @@ deps =
     pytest
     attrs
     {py3.8,pypy3.8}: twisted
-    py3.8: asgiref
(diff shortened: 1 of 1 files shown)


csmarchbanks:
Yeah, since that PR and others are working my guess is we're ok without this now... Going to close but will open and merge if we run into issues in the future.

PR #1065, opened 2024-10-16
Fix AttributeError: Remove unnecessary 'decode' for Python 3 compatibility

amitaiturkel:
Fix AttributeError: Remove unnecessary 'decode' for Python 3 compatibility


csmarchbanks:
It looks like the test failures for 3.8 are real, I don't remember if something changed there. That said 3.8 was just EOL'd so maybe we can just remove it, but I would wait a bit longer before doing that.

PR #1066, opened 2024-10-16
Fix `write_to_textfile` leaves back temp files on errors

### LINKED ISSUE #1044: `write_to_textfile` leaves back temp files on errors
Hey.

A simple:
```
prometheus_client.write_to_textfile("", registry=registry)
```
which causes an exception:
```
  File "/home/calestyo/prj/prometheus-smart-array-collector/src/./smart-array.py", line 1185, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/calestyo/prj/prometheus-smart-array-collector/src/./smart-array.py", line 1176, in main
    prometheus_client.write_to_textfile("", registry=registry)
  File "/usr/lib/python3/dist-packages/prometheus_client/exposition.py", line 363, in write_to_textfile
    os.rename(tmppath, path)
FileNotFoundError: [Errno 2] No such file or directory: '.969277.139657657282624' -> ''
```
shows that the module doesn't properly clean up it's temp files on errors, as that file is left over afterwards.

Cheers,
Chris.

ethanschen:
Fix #1044

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -367,14 +367,19 @@ def write_to_textfile(path: str, registry: CollectorRegistry) -> None:
     This is intended for use with the Node exporter textfile collector.
     The path must end in .prom for the textfile collector to process it."""
     tmppath = f'{path}.{os.getpid()}.{threading.current_thread().ident}'
-    with open(tmppath, 'wb') as f:
-        f.write(generate_latest(registry))
+    try:
+        with open(tmppath, 'wb') as f:
+            f.write(generate_latest(registry))
 
-    # rename(2) is atomic but fails on Windows if the destination file exists
-    if os.name == 'nt':
-        os.replace(tmppath, path)
-    else:
-        os.rename(tmppath, path)
+        # rename(2) is atomic but fails on Windows if the destination file exists
+        if os.name == 'nt':
+            os.replace(tmppath, path)
+        else:
+            os.rename(tmppath, path)
+    except Exception:
+        if os.path.exists(tmppath):
+            os.remove(tmppath)
+        raise
 
 
 def _make_handler(


calestyo:
Looks in good to me.

The only two things I'd comment are:
- Catching `Exception` won’t e.g. catch `KeyboardInterrupt` ... but probably this doesn’t matter, since I guess there’s no clean-up in that case anyway (e.g. when metrics would be only half-written or so).
- At least in principle `os.remove(tmppath)` could raise an exception, too. ... One could catch that with another `try` block and, if caught, set the `__cause__` of that exception to the original. But probably overkill.

ethanschen:
> Looks in good to me.
> 
> The only two things I'd comment are:
> 
> * Catching `Exception` won’t e.g. catch `KeyboardInterrupt` ... but probably this doesn’t matter, since I guess there’s no clean-up in that case anyway (e.g. when metrics would be only half-written or so).
> * At least in principle `os.remove(tmppath)` could raise an exception, too. ... One could catch that with another `try` block and, if caught, set the `__cause__` of that exception to the original. But probably overkill.

Thanks.

Regarding your first point, I believe the `KeyboardInterrupt` exception would only occur when triggered manually by the user.I could add handling for this exception, but I'm not sure if it’s necessary for `prometheus_client`. Perhaps @csmarchbanks  could provide some insights.

As for the second point, if the `os.remove` method throws an exception, the error stack trace is quite clear, as shown below: 
```shell
  File "/WorkSpace/venv/lib/python3.8/site-packages/prometheus_client/exposition.py", line 378, in write_to_textfile
    os.rename(tmppath, path)
FileNotFoundError: [Errno 2] No such file or directory: '.80904.140704537089984' -> ''

During handling of the above exception, another exception occurred:

  File "x.py", line 6, in <module>
    write_to_textfile('', registry)
  File "/WorkSpace/venv/lib/python3.8/site-packages/prometheus_client/exposition.py", line 381, in write_to_textfile
    os.remove(1/0)
ZeroDivisionError: division by zero
```
Therefore, I think avoiding nested exception handling code might make it more elegant.

csmarchbanks (approved):
:+1: I agree that handling keyboard interrupts is not needed and keeping this as simple as possible is nice. Thanks!

Issue #1013, opened 2024-03-05
UTF-8: Implement support in Python client library


ywwg:
As in https://github.com/prometheus/client_golang/issues/1369 and https://github.com/prometheus/client_java/issues/916, the Python library needs to be updated to support UTF-8. 

Tasks:

- [x] Update exposition format parsers for the new syntax
- [x] Update PromQL parsers (if any) for the new syntax
- [x] Update content negotiation logic
- [x] Add flag to enable change in validation logic to check that metric and label names are UTF-8 valid instead of the old letters/numbers/underscores/colons set

For background and references see https://github.com/prometheus/prometheus/issues/13095

ywwg:
I will take this

ywwg:
A lot of the existing python parser does not at all take into account quoted characters and makes many calls to .index, .find, and regex matches that ignore quoted characters. This will break today if label values contain things like {, }, and =. So, this change will require a lot of rewriting to protect the existing code against such issues before it can even be updated to support quoted names

ywwg:
still need to do content negotiation!

PR #1070, opened 2024-11-06
Support UTF-8 in metric creation, parsing, and exposition

ywwg:
Adding support for UTF-8 required substantial reworking of the parsing code because the previous code relied on regexes and .index calls that would incorrectly find characters inside quote marks. Note that this means the previous code was already broken for parsing exposition text where characters like braces and hashes were in label values. I have added tests to exercise the edge cases I can think of.

Does not address content negotiation, which will be a followon.

part of https://github.com/prometheus/client_python/issues/1013

### DIFF
--- a/prometheus_client/openmetrics/exposition.py
+++ b/prometheus_client/openmetrics/exposition.py
@@ -2,6 +2,9 @@
 
 
 from ..utils import floatToGoString
+from ..validation import (
+    _is_valid_legacy_labelname, _is_valid_legacy_metric_name,
+)
 
 CONTENT_TYPE_LATEST = 'application/openmetrics-text; version=1.0.0; charset=utf-8'
 """Content type of the latest OpenMetrics text format"""
@@ -24,18 +27,27 @@ def generate_latest(registry):
         try:
             mname = metric.name
             output.append('# HELP {} {}\n'.format(
-                mname, metric.documentation.replace('\\', r'\\').replace('\n', r'\n').replace('"', r'\"')))
-            output.append(f'# TYPE {mname} {metric.type}\n')
+                escape_metric_name(mname), _escape(metric.documentation)))
+            output.append(f'# TYPE {escape_metric_name(mname)} {metric.type}\n')
             if metric.unit:
-                output.append(f'# UNIT {mname} {metric.unit}\n')
+                output.append(f'# UNIT {escape_metric_name(mname)} {metric.unit}\n')
             for s in metric.samples:
-                if s.labels:
-                    labelstr = '{{{0}}}'.format(','.join(
-                        ['{}="{}"'.format(
-                            k, v.replace('\\', r'\\').replace('\n', r'\n').replace('"', r'\"'))
-                            for k, v in sorted(s.labels.items())]))
+                if not _is_valid_legacy_metric_name(s.name):
+                    labelstr = escape_metric_name(s.name)
+                    if s.labels:
+                        labelstr += ', '
                 else:
                     labelstr = ''
+                
+                if s.labels:
+                    items = sorted(s.labels.items())
+                    labelstr += ','.join(
+                        ['{}="{}"'.format(
+                            escape_label_name(k), _escape(v))
+                            for k, v in items])
+                if labelstr:
+                    labelstr = "{" + labelstr + "}"
+                    
                 if s.exemplar:
                     if not _is_valid_exemplar_metric(metric, s):
                         raise ValueError(f"Metric {metric.name} has exemplars, but is not a histogram bucket or counter")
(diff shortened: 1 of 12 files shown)


tests/test_core.py
```diff
@@ -663,6 +675,12 @@ def test_invalid_names_raise(self):
         self.assertRaises(ValueError, Counter, 'c_total', '', labelnames=['a:b'])
         self.assertRaises(ValueError, Counter, 'c_total', '', labelnames=['__reserved'])
         self.assertRaises(ValueError, Summary, 'c_total', '', labelnames=['quantile'])
+        disable_legacy_validation()
+        self.assertRaises(ValueError, Counter, '', 'help')
```
ywwg, line 679: need to doublecheck these

ywwg:
I'll squash commits when we're done

prometheus_client/validation.py
```diff
@@ -0,0 +1,104 @@
...
+    _legacy_validation = False
+
+
+def enable_legacy_validation():
+    """Enable legacy name validation instead of allowing all UTF8 characters."""
+    global _legacy_validation
+    _legacy_validation = True
+
+
+def validate_metric_name(name: str) -> None:
```
csmarchbanks, line 34: Do we want these to be public now? Might be worth having them still prefixed with `_` but I also think they could be useful general functions for other libraries to use.

csmarchbanks (commented):
I gave this a first runthrough. Generally looking ok to me. A couple thoughts:
1. If you want to split it up I think refactoring validation + removing regex parsing could be good opportunities. Plus have some tests that catch the incorrect behavior you mentioned (e.g. curly braces in a label value).
   - Tests for this would be nice in this PR too
2. Some more tests for various permutations would be nice, specifically there are not many tests for escaped label names/values right now

tests/openmetrics/test_parser.py
```diff
@@ -187,17 +207,53 @@ def test_native_histogram(self):
...
+       
+        hfm = HistogramMetricFamily("native{histogram", "Is a basic example of a native histogram")
+        hfm.add_sample("native{histogram", {'xx{} # {}': ' EOF # {}}}'}, None, None, None, NativeHistogram(24, 100, 0, 0.001, 4, (BucketSpan(0, 2), BucketSpan(1, 2)), (BucketSpan(0, 2), BucketSpan(1, 2)), (2, 1, -3, 3), (2, 1, -2, 3)))
+        self.assertEqual([hfm], families)
 
     def test_native_histogram_with_labels(self):
         families = text_string_to_metric_families("""# TYPE hist_w_labels histogram
 # HELP hist_w_labels Is a basic example of a native histogram with labels
-hist_w_labels{foo="bar",baz="qux"} {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-3,3],negative_deltas:[2,1,-2,3]}
+hist_w_labels{foo="bar",baz="qux"} {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-3,3],negative_deltas:[2,1,-2,2]}
```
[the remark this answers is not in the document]
ywwg: oh hah, I changed this to make the test fail so I could see how it worked

prometheus_client/validation.py
ywwg: the functions are used by both prometheus_client and prometheus_client/openmetrics so I thought they had to be public

csmarchbanks: Ahh, yeah Python doesn't really do public/private, it's just a convention.

ywwg: I am confused what change you want me to make, then -- leave it?  prepend _? something else?

ywwg:
Added a test, fixed an issue using  .split

prometheus_client/parser.py
```diff
@@ -45,54 +49,161 @@ def _is_character_escaped(s: str, charpos: int) -> bool:
...
+        if maxsplit > 0 and len(tokens) > maxsplit:
+            tokens[-1] = text[x:]
+            break
+        tokens[-1] = text[x:split_pos]
+        x = split_pos + 1
+        tokens.append('')
+    return tokens
+
+
+def unquote_unescape(text):
```
ywwg, line 193: maybe these should be private too?

prometheus_client/openmetrics/exposition.py
```diff
@@ -59,16 +69,40 @@ def generate_latest(registry):
...
+    return '"{}"'.format(escape(s))
+
+
+def escape_label_name(s: str) -> str:
+    if _is_valid_legacy_labelname(s):
+        return s
+    return '"{}"'.format(escape(s))
+
+
+def escape(s: str) -> str:
```
csmarchbanks, line 107: This one as private could be nice s well. I like the other escapes for metric and label names being public though as part of this work.

prometheus_client/parser.py
csmarchbanks: I would say yes, thanks!

prometheus_client/parser.py
```diff
@@ -104,34 +215,50 @@ def _parse_value_and_timestamp(s: str) -> Tuple[float, Optional[float]]:
...
-    value = float(values[0])
-    timestamp = (float(values[-1]) / 1000) if len(values) > 1 else None
+    value = parse_value(values[0])
+    timestamp = (parse_value(values[-1]) / 1000) if len(values) > 1 else None
     return value, timestamp
 
 
-def _parse_sample(text: str) -> Sample:
-    # Detect the labels in the text
+def parse_value(value):
```
csmarchbanks, line 223:
```suggestion
def _parse_value(value):
```
here as well to match `_parse_sample`

csmarchbanks (commented):
Just a bit of prepending with more `_` and I think this is good to go! Plus cleaning up the commit history and such.

prometheus_client/openmetrics/exposition.py
```diff
@@ -59,16 +71,47 @@ def generate_latest(registry):
...
         except Exception as exception:
             exception.args = (exception.args or ('',)) + (metric,)
             raise
 
     output.append('# EOF\n')
     return ''.join(output).encode('utf-8')
+
+
+def escape_metric_name(s: str) -> str:
+    """Escapes the metric name and puts it in quotes iff the name does not
```
vesari, line 98:
```suggestion
    """Escapes the metric name and puts it in quotes if the name does not
```

prometheus_client/openmetrics/exposition.py
```diff
@@ -59,16 +71,47 @@ def generate_latest(registry):
...
+    conform to the legacy Prometheus character set.
+    """
+    if _is_valid_legacy_metric_name(s):
+        return s
+    return '"{}"'.format(_escape(s))
+
+
+def escape_label_name(s: str) -> str:
+    """Escapes the label name and puts it in quotes iff the name does not
```
vesari, line 107:
```suggestion
    """Escapes the label name and puts it in quotes if the name does not
```

vesari (approved):
Thanks for this great work, LGTM!

prometheus_client/openmetrics/exposition.py
ywwg: This is shorthand for "if and only if" which is perhaps obscure, but adds detail to the explanation.  I can write out the full phrase if that's better

vesari: Oops, my bad, please leave them as they are :)

csmarchbanks (approved):

PR #1074, opened 2024-11-22
Use specific exception for duplicate timeseries

kajinamit:
Use sub-class of ValueError instead of ValueError, so that we can distinguish issues caused by wrong input (like invalid name format) from duplicate metrics being registered into the same registry.

### DIFF
--- a/prometheus_client/registry.py
+++ b/prometheus_client/registry.py
@@ -1,6 +1,6 @@
 import copy
 from threading import Lock
-from typing import Dict, Iterable, List, Optional, Protocol
+from typing import Dict, Iterable, List, Optional, Protocol, Set
 
 from .metrics_core import Metric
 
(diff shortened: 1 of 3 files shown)


prometheus_client/registry.py
```diff
@@ -18,6 +18,14 @@ def collect(self) -> Iterable[Metric]:
         return []
 
 
+class DuplicateTimeseries(ValueError):
+    def __init__(self, duplicates):
+        msg = 'Duplicated timeseries in CollectorRegistry: {}'.format(
+            duplicates)
+        super().__init__(msg)
+        self.duplicates = duplicates
```
csmarchbanks, line 26: Is the idea here that a client of this library can see what the duplicate names are? If so I think some typing would be nice here as it is for external consumption.

csmarchbanks (commented):
Just curious what you would do when you encounter a duplicate name vs a different ValueError? I am not opposed to this if it is helpful but would like to understand as well/not encourage bad practices.

prometheus_client/registry.py
kajinamit: That's a fair point. I've added the typing definition.

kajinamit:
> Just curious what you would do when you encounter a duplicate name vs a different ValueError? I am not opposed to this if it is helpful but would like to understand as well/not encourage bad practices.

We are currently using this client to export metrics of virtual machines running on a specific hypervisor. In this use case  we have to register virtual machines dynamically, because virtual machines might be created/deleted/moved anytime. To achieve the dynamic detection, we check all virtual machines in a host and register metrics for them periodically, but we can't safely check if the virtual machine metrics were registered in the previous cycle without this. (For now we inspect _names_to_collectors before calling register but that's requires external access to a private item, which I want to get rid of)

csmarchbanks (approved):

PR #1077, opened 2024-12-03
Check if labelvalues is in _metrics before deletion in MetricWrapperBase.remove()

GlorifiedPig:
Currently raises an error if labelvalues is not in _metrics. Since the labelvalues get removed anyways, there isn't really any point of this error, since we would want to assume that after running this function call labelvalues will not exist regardless of it existed beforehand or not.


csmarchbanks (approved):
Seems like a reasonable quality of life improvement. Thanks!

PR #1079, opened 2024-12-03
Add deadlock detection

suligap:
Detect deadlocks during the library misuse, eg. by injecting code into the critical sections that itself might want to obtain the relevant lock.

A follow up to #1076.


csmarchbanks:
Just want to say I have been busy this week and am out next week, but I will get to this, and thank you for your contributions!

suligap:
No worries, nothing urgent here and thanks. I thought about it a bit more and ended up pulling out the deadlock prevention from the "hottest" path: metrics.py and values.py. It's quite hard to justify the overhead of this double locking approach there. But I think it's not an issue in registry.py.

PR #1082, opened 2025-01-08
Correct nh sample span structure and parsing

vesari:
While working on the exposition part of the issue https://github.com/prometheus/OpenMetrics/issues/279, I realized I made a mistake in the code relative to the NH spans and deltas parsing, in that I was not taking into account that span lists have no fixed length, span lists can be absent and that deltas can consequently also not be there. I thought I’d get this right before continuing on the rest of the work.


prometheus_client/openmetrics/parser.py
```diff
@@ -364,6 +342,32 @@ def _parse_nh_struct(text):
         pos_deltas=pos_deltas,
         neg_deltas=neg_deltas
     )
+  
+
+def _compose_spans(spans, spans_name):
+    try:
```
krajorama, line 348: nit: i'd prefer early return instead of exception handling
```
if spans_name not in spans:
  return None
```

prometheus_client/openmetrics/parser.py
```diff
@@ -364,6 +342,32 @@ def _parse_nh_struct(text):
...
+        pos_spans_text = spans[spans_name]
+        pos_spans = []
+        for start, end in pos_spans_text:
```
krajorama, line 351: That _text is Hungarian notation :)
```suggestion
        for start, end in spans[spans_name]:
```

prometheus_client/openmetrics/parser.py
```diff
@@ -307,12 +307,17 @@ def _parse_nh_sample(text, suffixes):
 
 def _parse_nh_struct(text):
     pattern = r'(\w+):\s*([^,}]+)'
-    
-    re_spans = re.compile(r'(positive_spans|negative_spans):\[(\d+:\d+,\d+:\d+)\]')
+    re_spans = re.compile(r'(positive_spans|negative_spans):\[(\d+:\d+(,\d+:\d+)*)\]')
     re_deltas = re.compile(r'(positive_deltas|negative_deltas):\[(-?\d+(?:,-?\d+)*)\]')
```
krajorama, line 311: That middle question mark looks off to me

```suggestion
    re_deltas = re.compile(r'(positive_deltas|negative_deltas):\[(-?\d+(:,-?\d+)*)\]')
```

prometheus_client/openmetrics/parser.py
```diff
@@ -307,12 +307,17 @@ def _parse_nh_sample(text, suffixes):
...
 
     items = dict(re.findall(pattern, text))
-    spans = dict(re_spans.findall(text))
+    matches = re_spans.findall(text)
+    spans = {}
```
krajorama, line 315: Adding this here makes the code a little inconsistent, you do the text->data conversion here for spans, but in a helper function for the deltas. Let's move this into the helper function.

prometheus_client/openmetrics/parser.py
vesari: I’m afraid it cannot be omitted. My understanding is that this `(?:,-?\d+)*` matches a pattern like this `1,2,-3`, whereas `(:,-?\d+)*` would end up matching something like this `1:,2:,-3:`. In any case, if I remove the question mark there, the related tests fail :/

krajorama: Ah, I didn't know that `(?:<exp>)` was special syntax. https://docs.python.org/3/library/re.html#regular-expression-syntax . Ack.

prometheus_client/openmetrics/parser.py
```diff
@@ -364,6 +336,31 @@ def _parse_nh_struct(text):
         pos_deltas=pos_deltas,
         neg_deltas=neg_deltas
     )
+  
+
+def _compose_spans(span_matches, spans_name):
```
krajorama, line 341: nit: might want to add a comment what this does, because the list comprehension takes a minute to understand :)

krajorama (approved):
LGTM

csmarchbanks (approved):
Also :+1: though I will give you a moment if you want to add a comment, I agree it would be nice.

prometheus_client/openmetrics/parser.py
vesari: Yes, that makes sense! I've also added comments to the other less complicated function, while I was at it :D

csmarchbanks (approved):

PR #1084, opened 2025-01-13
Migrate from setup.py to pyproject.toml

csmarchbanks:
Update the project configuration to use a pyproject.toml file instead of setup.py. This is the preferred tool and will allow easier integration with other tools in the future. We can also get rid of MANIFEST.in as the cache and compiled files are automatically excluded.


pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+
+[project]
+name = "prometheus_client"
+version = "0.22.0"
+description = "Python client for the Prometheus monitoring system."
+readme = "README.md"
+license = { file = "LICENSE" }
+requires-python = ">=3.9"
+authors = [
+    { name = "The Prometheus Authors", email = "prometheus-developers@googlegroups.com" },
```
csmarchbanks, line 13: This should be the only substantive change, moving from a specific author to the Promtheus developers as a whole.

pyproject.toml
```diff
@@ -0,0 +1,49 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "prometheus_client"
+version = "0.21.1"
```
csmarchbanks, line 7: This also changed to match the most recent release which had a backport applied to it.

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
sd2k, line 35: Very minor but it bugs me that these are no longer sorted :sweat_smile:

sd2k (commented):
Looks fine to me (although I've never used the pyproject / setuptools build system). One very minor comment!

pyproject.toml
csmarchbanks: Ah good call. I had a tool that did the initial transform, no idea why it chose that weird sort (or lack thereof?). Fixed.

PR #1087, opened 2025-01-26
OM text exposition for NH

vesari:
For the sake of separation of concerns, this PR focuses only on the OM exposition format for native histograms and it does so by mocking native histograms in tests i.e. it uses the helper function `custom_collector` (as it is already done for gauge histograms) instead of the `observe` function. I will carry out the implementation of the "complete" support for native histograms in client_python in another iteration (I'm already on it). 
Incidentally, in the exposition logic, the `#HELP` line gets appended before the `#TYPE` line, whereas in the parser logic is the other way around; also, in the exposition logic the value is always a float in the OM format, whereas in the parser it doesn't seem to be the norm. In this PR I follow the behaviour already existing in the exposition logic. Let me know if I should change anything in this regard.

### DIFF
--- a/prometheus_client/openmetrics/exposition.py
+++ b/prometheus_client/openmetrics/exposition.py
@@ -26,6 +26,7 @@ def generate_latest(registry):
     for metric in registry.collect():
         try:
             mname = metric.name
+            # (Vesari): TODO: this is wrong. TYPE should come before HELP!!!!
             output.append('# HELP {} {}\n'.format(
                 escape_metric_name(mname), _escape(metric.documentation)))
             output.append(f'# TYPE {escape_metric_name(mname)} {metric.type}\n')
@@ -67,24 +68,72 @@ def generate_latest(registry):
                             floatToGoString(s.exemplar.value),
                         )
                 else:
-                    exemplarstr = ''
+                    exemplarstr = ''   
+                
                 timestamp = ''
                 if s.timestamp is not None:
                     timestamp = f' {s.timestamp}'
+
+                native_histogram = ''
+                positive_spans = ''
+                positive_deltas = ''
+                negative_spans = ''
+                negative_deltas = ''      
+                if s.native_histogram:
+                    if s.name is not metric.name:
+                        raise ValueError(f"Metric {metric.name} is native histogram, but sample name is not valid")         
+                    if s.native_histogram.pos_spans:
+                        positive_spans = ','.join([f'{ps[0]}:{ps[1]}' for ps in s.native_histogram.pos_spans])
+                        positive_deltas = ','.join(str(pd) for pd in s.native_histogram.pos_deltas)
+                    if s.native_histogram.neg_spans:        
+                        negative_spans = ','.join([f'{ns[0]}:{ns[1]}' for ns in s.native_histogram.neg_spans])                  
+                        negative_deltas = ','.join(str(nd) for nd in s.native_histogram.neg_deltas)
+                    
+                    nh_sample_template = '{{count:{},sum:{},schema:{},zero_threshold:{},zero_count:{}}}'
+                    if positive_spans != '':
+                        nh_sample_template += ',positive_spans:[{{{}}}]'
+                    if negative_spans != '':
+                       nh_sample_template += ',negative_spans:[{{{}}}]'
+                    if positive_deltas != '':
+                       nh_sample_template += ',positive_deltas:[{{{}}}]'
+                       if negative_deltas == '':
+                           nh_sample_template += '}}'
+                    if negative_deltas != '':
+                       nh_sample_template += ',negative_deltas:[{{{}}}]'
+                       nh_sample_template += '}}'
+                      
+                    native_histogram = nh_sample_template.format(
+                        s.native_histogram.count_value,
+                        s.native_histogram.sum_value,
+                        s.native_histogram.schema,
+                        s.native_histogram.zero_threshold,
+                        s.native_histogram.zero_count,
+                        positive_spans,
+                        negative_spans,
+                        positive_deltas,
+                        negative_deltas,
+                    )   
+                   
+                value = ''    
+                if s.value is not None or not s.native_histogram:
+                    value = floatToGoString(s.value)       
                 if _is_valid_legacy_metric_name(s.name):
-                    output.append('{}{} {}{}{}\n'.format(
+                    output.append('{}{} {}{}{}{}\n'.format(
                         s.name,
                         labelstr,
-                        floatToGoString(s.value),
+                        value,
                         timestamp,
                         exemplarstr,
+                        native_histogram
                     ))
+
                 else:
-                    output.append('{} {}{}{}\n'.format(
+                    output.append('{} {}{}{}{}\n'.format(
                         labelstr,
-                        floatToGoString(s.value),
+                        value,
                         timestamp,
                         exemplarstr,
+                        native_histogram
                     ))
         except Exception as exception:
             exception.args = (exception.args or ('',)) + (metric,)
--- a/prometheus_client/openmetrics/exposition.py
+++ b/prometheus_client/openmetrics/exposition.py
@@ -26,61 +26,22 @@ def generate_latest(registry):
     for metric in registry.collect():
         try:
             mname = metric.name
-            # (Vesari): TODO: this is wrong. TYPE should come before HELP!!!!
             output.append('# HELP {} {}\n'.format(
                 escape_metric_name(mname), _escape(metric.documentation)))
             output.append(f'# TYPE {escape_metric_name(mname)} {metric.type}\n')
             if metric.unit:
-                output.append(f'# UNIT {escape_metric_name(mname)} {metric.unit}\n')
+                output.append(f'# UNIT {escape_metric_name(mname)} {metric.unit}\n')    
             for s in metric.samples:
-                if not _is_valid_legacy_metric_name(s.name):
-                    labelstr = escape_metric_name(s.name)
-                    if s.labels:
-                        labelstr += ', '
-                else:
-                    labelstr = ''
+                labelstr, exemplarstr, timestamp = _expose_classic_metrics_sample(metric, s)
                 
-                if s.labels:
-                    items = s.labels.items()
-                    labelstr += ','.join(
-                        ['{}="{}"'.format(
-                            escape_label_name(k), _escape(v))
-                            for k, v in items])
-                if labelstr:
-                    labelstr = "{" + labelstr + "}"
-                    
-                if s.exemplar:
-                    if not _is_valid_exemplar_metric(metric, s):
-                        raise ValueError(f"Metric {metric.name} has exemplars, but is not a histogram bucket or counter")
-                    labels = '{{{0}}}'.format(','.join(
-                        ['{}="{}"'.format(
-                            k, v.replace('\\', r'\\').replace('\n', r'\n').replace('"', r'\"'))
-                            for k, v in sorted(s.exemplar.labels.items())]))
-                    if s.exemplar.timestamp is not None:
-                        exemplarstr = ' # {} {} {}'.format(
-                            labels,
-                            floatToGoString(s.exemplar.value),
-                            s.exemplar.timestamp,
-                        )
-                    else:
-                        exemplarstr = ' # {} {}'.format(
-                            labels,
-                            floatToGoString(s.exemplar.value),
-                        )
-                else:
-                    exemplarstr = ''   
-                
-                timestamp = ''
-                if s.timestamp is not None:
-                    timestamp = f' {s.timestamp}'
-
                 native_histogram = ''
                 positive_spans = ''
                 positive_deltas = ''
                 negative_spans = ''
                 negative_deltas = ''
                 pos = False
-                neg = False      
+                neg = False
+                     
                 if s.native_histogram:                
                     # Initialize basic nh template
                     nh_sample_template = '{{count:{},sum:{},schema:{},zero_threshold:{},zero_count:{}'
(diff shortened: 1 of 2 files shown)


prometheus_client/openmetrics/exposition.py
```diff
@@ -67,24 +67,87 @@ def generate_latest(registry):
...
                 if _is_valid_legacy_metric_name(s.name):
-                    output.append('{}{} {}{}{}\n'.format(
+                    output.append('{}{} {}{}{}{}\n'.format(
                         s.name,
                         labelstr,
-                        floatToGoString(s.value),
+                        value,
                         timestamp,
                         exemplarstr,
+                        native_histogram
```
krajorama, line 142: this must be placed immediately after `value,` because the timestamp and exemplars come last

prometheus_client/openmetrics/exposition.py
```diff
@@ -67,24 +67,87 @@ def generate_latest(registry):
...
                     ))
                 else:
-                    output.append('{} {}{}{}\n'.format(
+                    output.append('{} {}{}{}{}\n'.format(
                         labelstr,
-                        floatToGoString(s.value),
+                        value,
                         timestamp,
                         exemplarstr,
+                        native_histogram
```
krajorama, line 150: this must be placed immediately after `value,` because the timestamp and exemplars come last

tests/openmetrics/test_exposition.py
```diff
@@ -94,6 +95,125 @@ def test_histogram(self):
 # EOF
 """, generate_latest(self.registry))
 
+
```
krajorama, line 98: please also add test for exemplars (note that native histograms may have multiple exemplars) see example in https://github.com/prometheus/proposals/blob/main/proposals/2024-01-29_native_histograms_text_format.md

prometheus_client/openmetrics/exposition.py
```diff
@@ -67,24 +67,87 @@ def generate_latest(registry):
...
+                        s.native_histogram.count_value,
+                        s.native_histogram.sum_value,
+                        s.native_histogram.schema,
+                        s.native_histogram.zero_threshold,
+                        s.native_histogram.zero_count,
+                    ]
+
+                    # Signal presence for pos/neg spans/deltas
+                    pos = False
+                    neg = False
```
krajorama, line 98: nit: these just shadow `s.native_histogram.pos_spans` and `s.native_histogram.neg_spans` so I'd just write that, but no strong opinion

prometheus_client/openmetrics/exposition.py
```diff
@@ -67,24 +67,87 @@ def generate_latest(registry):
...
+                        args.append(negative_deltas)
+
+                    # Add closing brace
+                    nh_sample_template += '}}'
+
+                    # Format the template with the args
+                    native_histogram = nh_sample_template.format(*args)    
+
+                value = ''    
+                if s.value is not None or not s.native_histogram:
```
krajorama, line 133: it should be enough to write
```suggestion
                if not s.native_histogram:
```

At least that's what we do in Prometheus. Type checks should prevent having None in the value.

tests/openmetrics/test_exposition.py
```diff
@@ -94,6 +95,125 @@ def test_histogram(self):
...
+    def test_native_histogram(self):
+        hfm = HistogramMetricFamily("nh", "nh")
+        hfm.add_sample("nh", None, None, None, None, NativeHistogram(24, 100, 0, 0.001, 4, (BucketSpan(0, 2), BucketSpan(1, 2)), (BucketSpan(0, 2), BucketSpan(1, 2)), (2, 1, -3, 3), (2, 1, -2, 3)))
```
krajorama, line 101: The type annotation of add_sample says `labels: Dict[str, str]`, which means that `None` is not allowed. I think adding `{}` should work. Same comment on the rest of the tests.

The type annotation of add_sample says ` value: float` , which means that `None` is not allowed. Also valid comment for the other tests here.

```suggestion
        hfm.add_sample("nh", None, 0, None, None, NativeHistogram(24, 100, 0, 0.001, 4, (BucketSpan(0, 2), BucketSpan(1, 2)), (BucketSpan(0, 2), BucketSpan(1, 2)), (2, 1, -3, 3), (2, 1, -2, 3)))
```

The reason why `mypy` CI check doesn't catch these is due to not having type annotation on the `def` line.

tests/openmetrics/test_exposition.py
```diff
@@ -94,6 +95,125 @@ def test_histogram(self):
 # EOF
 """, generate_latest(self.registry))
 
+
+    def test_native_histogram(self):
```
krajorama, line 99:
```suggestion
    def test_native_histogram(self) -> None:
```
To get type checks. Also for all the other tests here.

krajorama (changes requested):
nice work, I have some suggestions to improve type checking and also the extra spaces need removal

prometheus_client/openmetrics/exposition.py
```diff
@@ -67,24 +67,87 @@ def generate_latest(registry):
...
+
+                    # Append pos deltas if pos spans were added
+                    if pos:
+                        nh_sample_template += ',positive_deltas:[{}]'
+                        args.append(positive_deltas)
+
+                    # Append neg deltas if neg spans were added
+                    if neg:
+                        nh_sample_template += ',negative_deltas:[{}]'
+                        args.append(negative_deltas)
```
csmarchbanks, line 124: Couldn't these just be included in the already present if statements above and then there would be no need for the `pos` and `neg` variables?

prometheus_client/openmetrics/exposition.py
csmarchbanks: I'd actually say we should just store native histogram into `value` and keep the format string intact. There is only one value field allowed in a line so having two fields feels wrong.

prometheus_client/openmetrics/exposition.py
```diff
@@ -115,3 +178,6 @@ def escape_label_name(s: str) -> str:
 def _escape(s: str) -> str:
     """Performs backslash escaping on backslash, newline, and double-quote characters."""
     return s.replace('\\', r'\\').replace('\n', r'\n').replace('"', r'\"')
+
+
+
```
csmarchbanks, line 183:
```suggestion
```

csmarchbanks (commented):
Just a couple more comments to add to krajorama's.

prometheus_client/openmetrics/exposition.py
```diff
@@ -30,24 +30,24 @@ def generate_latest(registry):
...
                     labelstr = escape_metric_name(s.name)
                     if s.labels:
                         labelstr += ', '
                 else:
                     labelstr = ''
-                
+                        
                 if s.labels:
-                    items = sorted(s.labels.items())
+                    items = s.labels.items()
```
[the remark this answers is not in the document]
vesari: I just wanted to use the exact same strings I used in the parser tests (there the labels are not sorted). Since I knew that ordering is not required, I removed the sorting, and once I saw that the pre-existing exposition tests don’t fail, I simply thought I’d kill two birds with one stone, i.e. keeping the same test strings and removing the sorting performance overhead (which I admit I haven’t quantified btw). Having said that, I realize this change is out of the scope of this PR, so of course I will revert it, if you wish.

prometheus_client/openmetrics/exposition.py
vesari: Storing native histograms in `value` was my first approach in the initial PR for the parser. That approach was discarded because, for `value` to support both `float` and `nativeHistogram` types, changes would have been necessary also in `registry.py` and in `bridge/graphite.py`. Those changes were deemed undesirable, thus the decision of adding `native_histogram` as a separate attribute of the `Sample` class. It was added for last (after `exemplars` and `timestamp`) as I thought that would have made it less likely to break things downstream rather than inserting it before. Please find the original conversation here: https://github.com/prometheus/client_python/pull/1040# . While it’s my bad for not having noticed that supporting exemplars had turned from a non-goal to a goal by the time I had already started working on the parser, I knew I should have dealt with that at some point, but I thought to do something along the lines of client_golang, where the exemplars for native histograms (if I understand the code correctly) are represented separately from the “classic” exemplars https://github.com/prometheus/client_golang/blob/e729ba11961abb3a50910c324221d06a35bdce4f/prometheus/histogram.go#L1678 by means of a `nativeExemplar` type.

I’ll be happy to open a new PR on the parser first in order to support exemplars and address your change requests, but I can’t do it without reaching a consensus on how class `Sample` should look like.  The way I see it, there are at least 3 possibilities:
1. Class `Sample` stays as it is, while I add an attribute (say, `native_exemplars` of some `NativeExemplar` type) to class `NativeHistogram`.

2. I revert class `Sample` to have the `value` attribute accepting both `float` and `NativeHistogram` types, (but then I wouldn’t know how to avoid changing `registry.py` and `bridge/graphite.py`).

 3. I move up the `native_histogram` attribute, so that class `Sample` becomes:

```
class Sample(NamedTuple):
    name: str
    labels: Dict[str, str]
    value: float
    native_histogram: Optional[NativeHistogram] = None
    timestamp: Optional[Union[float, Timestamp]] = None
    exemplar: Optional[Exemplar] = None
```
   
and change things accordingly elsewhere. 

Which one shall I opt for?

vesari:
@csmarchbanks and @krajorama , thanks a lot for your reviews! Before I proceed addressing your change requests, let me know how you think class `Sample` should look like, as per my comment above. Many thanks in advance!

prometheus_client/openmetrics/exposition.py
krajorama: In Go we opted for separate sample types for floats , and native histograms (actually there's two native histogram sample types in Go, but I digress) due to performance issues negatively affecting the float sample values.

I don't know how much of a concern it is in Python, I suspect it's not a big issue.

So my vote is:
let's finish this PR with option 1. My comment about the order of __printing__ is still valid.

Let's try to implement option 2 in a separate PR . I wish we could unify the types in Go as well...

I don't understand option 3, my comments are about the printing order , not the order in the tuple which I don't care about.

csmarchbanks: I wasn't thinking about changing types elsewhere, just in this small bit of code when doing the formatting. E.g. just above this line do:
```python
value = ''    
if s.native_histogram:
    value = native_histogram
elif s.value is not None
    value = floatToGoString(s.value)
```

prometheus_client/openmetrics/exposition.py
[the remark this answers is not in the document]
csmarchbanks: Let's keep the sorting. I'd have to go and make sure that it doesn't cause an issue anywhere not tested. Probably ok, but for example the sorting is important for multiprocess as that is how we match keys.

prometheus_client/openmetrics/exposition.py
vesari: In order to do that, assuming we don't want to print out the keys of unpopulated fields, I would have to compose all the partial strings before knowing if they'll have to be added or not and then write the necessary conditionals along the lines of:

If pos spans and not neg spans:
   add pos spans and pos deltas

If pos spans and neg spans:
   add pos spans neg spans pos deltas and neg deltas

If neg spans and not pos spans:
   add neg spans and neg deltas

Maybe there are nicer solutions, but maybe now that I'm not shadowing `s.native_histogram.pos_spans` and `s.native_histogram.neg_spans` anymore, it looks more acceptable than before? XD

vesari:
> I have started the OpenMetrics 2.0 specification. Might be worth aligning on that first: [prometheus/docs#2634](https://github.com/prometheus/docs/pull/2634) .
> 
> Some things that came up due to this work are listed in a doc pointed out by the PR: https://docs.google.com/document/d/1FCD-38Xz1-9b3ExgHOeDTQUKUatzgj5KbCND9t-abZY/edit?tab=t.nvipu0hrna3u


Thank you for the heads-up! I'll read the PR + doc and align the code with the new decisions as soon as possible.

vesari:
Following yesterday's OM 2.0 working group meeting, I guess I can continue implementing changes gradually in this PR every time consensus gets reached on the open questions regarding native histograms (no matter how long it will take for the group to decide). At the same time I could also work to get the parser "up to speed" in a similar way (I think the case when there are nh with exemplars to be parsed isn't yet contemplated, for example), in a separate, new PR. Do you have any objections or other plans @csmarchbanks @krajorama ? Thanks in advance!

vesari:
P.S. What I mean is that I'm fine with this or any similar PR of mine to stay open for as long as it takes :)

vesari:
@csmarchbanks conflicts solved :)

prometheus_client/openmetrics/exposition.py
```diff
@@ -58,44 +81,80 @@ def generate_latest(registry, escaping=UNDERSCORES):
...
+                        s.native_histogram.schema,
+                        s.native_histogram.zero_threshold,
+                        s.native_histogram.zero_count,
+                    ]
+                  
+                    # If there are neg spans, append them and the neg deltas to the template and args
+                    if s.native_histogram.neg_spans:
+                        negative_spans = ','.join([f'{ns[0]}:{ns[1]}' for ns in s.native_histogram.neg_spans])
+                        negative_deltas = ','.join(str(nd) for nd in s.native_histogram.neg_deltas)
+                        nh_sample_template += ',n_spans:[{}]'
```
krajorama, line 114: Spans and buckets are prefixed with `positive` and `negative` in the proposal, not `n`, `p`. Same in OM2.0 proposal. https://github.com/prometheus/proposals/blob/main/proposals/0032-native_histograms_text_format.md#how

csmarchbanks (approved):
Thanks for all the updates! Approving this but I'll give until Monday for any final comments before merging!

vesari:
Following up on the Slack conversation in the prometheus-histogram-dev channel, @krajo did you want for me to change the timestamp implementation and to emit native histograms only when OM 2.0 is requested, in this PR or in another iteration?

PR #1090, opened 2025-02-13
docs: update multiprocess collector code example

moret:
**Why is this PR necessary, what does it do?**
As far as I could test locally @csmarchbanks, the documented example does not work in practice, and different web workers will show different sets of metrics. This PR will update the documentation to the manner which seems to have worked for serving the collected data from all processes.

**References:**
https://prometheus.github.io/client_python/multiprocess/

**Reproduction and example:**
https://github.com/moret/prometheus-multiprocess-example

### DIFF
--- a/docs/content/multiprocess/_index.md
+++ b/docs/content/multiprocess/_index.md
@@ -49,8 +49,7 @@ MY_COUNTER = Counter('my_counter', 'Description of my counter')
 # Expose metrics.
 def app(environ, start_response):
     registry = CollectorRegistry()
-    multiprocess.MultiProcessCollector(registry)
-    data = generate_latest(registry)
+    data = generate_latest(multiprocess.MultiProcessCollector(registry))
     status = '200 OK'
     response_headers = [
         ('Content-type', CONTENT_TYPE_LATEST),


csmarchbanks (commented):
I just want to say that I am looking at this. I am a bit confused as well, I use basically this exact code in an app that does work...

csmarchbanks:
Ahh, I see it, you can't do https://github.com/moret/prometheus-multiprocess-example/blob/master/main.py#L18 in multiprocess mode, we call it out in the docs: "Registering metrics to a registry later used by a MultiProcessCollector may cause duplicate metrics to be exported" in https://prometheus.github.io/client_python/multiprocess/, but maybe we should also add at least a warning?

moret:
Hi @csmarchbanks . About:

> "Registering metrics to a registry later used by a MultiProcessCollector may cause duplicate metrics to be exported"

I had a hard time getting what you meant with it, but I think I got it now. Indeed [moving the metrics earlier](https://github.com/moret/prometheus-multiprocess-example/commit/fec25956be947a0247fa829f128721a267728377) solves it. I think what I was struggling with was that I assumed if we weren't passing the same registry we would use to `generate_latest` to the metrics they would miss out on each other.

Perhaps a warning or avoiding the duplication would indeed be fine, but clearly the documentation is correct, I was over-doing it. 😄 

Thank you a lot for your attention, I'll close this PR, let me know if I can help somehow reviewing any attempt to go for a warning or avoiding the duplication!

PR #1093, opened 2025-02-24
Fix order-dependent flaky tests related to UTF-8 support

dg98:
For a university course, I tested this project for flaky tests, which are tests that can both pass or fail without changes to their code. I noticed that the tests changed in #1070 fail when their execution order is changed because the legacy validation is enabled in some tests but disabled in the test that follows. Since the execution order is not ensured, it's safer to disable them in the same test.
With these changes applied, the whole test suit passes every time even when executed in random order.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -513,9 +513,9 @@ class TestHistogram(unittest.TestCase):
         enable_legacy_validation()
         self.assertRaises(ValueError, self.histogram.observe, 3.0, exemplar={':o)': 'smile'})
         self.assertRaises(ValueError, self.histogram.observe, 3.0, exemplar={'1': 'number'})
+        disable_legacy_validation()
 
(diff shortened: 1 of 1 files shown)


csmarchbanks (commented):
Thanks!

The downside to this approach is that if one of the assertions raise then the next test will also fail since `disable_legacy_validation()` will no longer be called. A couple options would be to:
1. Wrap the validation enablement/disabling with a [With Statement Context Manager](https://docs.python.org/3/reference/datamodel.html#context-managers)
2. Create a test fixture to be used when we need to use legacy validation.

dg98:
Thanks for your feedback, I didn't think of that. I added a context manager that enables and disables legacy validation and changed all uses of these functions to with statements

csmarchbanks (approved):
Looks great, thanks!

PR #1098, opened 2025-03-19
Add note on gauge.set_function not working with multiprocessing

aapeliv:
https://github.com/prometheus/client_python/issues/504#issuecomment-1358161560

So the next person doesn't have to spend 30 min on this.


csmarchbanks (approved):
Thanks!

PR #1102, opened 2025-04-08
UTF-8 Content Negotiation

ywwg:
Part of https://github.com/prometheus/client_python/issues/1013

The remaining piece is configuration to force escaping even if a scraper does not request it

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -1,5 +1,6 @@
 import base64
 from contextlib import closing
+from functools import partial
 import gzip
 from http.server import BaseHTTPRequestHandler
 import os
@@ -17,13 +18,16 @@ from urllib.request import (
 )
 from wsgiref.simple_server import make_server, WSGIRequestHandler, WSGIServer
 
+from packaging.version import Version
+
 from .openmetrics import exposition as openmetrics
 from .registry import CollectorRegistry, REGISTRY
 from .utils import floatToGoString
-from .validation import _is_valid_legacy_metric_name
 
 __all__ = (
     'CONTENT_TYPE_LATEST',
+    'CONTENT_TYPE_PLAIN_0_0_4',
+    'CONTENT_TYPE_PLAIN_1_0_0',
     'delete_from_gateway',
     'generate_latest',
     'instance_ip_grouping_key',
@@ -37,8 +41,13 @@ __all__ = (
     'write_to_textfile',
 )
 
-CONTENT_TYPE_LATEST = 'text/plain; version=0.0.4; charset=utf-8'
-"""Content type of the latest text format"""
+CONTENT_TYPE_PLAIN_0_0_4 = 'text/plain; version=0.0.4; charset=utf-8'
+"""Content type of the compatibility format"""
+
+CONTENT_TYPE_PLAIN_1_0_0 = 'text/plain; version=1.0.0; charset=utf-8'
+"""Content type of the latest format"""
+
+CONTENT_TYPE_LATEST = CONTENT_TYPE_PLAIN_1_0_0
 
 
 class _PrometheusRedirectHandler(HTTPRedirectHandler):
@@ -245,14 +254,23 @@ def start_wsgi_server(
 start_http_server = start_wsgi_server
 
 
-def generate_latest(registry: CollectorRegistry = REGISTRY) -> bytes:
-    """Returns the metrics from the registry in latest text format as a string."""
+def generate_latest(registry: CollectorRegistry = REGISTRY, escaping: str = openmetrics.UNDERSCORES) -> bytes:
+    """
+    Generates the exposition format using the basic Prometheus text format.
+
+    Params:
+        registry: CollectorRegistry to export data from.
+        escaping: Escaping scheme used for metric and label names.
+
+    Returns: UTF-8 encoded string containing the metrics in text format.
+    """
 
     def sample_line(samples):
         if samples.labels:
             labelstr = '{0}'.format(','.join(
+                # Label values always support UTF-8
                 ['{}="{}"'.format(
-                    openmetrics.escape_label_name(k), openmetrics._escape(v))
+                    openmetrics.escape_label_name(k, escaping), openmetrics._escape(v, openmetrics.ALLOWUTF8, False))
                     for k, v in sorted(samples.labels.items())]))
         else:
             labelstr = ''
@@ -260,14 +278,14 @@ def generate_latest(registry: CollectorRegistry = REGISTRY) -> bytes:
         if samples.timestamp is not None:
             # Convert to milliseconds.
             timestamp = f' {int(float(samples.timestamp) * 1000):d}'
-        if _is_valid_legacy_metric_name(samples.name):
+        if escaping != openmetrics.ALLOWUTF8 or openmetrics._is_valid_legacy_metric_name(samples.name):
             if labelstr:
                 labelstr = '{{{0}}}'.format(labelstr)
-            return f'{samples.name}{labelstr} {floatToGoString(samples.value)}{timestamp}\n'
+            return f'{openmetrics.escape_metric_name(samples.name, escaping)}{labelstr} {floatToGoString(samples.value)}{timestamp}\n'
         maybe_comma = ''
         if labelstr:
             maybe_comma = ','
-        return f'{{{openmetrics.escape_metric_name(samples.name)}{maybe_comma}{labelstr}}} {floatToGoString(samples.value)}{timestamp}\n'
+        return f'{{{openmetrics.escape_metric_name(samples.name, escaping)}{maybe_comma}{labelstr}}} {floatToGoString(samples.value)}{timestamp}\n'
 
     output = []
     for metric in registry.collect():
@@ -290,8 +308,8 @@ def generate_latest(registry: CollectorRegistry = REGISTRY) -> bytes:
                 mtype = 'untyped'
 
             output.append('# HELP {} {}\n'.format(
-                openmetrics.escape_metric_name(mname), metric.documentation.replace('\\', r'\\').replace('\n', r'\n')))
-            output.append(f'# TYPE {openmetrics.escape_metric_name(mname)} {mtype}\n')
+                openmetrics.escape_metric_name(mname, escaping), metric.documentation.replace('\\', r'\\').replace('\n', r'\n')))
+            output.append(f'# TYPE {openmetrics.escape_metric_name(mname, escaping)} {mtype}\n')
 
             om_samples: Dict[str, List[str]] = {}
             for s in metric.samples:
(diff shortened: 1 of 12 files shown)


prometheus_client/exposition.py
```diff
@@ -37,7 +37,7 @@
     'write_to_textfile',
 )
 
-CONTENT_TYPE_LATEST = 'text/plain; version=0.0.4; charset=utf-8'
+CONTENT_TYPE_PLAIN = 'text/plain; version=0.0.4; charset=utf-8'
```
ywwg, line 40: text/plain v 0.0.4 is definitely not the latest, so I thought it best to rename this

prometheus_client/openmetrics/exposition.py
```diff
@@ -94,24 +108,114 @@ def generate_latest(registry):
...
+        return _escape(s, escaping, True)
+    elif escaping == VALUES:
+        if _is_valid_legacy_labelname(s):
+            return s
+        return _escape(s, escaping, True)
+    return s
 
 
-def _escape(s: str) -> str:
+def _escape(s: str, escaping: str, is_labelname: bool) -> str:
```
ywwg, line 157: I don't love how messy this function signature is getting...

prometheus_client/exposition.py
csmarchbanks: Though I agree with you, this will be a breaking change as other programs use `CONTENT_TYPE_LATEST`. Perhaps we should update this to version=1.0.0 instead?

prometheus_client/exposition.py
```diff
@@ -307,21 +316,69 @@ def sample_line(samples):
...
-            return (openmetrics.generate_latest,
-                    openmetrics.CONTENT_TYPE_LATEST)
-    return generate_latest, CONTENT_TYPE_LATEST
-
+            toks = accepted.split(';')
+            version = _get_version(toks)
+            escaping = _get_escaping(toks)
+            # Only return an escaping header if we have a good version and
+            # mimetype.
+            if version == '1.0.0':
```
csmarchbanks, line 338: Should this test that the version is at least 1.0.0? I imagine at least anything in 1.x should be ok?

prometheus_client/openmetrics/exposition.py
csmarchbanks: Yeah, it is internal only so not the end of the world. Would it help at all to have a second function for specifically escaping label names?

tools/simple_client.py
csmarchbanks, line 1: Is this used anywhere? Or just a nice example for testing locally?

ywwg: just for testing locally

ywwg: but is also pretty nice for people to get off the ground

prometheus_client/openmetrics/exposition.py
ywwg: I tried another approach that looks better

ywwg:
> Though I agree with you, this will be a breaking change as other programs use CONTENT_TYPE_LATEST. Perhaps we should update this to version=1.0.0 instead?

we do still need a 0.0.4 constant as the fallback when all else fails.  What I could do is update LATEST to be 1.0.0 and then create the new PLAIN constant for 0.0.4

prometheus_client/exposition.py
ywwg: looks like there's a python lib for that!

ywwg: lmk if I should just test for major version versus any >= version

ywwg: (I think any greater version should be fine)

prometheus_client/exposition.py
csmarchbanks: Since this is needed for compatibility, but LATEST and PLAIN are both really different versions of PLAIN that is confusing for external users to use. I'd propose a couple options:
1. Keep `CONTENT_TYPE_PLAIN` internal, so rename it to `_CONTENT_TYPE_PLAIN`.
2. Support content types for each version, so have `CONTENT_TYPE_PLAIN_0_4_0`, `CONTENT_TYPE_PLAIN_1_0_0`, and then alias `CONTENT_TYPE_LATEST` to `CONTENT_TYPE_PLAIN_1_0_0`.

prometheus_client/exposition.py
```diff
@@ -17,13 +17,14 @@
...
+from packaging.version import Version
+
 from .openmetrics import exposition as openmetrics
 from .registry import CollectorRegistry, REGISTRY
 from .utils import floatToGoString
-from .validation import _is_valid_legacy_metric_name
 
 __all__ = (
-    'CONTENT_TYPE_LATEST',
+    'CONTENT_TYPE_PLAIN',
```
csmarchbanks, line 27: This needs to stay as `CONTENT_TYPE_LATEST` as well. We could add in other versions if we choose to go the version route in the options below.

prometheus_client/exposition.py
```diff
@@ -307,20 +320,68 @@ def sample_line(samples):
...
-    return generate_latest, CONTENT_TYPE_LATEST
+            toks = accepted.split(';')
+            version = _get_version(toks)
+            escaping = _get_escaping(toks)
+            # Only return an escaping header if we have a good version and
+            # mimetype.
+            if Version(version) >= Version('1.0.0'):
+                return (openmetrics.generate_latest_fn(escaping), 
+                        openmetrics.CONTENT_TYPE_LATEST + '; escaping=' + str(escaping))
+    return generate_latest, CONTENT_TYPE_PLAIN
```
csmarchbanks, line 345: Shouldn't we check for the version being >= 1.0.0 for Prometheus as well? Right now we would only be able to use UTF-8 for OpenMetrics?

prometheus_client/openmetrics/exposition.py
```diff
@@ -94,24 +111,122 @@ def generate_latest(registry):
     return ''.join(output).encode('utf-8')
 
 
-def escape_metric_name(s: str) -> str:
+def escape_metric_name(s: str, escaping: str) -> str:
     """Escapes the metric name and puts it in quotes iff the name does not
     conform to the legacy Prometheus character set.
     """
-    if _is_valid_legacy_metric_name(s):
+    print("hi", s)
```
csmarchbanks, line 118: Debugging code here and a couple more times in this function :)

prometheus_client/openmetrics/exposition.py
```diff
@@ -94,24 +111,122 @@ def generate_latest(registry):
...
     """Escapes the label name and puts it in quotes iff the name does not
     conform to the legacy Prometheus character set.
     """
-    if _is_valid_legacy_labelname(s):
+    if len(s) == 0:
         return s
-    return '"{}"'.format(_escape(s))
+    if escaping == ALLOWUTF8:
+        if not _is_valid_legacy_labelname(s):
+            print("adding quotes here")
```
csmarchbanks, line 148: Some more debugging code.

prometheus_client/validation.py
```diff
@@ -51,6 +51,9 @@ def _validate_metric_name(name: str) -> None:
 
 def _is_valid_legacy_metric_name(name: str) -> bool:
     """Returns true if the provided metric name conforms to the legacy validation scheme."""
+    if len(name) == 0:
+        return False
+    print("sooooo ", METRIC_NAME_RE.match(name))
```
csmarchbanks, line 56:
```suggestion
```

csmarchbanks (commented):
Generally looks to be getting there! A few comments mostly around the content types still. CONTENT_TYPE_LATEST from just exposition.py (not open metrics) doesn't actually seem to be used?

prometheus_client/exposition.py
ywwg: I like option 2

prometheus_client/openmetrics/exposition.py
ywwg: haha oops

prometheus_client/exposition.py
```diff
@@ -307,20 +324,77 @@ def sample_line(samples):
...
         if accepted.split(';')[0].strip() == 'application/openmetrics-text':
-            return (openmetrics.generate_latest,
-                    openmetrics.CONTENT_TYPE_LATEST)
-    return generate_latest, CONTENT_TYPE_LATEST
+            toks = accepted.split(';')
+            version = _get_version(toks)
+            escaping = _get_escaping(toks)
+            # Only return an escaping header if we have a good version and
+            # mimetype.
+            if Version(version) >= Version('1.0.0'):
```
csmarchbanks, line 346: If no `version` is specified, e.g. `Accept: application/openmetrics-text`, this is throwing an `packaging.version.InvalidVersion` due to version being an empty string.

We should have a tests for sure, but also what version would we want to use in this case? I guess LATEST for each should be fine?

prometheus_client/exposition.py
```diff
@@ -307,20 +324,77 @@ def sample_line(samples):
...
+                return (openmetrics.generate_latest_fn(escaping), 
+                        openmetrics.CONTENT_TYPE_LATEST + '; escaping=' + str(escaping))
+        elif accepted.split(';')[0].strip() == 'text/plain':
+            toks = accepted.split(';')
+            version = _get_version(toks)
+            escaping = _get_escaping(toks)
+            # Only return an escaping header if we have a good version and
+            # mimetype.
+            if Version(version) >= Version('1.0.0'):
+                return (openmetrics.generate_latest_fn(escaping),
```
csmarchbanks, line 356: This returns an openmetrics payload wrapped inside of `text/plain` which is not correct, it needs to be calling `generate_latest` in this module instead. I'd suggest using `partial(generate_latest, escaping=escaping)` from `functools`.

prometheus_client/openmetrics/exposition.py
```diff
@@ -20,34 +30,40 @@ def _is_valid_exemplar_metric(metric, sample):
     return False
 
 
-def generate_latest(registry):
+def generate_latest_fn(escaping):
```
csmarchbanks, line 33: I think this can be removed an replaced with `functools.partial` where called. If not, I think at least prefixing with `_` would be nice.

prometheus_client/openmetrics/exposition.py
```diff
@@ -20,34 +30,40 @@ def _is_valid_exemplar_metric(metric, sample):
...
+    '''Returns a generate_latest function that will always use the given escaping.'''
+    return lambda registry: generate_latest(registry, escaping)
+
+
+def generate_latest(registry, escaping):
```
csmarchbanks, line 38: `escaping` needs a default value for compatibility. Probably underscores?

prometheus_client/openmetrics/exposition.py
```diff
@@ -94,24 +110,118 @@ def generate_latest(registry):
     return ''.join(output).encode('utf-8')
 
 
-def escape_metric_name(s: str) -> str:
+def escape_metric_name(s: str, escaping: str) -> str:
```
csmarchbanks, line 113: Do we want defaults on this and `escape_label_name` as well? Technically this would break users if anyone else uses these functions, but they are pretty new so not as necessary compared to generate_latest.

prometheus_client/exposition.py
ywwg: I believe the correct behavior is to always fall back to 0.0.4 rather than use the latest (https://github.com/prometheus/docs/blob/main/docs/instrumenting/content_negotiation.md#selection-of-format)

Python doesn't seem to support  open metrics 0.0.4 so I believe we have to fall back to prometheus plain there too.

ywwg: Added some test cases

prometheus_client/openmetrics/exposition.py
ywwg: didn't know about that one!

prometheus_client/openmetrics/exposition.py
csmarchbanks: I don't see any public usage of it on Github so I'd say it is ok. We're still a 0.x library anyway.

prometheus_client/exposition.py
csmarchbanks: You could probably use mocks to avoid generating lines and then parsing them appropriately. See https://docs.python.org/3/library/unittest.mock.html#where-to-patch for how to patch a method in a module. Just checking for the presence of `# EOF` at the end of the output could also tell the difference between plain and openmetrics if you do generate lines.

prometheus_client/exposition.py
ywwg: ok great, I'll keep OM falling back to OM

tests/openmetrics/test_parser.py
```diff
@@ -723,7 +723,7 @@ def collect(self):
 
         registry = CollectorRegistry()
         registry.register(TextCollector())
-        self.assertEqual(text.encode('utf-8'), generate_latest(registry))
+        self.assertEqual(text.encode('utf-8'), generate_latest(registry, ALLOWUTF8))
```
csmarchbanks, line 726: This probably doesn't actually need to change since there is a default value now.

tests/test_parser.py
```diff
@@ -367,7 +368,7 @@ def collect(self):
 
         registry = CollectorRegistry()
         registry.register(TextCollector())
-        self.assertEqual(text.encode('utf-8'), generate_latest(registry))
+        self.assertEqual(text.encode('utf-8'), generate_latest(registry, ALLOWUTF8))
```
csmarchbanks, line 371: Also unneeded here (also in various other tests, but these were the files that had the fewest changes).

csmarchbanks (approved):
👍 Thanks! Tested it for a bit and it is working well for me. There are some test changes where generate_latest now takes an escaping that isn't necessary now that there is a default value, but I am not opposed to passing in explicit values either.

tests/test_parser.py
ywwg: this one is actually needed. the default for generate_latest is underscores, not utf8, and this test contains dots.

csmarchbanks (approved):

PR #1103, opened 2025-04-19
docs: Add FastAPI + Uvicorn monitoring example

yuhao541:
## Description
This PR adds documentation for monitoring FastAPI applications running with Uvicorn ASGI server.


docs/content/exporting/http/fastapi-uvicorn.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+### Production Checklist
+1. **Directory Configuration**  
+   ```bash
+   chmod 750 ./metrics_data  # Secure write permissions
+   ```
+2. **Storage Management**  
+   - Use dedicated volume for `PROMETHEUS_MULTIPROC_DIR`
+   - Implement cleanup cron job:
```
csmarchbanks, line 105: Rather than a cron-job it is usually best practice to delete the data (if it exists) when your service restarts. Otherwise some problems with counter resets can occur.

csmarchbanks (commented):
Some comments on this approach, I do wonder if we should combine this with the existing Gunicorn page? Start with native, then have an additional section for gunicorn rather than two separate pages?

Issue #1108, opened 2025-05-27
[BUG] Prometheus parser 0.22.0 does not early skip comment lines


wissamir:
**Issue**

Starting with version 0.22.0, comment lines do not skip validation and parsing early enough and third token goes into [_is_valid_legacy_metric_name](https://github.com/prometheus/client_python/blob/master/prometheus_client/validation.py#L52) which results with an error when it does not match the [METRIC_NAME_RE](https://github.com/prometheus/client_python/blob/master/prometheus_client/validation.py#L4)

**Example:**

"""#  A random comment which goes into validation and passes"""
"""#  Another random. comment which goes into validation and fails"""
"""#  Whatever 335/125 random comment which goes into validation and fails"""

**Expected behaviour:**
Skip all comment  lines.  list([text_string_to_metric_families](https://github.com/prometheus/client_python/blob/master/prometheus_client/parser.py#L13)) should result with empty list []

**Current behaviour:**
A [ValueError](https://github.com/prometheus/client_python/blob/master/prometheus_client/parser.py#L313) is raised for the second and third comment lines above

I'll be pushing a PR for the fix

wissamir:
Fix proposal PR https://github.com/prometheus/client_python/pull/1109

PR #1109, opened 2025-05-27
BugFix: Skip validating and parsing commit lines early (#1108)

wissamir:
Related to https://github.com/prometheus/client_python/issues/1108


wissamir:
@csmarchbanks please mind reviewing this

csmarchbanks (commented):
Thank you for both the report and the fix! This looks reasonable to me, but I think the `else` can be removed from line 347 of your branch and the comment describing that this ignores other comment tokens would be nice to move up here.

wissamir:
Yep, that `else` branch became unreachable

All sorted out

csmarchbanks (approved):
Thanks!

PR #1115, opened 2025-06-26
Add support to `write_to_textfile` for custom tmpdir

aadityadhruv:
While the `try/except` block does prevent most of the temp files from persisting, if there is a non catchable exception, those temp files continue to pollute the directory. Optionally set the temp directory would let us write to something like `/tmp`, so the target directory isn't polluted. 

This does mean that the target and source filesystems could be different, hence I've used `shutil` here. I didn't want to modify the original behavior, specifically for the Windows bit, so I've kept it as is. `shutil` will use `os.rename` if the filesystems are the same, so we don't lose that atomic behavior in that case.

### DIFF
--- a/prometheus_client/exposition.py
+++ b/prometheus_client/exposition.py
@@ -4,6 +4,7 @@ from functools import partial
 import gzip
 from http.server import BaseHTTPRequestHandler
 import os
+import shutil
 import socket
 from socketserver import ThreadingMixIn
 import ssl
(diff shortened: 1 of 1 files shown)


csmarchbanks (commented):
Thanks!

I think I would prefer to just not support different filesystems. The atomic operation is important for the expected usage and I would prefer to fail than work around that.

Would you also add a bit to the help text describing what the tmpdir is there for and that it needs to be on the same filesystem?

aadityadhruv:
Thanks for the reply. The reason I've ended up adding support for the different filesystems is because `/tmp`, which is a pretty common location for temp files, is pretty much always on a different filesystem on Linux (`tmpfs`). If we don't allow different filesystem support, most people wouldn't be able to use `/tmp`.

aadityadhruv:
@csmarchbanks I've made the requested changes. Please let me know if it looks good!

csmarchbanks (approved):
Thanks!

Issue #1118, opened 2025-07-07
Implement deletion of Gauge metrics based on partially specified labels


TheGameProfi:
In many cases, the labels of a Gauge are fixed and known, but sometimes it’s hard to keep track of all labels, for example with an HTTP request collector per endpoint.

It would be very useful to have a method to delete metrics by matching a subset of labels. For example, deleting all metrics with `status_code="200"` regardless of other labels.

I ran into this issue myself when scraping disk data from multiple instances. On reset I wanted to remove only a specific instance's metrics, but ended up removing all.

The Go client (`client_golang`) has a `DeletePartiallyMatching` method.
https://github.com/prometheus/client_golang/issues/834

TheGameProfi:
Hey @hazel-shen , that really sounds like exactly what I was looking for.
I had thought about looking into how it could be implemented myself, but I just don’t have the time or focus for it right now.
So I really appreciate you taking the initiative to create a PR for this.

PR #1121, opened 2025-08-13
Add remove_matching() method for metric label deletion

hazel-shen:
**Description**
Fixes  [#1118](https://github.com/prometheus/client_python/issues/1118)
This PR introduces a new `remove_by_labels()` method to allow removing metric samples by matching a set of label values.

**Changes**
- Added `remove_by_labels()` to `MetricWrapperBase` for selective deletion of metrics by label.
- Included validation for label names and types.
- Added safeguards for multi-process mode (warning issued if used).
- Kept the return value consistent with `remove()`: returns None, and acts as a no-op if no series match.

**Motivation**
Currently, the Python client does not support removing metrics with specific label values without clearing the whole metric. This feature enables more granular control, which can be useful for scenarios such as:
- Cleaning up metrics for tenants after deprovisioning
- Removing outdated or irrelevant labeled data without impacting other metrics

**Example Usage**

```Python

c = Counter('my_counter', 'Example counter', ['tenant', 'status'])
c.labels(tenant='acme', status='ok').inc()
c.labels(tenant='acme', status='fail').inc()

# remove all series with tenant='acme', regardless of status
c.remove_by_labels({'tenant': 'acme'})

```

@csmarchbanks  Would you take a look when you have a chance?

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -203,7 +203,8 @@ class MetricWrapperBase(Collector):
             if labelvalues in self._metrics:
                 del self._metrics[labelvalues]
 
-    def remove_by_labels(self, labels: dict[str, str]) -> int:
+    def remove_by_labels(self, labels: dict[str, str]) -> None:
+        """Remove all series whose labelset partially matches the given labels."""
         if 'prometheus_multiproc_dir' in os.environ or 'PROMETHEUS_MULTIPROC_DIR' in os.environ:
             warnings.warn(
                 "Removal of labels has not been implemented in  multi-process mode yet.",
@@ -217,7 +218,7 @@ class MetricWrapperBase(Collector):
             raise TypeError("labels must be a dict of {label_name: label_value}")
     
         if not labels:
-            return 0
+            return  # no operation
     
         invalid = [k for k in labels.keys() if k not in self._labelnames]
         if invalid:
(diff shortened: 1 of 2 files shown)


prometheus_client/metrics.py
```diff
@@ -203,6 +203,41 @@ def remove(self, *labelvalues: Any) -> None:
             if labelvalues in self._metrics:
                 del self._metrics[labelvalues]
 
+    def remove_matching(self, partial: dict[str, str]) -> int:
```
csmarchbanks, line 206: What would you think of `remove_partial_match` for the name of this function? When I first see matching I think that all labelnames/values need to match to be removed not a partial. The name of the variable helps, but could also then be `labels` instead of `partial` which helps with confusion around thinking it could be a partial function.

prometheus_client/metrics.py
```diff
@@ -203,6 +203,41 @@ def remove(self, *labelvalues: Any) -> None:
...
+    
+        invalid = [k for k in partial.keys() if k not in self._labelnames]
+        if invalid:
+            raise ValueError(
+                'Unknown label names: %s; expected %s' % (invalid, self._labelnames)
+            )
+        
+        pos_filter = {self._labelnames.index(k): str(v) for k, v in partial.items()}
+
+        deleted = 0
```
csmarchbanks, line 230: What's the reasoning for returning the number of deleted items? We don't do that in remove today, but if there is a good use case I am not opposed.

csmarchbanks (commented):
Thanks!  A couple comments/questions to start with

TheGameProfi:
Thanks for the updated code, @csmarchbanks could we get another review on this PR?

csmarchbanks (approved):
Thanks!

PR #1124, opened 2025-08-28
Histogram observe() support count

acelyc111:


csmarchbanks (commented):
So the idea is that you want to observe `count` instances of the exact same `amount`? Can I ask what your use case for this is? My first thought is that it is quite niche and might be better served with a custom collector, but I want to keep my mind open.

acelyc111:
@csmarchbanks 
Thanks for your reply!
My use case is using Heatmap in Grafana via a Histogram in Prometheus, I can only get the sum of each bucket when I using Histogram, because the data source is from another module which I can't invasive modify too much, so I have to observe like:
```
age_to_count = ... # another module
for age, count in age_to_count:
  for _ in range(count):
    age_to_count_hist.observe(age)
```

It's a bit of waste because every `observe` is O(n) to find the corresponding bucket.

acelyc111:
I see, thanks!

csmarchbanks:
Glad that helped, I am going to close this for now, but let me know if the custom collector route is not working.

PR #1125, opened 2025-08-28
Fix bug which caused metric publishing to not accept query string parameters in ASGI app

hacksparr0w:
The `_bake_output` function in `exposition.py` expects the `params` argument to be a `dict[str, Any]` but the ASGI interface does not handle encoding, thus the parameters are passed in as `dict[bytes, Any]`.


csmarchbanks (commented):
Thanks! Would you be willing to write a test for this please? Or add enough type hinting that it would be caught by mypy.

csmarchbanks (approved):
Thanks!

PR #1127, opened 2025-08-30
Bump flake8 libraries

kajinamit:
... to the latest version available now. Also, do not pin bugfix version because any bugfix update is supposed to bring no breaking fix.


csmarchbanks (approved):
Thanks!

Issue #1122, opened 2025-08-22
Allow native histograms emission only when OM 2.0 is requested


vesari:
Native histograms should be emitted only when OM 2.0 is requested

PR #1128, opened 2025-09-03
Emit native histograms only when OM 2.0.0 is requested

vesari:
This PR addresses issue #1122 . In the present implementation, if the OM version negotiated is < 2.0.0, native histogram samples are completely ignored during the exposition.

As agreed with @csmarchbanks, I choose the approach of adding a `version` parameter to the `generate_latest` function, which this way can be intended as "generating the latest of a major version".

I added some tests; probably the `test_native_histogram_version_comparison` is a bit redundant, but gives an immediate idea of the different outputs depending on the OM version selected.

### DIFF
--- a/prometheus_client/openmetrics/exposition.py
+++ b/prometheus_client/openmetrics/exposition.py
@@ -4,6 +4,8 @@ from io import StringIO
 from sys import maxunicode
 from typing import Callable
 
+from packaging.version import Version
+
 from ..utils import floatToGoString
 from ..validation import (
     _is_valid_legacy_labelname, _is_valid_legacy_metric_name,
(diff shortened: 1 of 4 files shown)


prometheus_client/openmetrics/exposition.py
```diff
@@ -136,8 +140,12 @@ def generate_latest(registry, escaping=UNDERSCORES):
                             nh_exemplarstr = _compose_exemplar_string(metric, s, nh_ex)
                             exemplarstr += nh_exemplarstr
 
+                # Skip native histogram samples entirely if version < 2.0.0
+                if s.native_histogram and Version(version) < Version('2.0.0'):
```
krajorama, line 144: Couldn't we do this skip on line 95 and then not have to add `and Version(version) >= Version('2.0.0')` in other places?

krajorama: or maybe even further up?

PR #1139, opened 2025-09-17
Add an AIOHTTP exporter

Lexicality:
I wrote a significantly cruder version of this for a project and then realised that other people might find it useful so I spent a few hours writing a nicer one with tests and docs and so on.

I'm sure you don't want to have custom exporters for every http library under the sun but I figured since it's a pretty widely used library it'd be ok. Feel free to close if I was wrong.

cc @csmarchbanks


tox.ini
```diff
@@ -3,6 +3,8 @@ envlist = coverage-clean,py{3.9,3.10,3.11,3.12,3.13,py3.9,3.9-nooptionals},cover
 
 [testenv]
 deps =
+    aiohttp
```
csmarchbanks, line 6: Similar to twisted it would be good to only install this onto certain versions so that we don't break something core in environments where aiohttp is not present.

PR #1143, opened 2025-10-30
Add Django exporter (#1088)

### LINKED ISSUE #1088: Django explicit support
Hello, I was thinking it would be great to add explicit support for Django. Besides updating the documentation, it would be useful to provide a view so it can be added easily in the urlconf of a django app.

django-prometheus does it like [this](https://github.com/korfuri/django-prometheus/blob/master/django_prometheus/exports.py#L111) using `generate_latest`, but from looking at the already provided exposition endpoints, I was thinking of using the `_bake_output` utility instead.

I'm volunteering to make a PR for it.
What do you think?

Chadys:
Fixes #1088 

Finally got the time to propose what I had in mind :)

I added a utility view for easy inclusion of a metrics endpoint inside a Django project.

I also added some explanation about django-prometheus and the difference between what it provides and the new view.
I hope the difference is clear that way.

### DIFF
--- a/prometheus_client/django/exposition.py
+++ b/prometheus_client/django/exposition.py
@@ -1,0 +1,44 @@
+import os
+
+from django.http import HttpResponse
+from django.views import View
+
+import prometheus_client
+from prometheus_client import multiprocess
+from prometheus_client.exposition import _bake_output
+from prometheus_client.registry import registry
+
+
+class PrometheusDjangoView(View):
+    multiprocess_mode: bool = "PROMETHEUS_MULTIPROC_DIR" in os.environ or "prometheus_multiproc_dir" in os.environ
+    registry: prometheus_client.CollectorRegistry = None
+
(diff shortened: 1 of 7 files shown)


Chadys:
I can't seem to satisfy both isort and flake8_import_order, hence the CI error

prometheus_client/django/exposition.py
```diff
@@ -0,0 +1,44 @@
+import os
+
+from django.http import HttpResponse
+from django.views import View
+
+import prometheus_client
+from prometheus_client import multiprocess
+from prometheus_client.exposition import _bake_output
+from tests.test_exposition import registry
```
csmarchbanks:
```suggestion
from prometheus_client.registry import registry
```

I think this is what you meant and should also fix the lint errors!

csmarchbanks (approved):
Thanks!

PR #1158, opened 2026-03-24
allow for metric value class customization

mpnowacki-reef:


tests/test_multiprocess.py
```diff
@@ -646,3 +647,70 @@ def test_file_syncpath(self):
...
+        os.environ['PROMETHEUS_MULTIPROC_DIR'] = '/tmp'
+        importlib.reload(values)
+        self.assertEqual(values.ValueClass, values.MmapedValue)
+        del os.environ['PROMETHEUS_MULTIPROC_DIR']
+
+    def test_env_var_value_class(self):
+        # We need a class to point to. Let's use MutexValue itself but via string
+        os.environ['PROMETHEUS_VALUE_CLASS'] = 'prometheus_client.values.MutexValue'
+        importlib.reload(values)
+        self.assertEqual(values.ValueClass, values.MutexValue)
```
agoncharov-reef, line 704: This doesn't test anything because `MutexValue` is the default. Better point at something non-default, maybe to `values.MmapedValue` or even some random class from stdlib (we won't use it anyway, just need to check assignment)

prometheus_client/values.py
```diff
@@ -44,96 +121,35 @@ def MultiProcessValue(process_identifier=os.getpid):
...
+    value_class_path = os.environ.get('PROMETHEUS_VALUE_CLASS')
+    if value_class_path:
+        if '.' not in value_class_path:
+            raise ImportError(f"PROMETHEUS_VALUE_CLASS must be a full python path (e.g. module.ClassName), got '{value_class_path}'")
+        try:
+            module_path, class_name = value_class_path.rsplit('.', 1)
+            module = importlib.import_module(module_path)
+            return getattr(module, class_name)
+        except (ImportError, AttributeError) as e:
+            raise ImportError(f"Could not import PROMETHEUS_VALUE_CLASS '{value_class_path}': {e}") from None
```
agoncharov-reef, line 148:
```suggestion
            raise ImportError(f"Could not import PROMETHEUS_VALUE_CLASS '{value_class_path}'") from e
```

csmarchbanks (commented):
Can you describe your use case here a bit more? My initial reaction is that this is too complicated/niche to be supported by this client but I am happy to learn more.

PR #1159, opened 2026-04-01
docs: add API reference for all metric types 

### LINKED ISSUE #1021: real reference for API
> https://prometheus.github.io/client_python/

this is like a getting start hello world example

is there like a real API ref for the classes etc..

k1chik:
Closes #1021

Adds constructor parameter tables, method documentation, and runnable real world examples for Counter, Gauge, Histogram, Summary, Info, and Enum. The _index.md quick-pick table now covers all six types. Also fixes labels.md, which was missing remove(), remove_by_labels(), and clear(). The metric pages were already linking to them.

cc @csmarchbanks

### DIFF
--- a/docs/content/instrumenting/_index.md
+++ b/docs/content/instrumenting/_index.md
@@ -5,7 +5,7 @@ weight: 2
 
 Six metric types are available. Pick based on what your value does:
 
-| Type | Value goes | Use for |
+| Type | Behavior | Use for |
 |------|-----------|---------|
 | [Counter](counter/) | only up | requests served, errors, bytes sent |
 | [Gauge](gauge/) | up and down | queue depth, active connections, memory usage |
--- a/docs/content/instrumenting/_index.md
+++ b/docs/content/instrumenting/_index.md
@@ -5,7 +5,7 @@ weight: 2
 
 Six metric types are available. Pick based on what your value does:
 
-| Type | Behavior | Use for |
+| Type | Update model | Use for |
 |------|-----------|---------|
 | [Counter](counter/) | only up | requests served, errors, bytes sent |
 | [Gauge](gauge/) | up and down | queue depth, active connections, memory usage |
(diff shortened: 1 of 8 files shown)


docs/content/instrumenting/counter.md
```diff
@@ -3,12 +3,14 @@ title: Counter
...
-Counters go up, and reset when the process restarts.
+A Counter tracks a value that only ever goes up. Use it for things you count — requests
+served, errors raised, bytes sent. When the process restarts, the counter resets to zero.
 
+If your value can go down, use a [Gauge](../gauge/) instead.
 
 ```python
-c = Counter('my_failures', 'Description of counter')
+c = Counter('my_failures_total', 'Description of counter')
```
csmarchbanks: The library should add `_total` automatically so this is not necessary to change.

docs/content/instrumenting/_index.md
```diff
@@ -3,10 +3,20 @@ title: Instrumenting
 weight: 2
 ---
 
-Four types of metric are offered: Counter, Gauge, Summary and Histogram.
-See the documentation on [metric types](http://prometheus.io/docs/concepts/metric_types/)
+Six metric types are available. Pick based on what your value does:
+
+| Type | Value goes | Use for |
```
csmarchbanks: "Value goes" is a bit awkward for non counter/gauges. Maybe instead something like how a value is set?

csmarchbanks (commented):
Thanks! Generally this looks great, just a couple of small comments.

docs/content/instrumenting/_index.md
k1chik: thanks for the review!                                                                  
  - reverted the quick example in counter.md back to my_failures (without `_total`) 
  - renamed the `Value goes` column to `Update model` (does this work?)

csmarchbanks (approved):
I think that works, thank you!

PR #1166, opened 2026-04-23
Expose measured duration on Timer context manager

vokracko:
### Motivation

  When you use `.time()` as a context manager, the duration is already computed internally but thrown away — if you want to log or reuse it, you end up measuring twice:

  ```python
  start = time.monotonic()
  with REQUEST_LATENCY.time():
      response = handle_request()
  logger.info("request took %.3fs", time.monotonic() - start)
  ```

  That's redundant and slightly off (the two measurements don't cover the same interval). This change stores the observed value on the `Timer` so it can be read back after the block exits.

  ### Usage

  ```python
  with REQUEST_LATENCY.time() as t:
      response = handle_request()

  logger.info("request took %.3fs", t.duration)
  ```

  `t.duration` is `None` before the block runs and holds the observed seconds (the same value passed to `observe()` / `set()`) afterwards.

  ### Compatibility

  Purely additive. Existing `with metric.time():` usage is unchanged — the attribute is only visible if you bind the context manager with `as`.

### DIFF
--- a/docs/content/instrumenting/gauge.md
+++ b/docs/content/instrumenting/gauge.md
@@ -108,6 +108,10 @@ def process():
 
 with g.time():
     pass
+
+with g.time() as t:
+    pass
+# t.duration holds the observed seconds
 ```
 
 ### `set_function(f)`
--- a/docs/content/instrumenting/gauge.md
+++ b/docs/content/instrumenting/gauge.md
@@ -111,7 +111,7 @@ with g.time():
 
 with g.time() as t:
     pass
-# t.duration holds the observed seconds
+print(t.duration) # observed time in seconds.
 ```
 
 ### `set_function(f)`
--- a/docs/content/instrumenting/histogram.md
+++ b/docs/content/instrumenting/histogram.md
@@ -86,6 +86,10 @@ def process():
 
 with h.time():
     pass
+
+with h.time() as t:
+    pass
+# t.duration holds the observed seconds
 ```
 
 ## Labels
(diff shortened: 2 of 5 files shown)


docs/content/instrumenting/gauge.md
```diff
@@ -108,6 +108,10 @@ def process():
 
 with g.time():
     pass
+
+with g.time() as t:
+    pass
+# t.duration holds the observed seconds
```
csmarchbanks:
```suggestion
print(t.duration) # observed time in seconds.
```

Maybe something like this in these examples so that they will do something a bit different?

csmarchbanks (commented):
One small suggestion, but :+1:, nice little quality of life improvement!

docs/content/instrumenting/summary.md
```diff
@@ -68,6 +68,10 @@ def process():
 
 with s.time():
     pass
+
+with s.time() as t:
+    pass
+# t.duration holds the observed seconds
```
vokracko:
```suggestion
print(t.duration) # observed time in seconds.
```

docs/content/instrumenting/histogram.md
```diff
@@ -86,6 +86,10 @@ def process():
 
 with h.time():
     pass
+
+with h.time() as t:
+    pass
+# t.duration holds the observed seconds
```
vokracko:
```suggestion
print(t.duration) # observed time in seconds.
```

csmarchbanks (approved):
Thanks!

PR #1168, opened 2026-04-23
docs: add Celery exporter example

### LINKED ISSUE #902: Add usage example for Celery
I recently wanted to collect custom application metrics from a rather large Celery app. I tried several approaches with varying degrees of awkwardness, until finally settling on having a "sidecar" HTTP server process for all Celery nodes.  
This works pretty well, but took me a while to get right. To spare others from wasting as much time as me, I would like to propose adding a section on getting Celery metrics set up with the Prometheus client. I wrote up the solution on StackOverflow: https://stackoverflow.com/a/75799358/2532203

If you're interested, I can properly summarise it and open a PR?

Also, if you look at the code, I needed to copy some swaths of the library because the built-in Prometheus WSGI server is [hard-coded to demonise the thread](https://github.com/prometheus/client_python/blob/30f83196ac1f1a1a2626da9d724d83955aff79b1/prometheus_client/exposition.py#L173); this causes trouble when running in a demonised process. Maybe this could be made optional with another keyword argument that defaults to True?

ayushanand293:
Closes #902

Testing:
- python -m pytest -q

### DIFF
--- a/docs/content/exporting/http/celery.md
+++ b/docs/content/exporting/http/celery.md
@@ -1,0 +1,65 @@
+---
+title: Celery
+weight: 60
+---
+
+## Celery
+
+When using Celery, a common approach is to instrument tasks with `prometheus_client` metrics and expose them via a small sidecar exporter process on each node. Prometheus scrapes the sidecar HTTP endpoint.
+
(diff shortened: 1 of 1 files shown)


ayushanand293:
Added a Celery sidecar exporter example per #902. Tests: python -m pytest -q.

PR #1169, opened 2026-04-24
docs: add API reference for CollectorRegistry and custom collector classes

### LINKED ISSUE #1163: docs: add API reference for CollectorRegistry and custom collector classes
Following the API reference work in #1021 (metric types) and #1161 (pushgateway, textfile, multiprocess), CollectorRegistry and the collector classes have the same gap, examples without formal parameter documentation.

### CollectorRegistry (no dedicated doc page exists)

The object every user interacts with implicitly (REGISTRY) or explicitly when creating isolated registries. The constructor has undocumented parameters:

- auto_describe
- target_info
- support_collectors_without_names

Undocumented public methods: register(collector), unregister(collector), collect(), restricted_registry(names), get_sample_value(name, labels), get_target_info(), set_target_info()

### Custom Collectors (docs/content/collector/custom.md)

Currently 38 lines with a working example. Missing:

- GaugeMetricFamily(name, documentation, value, labels, unit) — constructor params
- CounterMetricFamily(name, documentation, value, labels, created, unit, exemplar) — constructor params
- SummaryMetricFamily(name, documentation, count_value, sum_value, labels, unit) — constructor params
- HistogramMetricFamily(name, documentation, buckets, sum_value, labels, unit) — constructor params
- InfoMetricFamily(name, documentation, value, labels) — constructor params
- add_metric(labels, value) on each family — not formally documented
- The Collector protocol (describe() and collect()) — only shown by example
- The mutual exclusivity of value vs labels in MetricFamily constructors — not explained

### Built-in Collectors (docs/content/collector/_index.md)

Currently 36 lines. Missing:

- ProcessCollector(namespace, pid, proc, registry) — proc and registry params undocumented
- PlatformCollector(registry, platform) — constructor params undocumented
- GCCollector(registry) — constructor params undocumented
- How to selectively disable specific default collectors

## Proposed approach

Add "## API Reference" sections to each existing page with:

- Function parameter tables (name, type, default, description)
- Return values
- Error conditions
- A validation script confirming every claim against the source

## Source files to reference

- prometheus_client/registry.py — CollectorRegistry
- prometheus_client/metrics_core.py — all MetricFamily classes
- prometheus_client/gc_collector.py — GCCollector
- prometheus_client/platform_collector.py — PlatformCollector
- prometheus_client/process_collector.py — ProcessCollector

kkukadia:
Closes #1163

Follows the same pattern as #1021 and #1162.

**collector/custom.md**: Collector protocol section (collect/describe), value vs labels mutual exclusivity note with correct per-type parameter names, full constructor and add_metric tables for GaugeMetricFamily, CounterMetricFamily, SummaryMetricFamily, HistogramMetricFamily, and InfoMetricFamily, plus a runnable real-world example.

**collector/_index.md**: constructor parameter tables for ProcessCollector, PlatformCollector, and GCCollector, with exported metrics listed for each.

**registry/_index.md** (new page): CollectorRegistry constructor and all public methods (register, unregister, collect, restricted_registry, get_sample_value, set_target_info, get_target_info), the global REGISTRY instance, and examples for isolated registry usage and registry=None.

All code examples verified by running them in Python.

cc @csmarchbanks

### DIFF
--- a/docs/content/collector/_index.md
+++ b/docs/content/collector/_index.md
@@ -18,8 +18,8 @@ ProcessCollector(namespace='mydaemon', pid=lambda: open('/var/run/daemon.pid').r
 # Platform Collector
 
 The client also automatically exports some metadata about Python. If using Jython,
-metadata about the JVM in use is also included. This information is available as 
-labels on the `python_info` metric. The value of the metric is 1, since it is the 
+metadata about the JVM in use is also included. This information is available as
+labels on the `python_info` metric. The value of the metric is 1, since it is the
 labels that carry information.
 
 # Disabling Default Collector metrics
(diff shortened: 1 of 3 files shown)


csmarchbanks (approved):
Thanks!

calestyo:
Hey.

Thanks for improving on the documentation... that really helps a lot, compared to few years ago when I had to basically fiddle out most of these things from the code :-)

But one thing:

Why the `yield`s in the `collect()`? I don’t quite get that.

Shouldn't `collect()` return sequences of `*MetricFamily` objects? You now return only a single one, and there even alternating ones like in:
https://github.com/prometheus/client_python/pull/1169/changes/a708480f60a0a03abb378c16be328fe9598f4a02#diff-eadfec94068584be6fec5cce354e065ead34a754bc84ead564adf9e8c497ab10R237-R243

Or maybe you mean this as two independent examples?
What makes it confusing though is (which is however not from your commit):
https://github.com/prometheus/client_python/blob/482656c8c07b78668adb5f3171dfe71ccc2396b2/docs/content/collector/custom.md?plain=1#L16-L21

There `collect()` alternates between returning (well yielding) a Gauge and Counter (also not as sequences)?

I generally don't quite understand why one would `yield` here at all.

Isn't the reason for custom collectors to start every time fresh (and get rid of all stale metrics, labels, etc.)?
With `yield` wouldn't you keep all that?

Thanks,
Chris.

calestyo:
Shouldn't do stuff late night... `yield` of course makes it an iterator o.O

Still I wonder, is there any advantage over using that in the current form rather than `return`?

I mean is Python' GC smart enough that it can already clean the objects that have already been `yield`ed and thus give an advantage by reducing e.g. memory use for very large metric sets?

In:
```python
        yield GaugeMetricFamily('my_gauge', 'Help text', value=7)
        c = CounterMetricFamily('my_counter_total', 'Help text', labels=['foo'])
        c.add_metric(['bar'], 1.7)
        c.add_metric(['baz'], 3.8)
        yield c
```
I could imagine that it may release the `GaugeMetricFamily` object early... but `c`? It still has a reference?

In any case it would be nice to have in the docs **why** generators are used (if it's memory advantages) and how to use them so that this actually works out. :-)

kkukadia:
Hi @calestyo, thanks for the read! glad to see you've found the docs useful. 

And yes, my bad, the code examples are missing the `def collect(self):` wrapper which makes it unclear what context they're meant to be in, i'll fix that asap. 

Your GC read is right too, yield `GaugeMetricFamily(...)` inline like that can be reclaimed as soon as the caller moves on, but `c` keeps a reference in the generator frame until it's done. The main reason for yield here isn't memory though, it's just that the registry iterates over whatever `collect()` returns and a generator is cleaner than building a list. For most collectors you'd never notice the difference anyway.

I'll open a follow-up PR with both fixes and add a note explaining why generators are used. Thanks again for your feedback. 

>Shouldn't do stuff late night... - :-)

calestyo:
Hmm, but if you say memory isn't the main reason, the real one should perhaps be given?

And if memory is at least part of the reason, then it might have made sense to explicitly mention that and also explain to people that ideally they should re-use the same identifier so that the GC can kick in.

Even perhaps if they have a case where it's reasonable to first collect a few metrics and only `yield` all of them afterwards, like in:
```
for loopOver in SomeJSONelements:
    metric[0].add_metric()
    metric[1].add_metric()
for m in metric:
    yield m

#... next metric again uses `metric` var
```
I.e. telling that it's still good to use only one identifier `metric` here, so that the GC can work

kkukadia:
Hi Chris, thanks for the response. Good point on both counts! I'll add a note that memory is at least part of the reason for using yield, and mention that reusing the same identifier helps the GC kick in sooner. PR coming shortly :)

PR #1170, opened 2026-04-26
A redis-backed storage backend

stefanor:
The multiprocess backend has an issue with database file explosion (#204, #518, etc.)

In the environment that I'm running into this problem, we have a redis database available, so why not take advantage of this?

Here is a fairly naive proof-of-concept example.

Also compatible with Valkey.


stefanor:
FWIW, we're using this in production: https://salsa.debian.org/freexian-team/debusine/-/merge_requests/3013

So far, it seems to work well.

SHxKM:
@stefanor this might solve quite a big headache for me, is it still working for you guys?

stefanor:
Yes, we haven't had to touch it since.

PR #1172, opened 2026-05-04
docs: follow-up fixes for collect() generator examples (#1169)

kkukadia:
@csmarchbanks

Three issues raised in review of #1169 by @calestyo:

1. The API Reference code examples used bare `yield` statements without
   any surrounding `def collect(self):` context, making it unclear what
   they belonged to. Added a preamble to the API Reference section
   pointing back to the top-level example.

2. The `InfoMetricFamily` example had two `yield` statements in a single
   block, which looked like one `collect()` yielding both patterns rather
   than two alternatives. Split into separate labelled blocks.

3. The `collect()` protocol section gave no explanation for why generators
   are used. Added a note explaining that yield is idiomatic (lazy
   iteration, no intermediate list) and that each scrape gets a fresh call
   with no state carried over.


csmarchbanks (approved):
Thanks!

PR #1173, opened 2026-05-08
docs: add memory benefit and variable-reuse note to collect() section

kkukadia:
Closes points made from comment section in #1169.

**Changes**

- Added a paragraph to the \`collect()\` protocol section noting that memory
  is a secondary benefit of using \`yield\`: yielding inline lets Python
  reclaim the object as soon as the registry advances, while a named variable
  stays alive until it is rebound.
- Added a short loop example showing that yielding inside a loop rather than         
  accumulating into a list keeps at most one object alive at a time.    
- Example verified by running it against the library.

cc @calestyo @csmarchbanks

### DIFF
--- a/docs/content/collector/custom.md
+++ b/docs/content/collector/custom.md
@@ -56,16 +56,15 @@ Memory is a secondary benefit. Yielding an object inline
 (`yield GaugeMetricFamily(...)`) lets Python reclaim it as soon as the registry
 advances past that point. A named variable keeps the object alive until the variable
 is rebound. For most collectors this difference is negligible, but it can matter for
-very large metric sets. When building families in a loop, yield inside the loop and
-reuse the same variable name so the GC can reclaim each object before the next one
-is built:
+very large metric sets. Yielding inside a loop rather than accumulating into a list
+keeps at most one object alive at a time:
 
 ```python
 def collect(self):
     for name in _QUEUE_NAMES:
         m = GaugeMetricFamily('queue_depth', 'Queue depth', labels=['queue'])
         m.add_metric([name], fetch_depth(name))
-        yield m  # m is rebound next iteration; the previous object can be reclaimed
+        yield m
 ```
 
 ### `describe()`
--- a/docs/content/collector/custom.md
+++ b/docs/content/collector/custom.md
@@ -52,6 +52,22 @@ into a generator, which the registry iterates lazily without building an interme
 list first. Each scrape calls `collect()` fresh, so no state carries over between
 scrapes.
 
+Memory is a secondary benefit. Yielding an object inline
+(`yield GaugeMetricFamily(...)`) lets Python reclaim it as soon as the registry
+advances past that point. A named variable keeps the object alive until the variable
+is rebound. For most collectors this difference is negligible, but it can matter for
+very large metric sets. When building families in a loop, yield inside the loop and
+reuse the same variable name so the GC can reclaim each object before the next one
+is built:
+
+```python
+def collect(self):
+    for name in _QUEUE_NAMES:
+        m = GaugeMetricFamily('queue_depth', 'Queue depth', labels=['queue'])
+        m.add_metric([name], fetch_depth(name))
+        yield m  # m is rebound next iteration; the previous object can be reclaimed
+```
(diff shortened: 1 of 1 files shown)


kkukadia:
Closing to reopen with a proper description.

calestyo:
Some comments on this... or maybe... questions:

First what I often have (actually in a exporter I'm developing right now) is something like:
```python
        labels = ("id", )
        metric_info = InfoMetricFamily(self.name_prefix + "space_reservation", "Information about a space reservation.",
                                       labels=labels)
        metric_space_total = GaugeMetricFamily(self.name_prefix + "space_reservation_space_total", "The total space of a space reservation in bytes.",
                                               unit="bytes",
                                               labels=labels)
        metric_space_used = GaugeMetricFamily(self.name_prefix + "space_reservation_space_used", "The used space of a space reservation in bytes.",
                                              unit="bytes",
                                              labels=labels)
        metric_space_allocated = GaugeMetricFamily(self.name_prefix + "space_reservation_space_allocated", "The allocated space of a space reservation in bytes.",
                                                   unit="bytes",
                                                   labels=labels)
        for s in raw_data.space_reservations:
            s_labels = ( str(s["id"]), )
            metric_info.add_metric(labels=s_labels,
                                   value={"description": s["description"]}
                                  )
            metric_space_total.add_metric(s_labels,
                                          value=s["sizeInBytes"]
                                         )
            metric_space_used.add_metric(s_labels,
                                         value=s["usedSizeInBytes"]
                                        )
            metric_space_allocated.add_metric(s_labels,
                                              value=s["allocatedSpaceInBytes"]
                                             )
        yield metric_info
        yield metric_space_total
        yield metric_space_used
        yield metric_space_allocated
```
That's not the final code I'm going to use (because it's actually a bit stupid with respect to the GC thingy.

What I mean to tell is the (IMO general) use case that I iterate over something and draw multiple metrics from it.
Now the above, with using all the different metric_* vars is bad with respect to the GC reclaim, in particular as I don't use the same names later (i.e. in other metrics after the `yield`s above)... so the objects would still be referenced and the GC couldn't delete them.

I think using a **single** var `metric` (for all) is also impossible or at least stupid: While I could of course `yield` in the for loop, using a single `metric` var, would also mean that I'd need to define the `GaugeMetricFamily` in ever iteration of the loop, which is of course also bad.

So my point here is:
I think it might be reasonable to teach users in the docs how to handle such case.

My "solution" (though) it's not perfect right now is:
- use a single `metric` var, but if there are multiple metrics, make it a `SimpleNamespace` like in:
```python
            metrics = SimpleNamespace(
                                      info=InfoMetricFamily(self.name_prefix + "cell", "Information about a cell.",
                                                            labels=l),
                                     other_metric= ...
                                     )

```
One could of course also use a tuple (which probably would be faster), but if one has many metrics in one go, referring them via numbers might be prone to human errors.
`namedtuple` has the problem that it actually generates a type (which seems a bit overkill for this), plus the creation would be a bit "decoupled" like in:
```python
metrics = namedtuple(typename="metrics", field_names=["info","other_metric"])(InfoMetricFamily(...),...)
```
The downside with `SimpleNamespace` is that it's not directly iterable but one can do e.g. `yield from metrics.__dict__.values()`.

`dict` would of course work, and is probably not slower than `SimpleNamespace` which also uses a `dict` (or even faster)... but it felt a bit more ugly to write for me.

So yeah... that might be something to include in the docs....


Second, and that's more of a question...

We try here to educate the users how to make their collection faster, right!?
But the general schema is still, that in every call to `collect()`, all the `*MetricFamily` objects are completely created from scratch... including all their label sets (which I guess typically won't change).

So:
- wouldn't it be better if the `*MetricFamily` objects are created once, e.g. in `__init__`
- and - to get rid of metrics - have them all cleared in the beginning of each `collect()`?

AFAICS, all these classes simply store their metrics in `.samples`, which is a list, everything else seems static.
One might even argue that `.samples` is already "public", though a `clear()` method wouldn't harm I guess.

calestyo:
Yeah... as you can see I made a draft PR for the 2nd point above... would be happy if you could have a look at it.

PR #1178, opened 2026-05-27
Add TLS version parameters for start_wsgi_server

vyzigold:
This adds the tls_min_version and tls_max_version parameters for the start_wsgi_server (and start_http_server), allowing to restrict the min/max TLS version negotiated for accessing the server.

New unit tests under TestWsgiTLS were added, which test the version parameters as well as the rest of TLS capabilities for the start_wsgi_server (TLS on/off, mTLS), since those previously didn't have any test coverage.


vyzigold:
@csmarchbanks what do you think?

prometheus_client/exposition.py
```diff
@@ -196,6 +196,8 @@ def _get_ssl_ctx(
         cafile: Optional[str] = None,
         capath: Optional[str] = None,
         client_auth_required: bool = False,
+        tls_min_version: ssl.TLSVersion = ssl.TLSVersion.MINIMUM_SUPPORTED,
```
csmarchbanks: I believe this would end up setting a default lower than Python's default. Can we use None instead and then only set `ssl_cxt.minimum_version` if a value is specified? 

For consistency probably update maximum too, but I don't think that is as critical.

csmarchbanks (commented):
Thanks, one concern but otherwise :+1:

prometheus_client/exposition.py
vyzigold: Good find, I agree. I pushed an update with your suggestion https://github.com/prometheus/client_python/compare/3406ec2da63f7436f9b62b89de83696523d5efe3..61b66b05aaeb954ef9d4ebff901ebbce96f83650

csmarchbanks (approved):
Thanks!

PR #1180, opened 2026-06-04
fix: escape colons in metric names when escaping=underscores

### LINKED ISSUE #1177: OpenMetrics content negotiation: `escaping=underscores` causes name mismatch between `HELP/TYPE` metadata and sample lines for metrics with colons
### Description

When exposing metrics in the OpenMetrics format using content negotiation (via `Accept: application/openmetrics-text; version=1.0.0`), metrics containing colons (`:`) have their names escaped in the sample lines but remain unescaped in the `# HELP` and `# TYPE` lines.

This produces mismatched exposition output that violates the OpenMetrics standard, causing strict standard parsers (including `promtool` check metrics and downstream ingesters) to fail or treat them as completely separate metrics (an orphaned metadata set and an untyped metric).

### Steps to Reproduce

1. Create a simple Python server using `prometheus_client`:
```python
from http.server import BaseHTTPRequestHandler, HTTPServer
from prometheus_client import CollectorRegistry, Gauge
from prometheus_client.exposition import choose_encoder
  
token_usage = Gauge(
    name="sglang:token_usage",
    documentation="Total token usage.",
    registry=registry,
)
token_usage.set(42.0)
  
    def do_GET(self):
        if self.path == "/metrics":
            accept_header = self.headers.get("Accept", "")
            encoder, content_type = choose_encoder(accept_header)
            data = encoder(registry)
            self.send_response(200)
            self.send_header("Content-Type", content_type)
            self.end_headers()
            self.wfile.write(data)
  
if __name__ == "__main__":
    server = HTTPServer(("0.0.0.0", 8000), MetricsHandler)
    server.serve_forever()
```

2. Query the endpoint requesting OpenMetrics formatting:
```bash
curl -i -H "Accept: application/openmetrics-text; version=1.0.0" http://localhost:8000/metrics
```

### Actual Output
Notice that the `Content-Type` is negotiated with `escaping=underscores`, but the comment lines mismatch the sample line:
```http
HTTP/1.0 200 OK
Content-Type: application/openmetrics-text; version=1.0.0; charset=utf-8; escaping=underscores
  
# HELP sglang:token_usage Total token usage.
# TYPE sglang:token_usage gauge
sglang_token_usage 42.0
# EOF
```

### Expected Output
The metric name inside the `# HELP` and `# TYPE` comment blocks should match the name used in the sample line:
```http
# HELP sglang_token_usage Total token usage.
# TYPE sglang_token_usage gauge
sglang_token_usage 42.0
# EOF
```

### Root Cause
In `prometheus_client/openmetrics/exposition.py`:

- The comment generator uses the unescaped `metric.name` directly (since a colon is treated as legacy-valid under default validation checks).
- The sample line generator, however, strictly sanitizes the sample's name against the narrower OpenMetrics name requirements, replacing the colon with an underscore under `escaping=underscores`.

johnsaurabh:
Fixes #1177

When a metric name contains a colon (e.g. `sglang:token_usage`) and the client negotiates `escaping=underscores`, the `# HELP` and `# TYPE` lines were emitting the raw name while the sample line replaced the colon with an underscore:

```
# HELP sglang:token_usage Total token usage.
# TYPE sglang:token_usage gauge
sglang_token_usage 42.0
```

This violates the OpenMetrics spec and causes strict parsers to treat the metadata and the sample as two separate metrics.

The bug is in `escape_metric_name` in `prometheus_client/openmetrics/exposition.py`. In `UNDERSCORES` mode, the function short-circuits for any name that matches the legacy Prometheus metric name regex. That regex allows colons, so names like `sglang:token_usage` were returned unchanged. The sample line uses `_is_legacy_labelname_rune` which does not allow colons, so it escaped them correctly.

The fix adds a colon check to the short-circuit condition and switches the fallback `_escape` call to use `_is_legacy_labelname_rune`, matching what the sample line already does.

Two existing test expectations that documented the old behavior are updated. A new end-to-end test reproduces the exact scenario from the issue.

@csmarchbanks

### DIFF
--- a/prometheus_client/openmetrics/exposition.py
+++ b/prometheus_client/openmetrics/exposition.py
@@ -181,9 +181,9 @@ def escape_metric_name(s: str, escaping: str = UNDERSCORES) -> str:
             return '"{}"'.format(_escape(s, escaping, _is_legacy_metric_rune))
         return _escape(s, escaping, _is_legacy_metric_rune)
     elif escaping == UNDERSCORES:
-        if _is_valid_legacy_metric_name(s):
+        if _is_valid_legacy_metric_name(s) and ':' not in s:
             return s
-        return _escape(s, escaping, _is_legacy_metric_rune)
+        return _escape(s, escaping, _is_legacy_labelname_rune)
     elif escaping == DOTS:
         return _escape(s, escaping, _is_legacy_metric_rune)
     elif escaping == VALUES:
--- a/tests/openmetrics/test_exposition.py
+++ b/tests/openmetrics/test_exposition.py
@@ -57,6 +57,15 @@ utf8_cc_created 123.456
 # EOF
 """ == generate_latest(self.registry, UNDERSCORES)
 
+    def test_gauge_colon_in_name_escaped_underscores(self):
+        g = Gauge('sglang:token_usage', 'Total token usage.', registry=self.registry)
+        g.set(42.0)
+        assert b"""# HELP sglang_token_usage Total token usage.
+# TYPE sglang_token_usage gauge
+sglang_token_usage 42.0
+# EOF
+""" == generate_latest(self.registry, UNDERSCORES)
+
     def test_counter_total(self) -> None:
         c = Counter('cc_total', 'A counter', registry=self.registry)
         c.inc()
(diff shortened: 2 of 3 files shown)


csmarchbanks (commented):
Thanks! Generally this looks good to me, though you need to sign the DCO.

PR #1182, opened 2026-06-15
Simplify label boundary detection in sample parser

afrkorsakow:
Optimize line parsing by using `str.find()` for locating `{` and `str.rfind()` for location `}`

Since metric names cannnot contain `{` and values are always numeric, escape handling is entirely unnecessary for bounday detection.  This imporvemnt scales with line length.

## benchmark

```
git switch master
pytest tests/test_parser.py  --benchmark-only --benchmark-save master
git switch feature/performe
pytest tests/test_parser.py  --benchmark-only --benchmark-compare 0002_master
```

```
----------------------------------------------------------------------------------------------------------- benchmark: 2 tests -----------------------------------------------------------------------------------------------------------
Name (time in us)                                                    Min                 Max                Mean             StdDev              Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_benchmark_text_string_to_metric_families (NOW)             115.9000 (1.0)      205.4800 (1.0)      128.7905 (1.0)      13.6625 (1.0)      123.5150 (1.0)       9.7862 (1.0)       438;402        7.7645 (1.0)        3587           1
test_benchmark_text_string_to_metric_families (0002_master)     147.6700 (1.27)     319.5870 (1.56)     168.5964 (1.31)     18.2601 (1.34)     161.9480 (1.31)     14.6172 (1.49)      480;383        5.9313 (0.76)       3373           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/prometheus_client/parser.py
+++ b/prometheus_client/parser.py
@@ -244,7 +244,7 @@ def _parse_value(value):
 def _parse_sample(text):
     separator = " # "
     # Detect the labels in the text
-    label_start = _next_unquoted_char(text, '{')
+    label_start = text.find('{')
     if label_start == -1 or separator in text[:label_start]:
         # We don't have labels, but there could be an exemplar.
         name_end = _next_unquoted_char(text, ' \t')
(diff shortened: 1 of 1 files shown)


csmarchbanks (commented):
Thanks!

With the new UTF-8 support there can be escaped braces in a metric name. I think that means this cannot be done as it could find a set of labels mid-metric and then error. We should probably add a test case for that to make sure however, and it is definitely possible we could continue to optimize the parser performance.

afrkorsakow:
Sorry for the late reply.
I hadn't realized UTF-8 support was an issue.
I'll try to see if I can also cover the UTF-8 approach.

PR #1184, opened 2026-06-23
Make clear() a no-op for label-less metrics (fixes #707)

### LINKED ISSUE #707: Calling clear() on Counter without labels fails
To reproduce:
```

c = Counter("c", "c")
c.inc()
c.clear()
```

The error returned is: 
```
AttributeError: 'Counter' object has no attribute '_lock'
```

Is using .clear() in this way supported?

ompathakbaap:
Fixes #707.

`clear()` unconditionally did `with self._lock:`, but `self._lock`
is only created in `__init__` for metrics with labelnames. Calling
`.clear()` on a label-less metric (e.g. `Counter('c', 'c')`) raised
`AttributeError: 'Counter' object has no attribute '_lock'`.

Per the discussion in the issue, a label-less metric has no
labelsets to clear, so this makes `clear()` a no-op in that case
rather than raising  consistent with the direction @csmarchbanks
and others converged on in the thread.

Added a regression test in `tests/test_core.py::TestCounter`.

@csmarchbanks would appreciate your review on this when you have a chance.

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -254,6 +254,8 @@ class MetricWrapperBase(Collector):
 
     def clear(self) -> None:
         """Remove all labelsets from the metric"""
+        if not self._labelnames:
+            return
         if 'prometheus_multiproc_dir' in os.environ or 'PROMETHEUS_MULTIPROC_DIR' in os.environ:
             warnings.warn(
                 "Clearing labels has not been implemented in multi-process mode yet",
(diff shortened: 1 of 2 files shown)


csmarchbanks (commented):
The change in this PR does not match the description, pushed the wrong change perhaps?

ompathakbaap:
I am so sorry about that, a branch mix up on my end (the WSL docs note and this fix ended up on the same branch pointer). Force pushed the corrected commit just now, the diff now shows the clear() change in metrics.py and regression test in test_core.py

csmarchbanks (approved):
Thanks!

PR #1185, opened 2026-06-25
fix: make test suite pass on Windows and Python 3.12+

MelvinjoseC:
- Fix event loop issues in ASGI tests under newer asgiref by wrapping communicator setup in an async coroutine run on the loop.
- Fix mmap file locking (PermissionError: [WinError 32]) on Windows in multiprocess tests by closing open database handles before unlinking/removing directories.
- Skip case-sensitive env var deprecation warning assertion on Windows where environment variables are case-insensitive.
- Conditionally define the parser benchmark test to avoid failure when pytest-benchmark is not installed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -108,3 +108,17 @@ jobs:
         run: tox
         env:
           TOXENV: pypy${{ env.PYTHON_VERSION }}
+
+  test_windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - name: Set up Python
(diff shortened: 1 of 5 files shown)


csmarchbanks (commented):
Thanks!

A few lint errors and such in CI, but generally I am happy with this. Would you also be interested in adding a CI pipeline for windows to avoid regressions? Just one python version is probably enough.

MelvinjoseC:
Thanks for the feedback! I've fixed the remaining lint/type issues and added a Windows CI job using Python 3.12. All local checks pass, and GitHub Actions are now running.

csmarchbanks:
Thanks! Everything looks good now, but the DCO needs to be signed on your most recent commit.

MelvinjoseC:
Thanks! I've added the DCO sign-off to the latest commit. The DCO check is now passing. The remaining CI workflow is awaiting maintainer approval.

csmarchbanks (approved):
Thanks!

PR #1186, opened 2026-06-26
Use object instead of Any for labels() parameters

gregoiredx:
I'm working on a project where we try to reduce lines flagged as `Any` by [mypy line precision report](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-lineprecision-report).

The type of `labels()` parameters is `Any`, reducing the whole line as `Any` (mypy takes the worst of the line typing for its report).

Using `object` here looks like the right choice.

From https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-any-type
> This should not be confused with the object type, which represents the set of all values. Unlike object, Any introduces type unsafety — see [Any vs. object](https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object) for more.

I first thought it would even be `str` since it's the natural type of labels, but `labels()` transforms parameters to strings automatically (`labelvalues = tuple(str(labelkwargs[l]) for l in self._labelnames)` and `labelvalues = tuple(str(l) for l in labelvalues)`).

So we mean "anything that can be `str`ed", `object` looks like a good candidate for that type.

Note: I had to introduce `str_labelvalues` to comply with `mypy` (since now `labelvalues` has type `tuple[object, ...]` which is not compatible with ` tuple[str, ...]`).

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -135,7 +135,7 @@ class MetricWrapperBase(Collector):
             if registry:
                 registry.register(self)
 
-    def labels(self: T, *labelvalues: Any, **labelkwargs: Any) -> T:
+    def labels(self: T, *labelvalues: object, **labelkwargs: object) -> T:
         """Return the child for the given labelset.
 
         All metrics can have labels, allowing grouping of related time series.
@@ -173,13 +173,13 @@ class MetricWrapperBase(Collector):
         if labelkwargs:
             if sorted(labelkwargs) != sorted(self._labelnames):
                 raise ValueError('Incorrect label names')
-            labelvalues = tuple(str(labelkwargs[l]) for l in self._labelnames)
+            str_labelvalues = tuple(str(labelkwargs[l]) for l in self._labelnames)
         else:
             if len(labelvalues) != len(self._labelnames):
                 raise ValueError('Incorrect label count')
-            labelvalues = tuple(str(l) for l in labelvalues)
+            str_labelvalues = tuple(str(l) for l in labelvalues)
         with self._lock:
-            if labelvalues not in self._metrics:
+            if str_labelvalues not in self._metrics:
 
                 original_name = getattr(self, '_original_name', self._name)
                 namespace = getattr(self, '_namespace', '')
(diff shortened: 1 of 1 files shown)


gregoiredx:
Hello @csmarchbanks 👋  Following the CONTRIBUTING.md, I should ping you to get your review for a trivial change.

This is just a type change from `Any` to `object`. What do you think about it?

csmarchbanks (commented):
Thanks for the PR!

I don't feel like it helps all that much though, I don't love introducing another variable just to appease the typechecker when Any still works (or will raise) today. What's the goal of reducing Any usage?

gregoiredx:
Thanks for looking into it!

> What's the goal of reducing Any usage?

The name Any may be a bit counter intuitive, but it's usage is meant to say "I don't know, I want to opt out from type checking here" and not just "this can be of any type". When you want the type checker to look at the code while making sure it can handle any type, then the proper type to use is `object`.
[mypy doc is pretty clear about that](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-any-type):
> You can use Any as an “escape hatch” when you can’t use a more precise type for some reason.
This should not be confused with the [object](https://docs.python.org/3/library/functions.html#object) type, which represents the set of all values. Unlike object, Any introduces type unsafety — see [Any vs. object](https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object) for more.

That's actually exactly what happened to `labelvalues`. It was using the "escape hatch" so mypy did not type check anything related to  `labelvalues` but when I switched to `object`, mypy started to actually check this variable. 

In our case, we have introduced type checking on a big code base (>1M lines). We can't type everything at once and we need a metric to be able to follow our progress. [mypy reports](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-lineprecision-report) help us do this by giving statistics per lines and modules for what it's able to infer. Lines using the Any "escape hatch" intruduce "type unsafety" (by mypy definition). Those are lines where mypy is in effect useless. Since we want to enforce typing, measuring that this number goes down is highly correlated to our code being actually type safe.

> I don't love introducing another variable just to appease the typechecker

Me neither. You can say it's just to appease the typechecker but one could say it's to fix types in this method.
It's the cost of going from a purely dynamically typed language to a type checked one. Some things are not doable anymore. 
In this case:
- assigning `tuple[str, ...]` to a variable already typed `tuple[object, ...]`, see ["invariance"](https://mypy.readthedocs.io/en/stable/common_issues.html#variance) about that
- comply with `_metrics: Dict[Sequence[str], T]`

We may be able to keep type checking and handle those problems via casts or ignored errors but I'm not sure it's better than introducing a variable.

From my perspective, it looks like we can easily remove "type unsafety" from `labels()` for a very small cost. It's already built to accept anything that can be `str`ed, a perfect candidate for `object` rather than `Any`.
Just in our code base, it would introduce type safety to hundreds of lines. So in the global ecosystem of Python projects using `prometheus-client`, it will be a big type safety improvement!

But you're the maintainer and it could make sense to consciously choose type unsafety in some parts of the lib. We can do some kind of wrapper on our side to confine the Prometheus client type unsafety part.

I hope this is clear and not too long 😄

csmarchbanks (approved):
Thanks for the additional context! I think I am good with this as is, we can always tweak it more in the future if needed.

gregoiredx:
great news, thanks!

Do you already know when we can expect the next release?

PR #1189, opened 2026-07-11
Validate Enum arguments before registering the collector

sean-kim05:
Enum.__init__ called super().__init__() -- which registers the collector
in the CollectorRegistry -- before validating that states is non-empty
and that the metric name does not overlap a label name. When either
check failed, the ValueError was raised as expected, but a half-built
Enum (whose _states was never assigned) had already been registered.

That left the registry in a broken state: the name was permanently
taken, so recreating the metric raised 'Duplicated timeseries', and any
subsequent scrape crashed with

    AttributeError: 'Enum' object has no attribute '_states'

when _child_samples iterated self._states. A realistic trigger is
building the states list from configuration that turns out to be empty.

Gauge and Histogram already validate before calling super().__init__();
this moves Enum's two guards ahead of registration to match, so a failed
constructor leaves the registry untouched.

Add test_failed_init_does_not_pollute_registry, which asserts that after
two failed Enum constructions the name is still free, the metric can be
created, and the registry scrapes cleanly.

@csmarchbanks

### DIFF
--- a/prometheus_client/metrics.py
+++ b/prometheus_client/metrics.py
@@ -769,6 +769,10 @@ class Enum(MetricWrapperBase):
                  _labelvalues: Optional[Sequence[str]] = None,
                  states: Optional[Sequence[str]] = None,
                  ):
+        if name in labelnames:
+            raise ValueError(f'Overlapping labels for Enum metric: {name}')
+        if not states:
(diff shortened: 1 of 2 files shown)


csmarchbanks (approved):
Thanks!

PR #1190, opened 2026-07-12
Do not emit a leading zero in floatToGoString exponents >= 10

sean-kim05:
`floatToGoString` is meant to mirror Go's `strconv.FormatFloat(f, 'g', -1, 64)`, which pads the exponent to a minimum of two digits (`1e+06`, `1e+10`, `1e+15`). The current formatting hard-codes a single `0` in front of the exponent:

```python
return f'{mantissa}e+0{dot - 1}'
```

That is only correct while `dot - 1 < 10`. As soon as the exponent needs two digits, the literal `0` becomes a spurious third digit:

| value | Go / expected | current output |
|-------|---------------|----------------|
| `1e6`  | `1e+06` | `1e+06` |
| `1e10` | `1e+10` | `1e+010` |
| `1e15` | `1e+15` | `1e+015` |
| `1234567890123.0` | `1.234567890123e+12` | `1.234567890123e+012` |

So any histogram/summary whose `le`/quantile boundary is at or above `1e10` serializes a bucket label that doesn't match the value Go/Prometheus would emit for the same float.

Fix: zero-pad the exponent to two digits rather than prepending a literal `0`:

```python
return f'{mantissa}e+{dot - 1:02d}'
```

This keeps the existing two-digit padding for small exponents (`1e+06`) and drops the extra zero for large ones (`1e+10`).

Added `tests/test_utils.py` covering both the single-digit (still zero-padded) and two-digit (no leading zero) cases.

### DIFF
--- a/prometheus_client/utils.py
+++ b/prometheus_client/utils.py
@@ -21,7 +21,7 @@ def floatToGoString(d):
         # We only need to care about positive values for le/quantile.
         if d > 0 and dot > 6:
             mantissa = f'{s[0]}.{s[1:dot]}{s[dot + 1:]}'.rstrip('0.')
-            return f'{mantissa}e+0{dot - 1}'
+            return f'{mantissa}e+{dot - 1:02d}'
         return s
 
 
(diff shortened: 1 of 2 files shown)


csmarchbanks (approved):

csmarchbanks:
Thanks!

PR #1192, opened 2026-07-13
Parse the native histogram sum as a float

sean-kim05:
_parse_nh_struct parsed the sum field with int(), so a native histogram
whose sum is not integral failed to parse with

    ValueError: invalid literal for int() with base 10: '100.5'

NativeHistogram.sum_value is a float and the OpenMetrics writer emits the
sum verbatim (e.g. sum:100.5), so the library could not parse its own
output for the common case of a non-integer sum. Parse sum with float()
to match; count, schema and zero_count remain int as they are genuinely
integers.

Add test_native_histogram_float_sum, which parses a native histogram
with a fractional sum and checks it round-trips to sum_value 100.5.

@csmarchbanks

### DIFF
--- a/prometheus_client/openmetrics/parser.py
+++ b/prometheus_client/openmetrics/parser.py
@@ -315,7 +315,7 @@ def _parse_nh_struct(text):
     deltas = dict(re_deltas.findall(text))
 
     count_value = int(items['count'])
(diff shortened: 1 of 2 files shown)


csmarchbanks (approved):
Thanks!

